PR #124, opened 2014-06-06
Ioslides levels

alequartz:
## Change
Added the slide_level  argument to the  ioslides_presentation function

## Rationale
The argument allows to specify what header level in the markdown should be used as a slide title. The behavior mimics the slide_level argument of the beamer_presentation function.

## Implementation
The slide_level argument is passed to the _lua_ file and added to the list of global variables.
All headers below the slide_level are considered as _segue_ title slides. The headers corresponding to the
the slide_level are considered as slide titles and used as slide separators.

In order to distinguish between _segue_ slides coming from different header levels a new class has been added to the slide class list: _level_**X**. Where X stands for the header level. It is therefore possible to create CSS styles specific for each type of _segue_ slide.

The _tests/testthat/test-ioslides.R_ file contains the relevant unit tests.


alequartz:
Hi jjallaire,
Could you merge this patch in the next release ? 
Thanks

jjallaire:
Yes we can definitely merge this. I'd ask a few things though:

1. Please send a contributor agreement (if you haven't previously): https://github.com/rstudio/rmarkdown/blob/master/CONTRIBUTING.md

2. There has been quite a bit of work done in ioslides_presentation.R since your PR so the merge commit has a lot of noise in it. Would you mind reapplying your changes against the current master branch?

3. There were some whitespace issues where I think your editor was inserting full tabs (whereas we use only spaces for indentation). You can see what I'm talking about if you look at the diffs for the PR (e.g. the slide_level parameter is off the screen to the right.

alequartz:
Hi, 
Fixed the code as requested.

As you can see my tests are currently not passing on github. 
However I cannot reproduce this error locally. To make sure I have also tried running the tests using the rocker/hadleyverse docker. All my tests pass.

Any clues of why my tests could be failing ? Could it be due to a difference in the version of packages installed for the github tests. If so, do you have a docker instance I could use to try and debug this ?
Thanks 



 I have also tried 
 have tried running in on the 
 seem to have a problem passing the checks in gitThe tests pass on my machine using the R installed in

yihui:
You are comparing characters with numbers, e.g. `grep("level1.*Header1", html_file, value = TRUE) > 0` (`value = TRUE` makes `grep()` return a character vector) That does not make sense, although R will not complain about it. You should do `length(grep()) > 0`.

alequartz:
Dear yihui,
thank you for pointing this out. Not only it doesn't make sense but it was the actual culprit for my error. R will indeed not complain, however depending on the locale the test will fail. Thanks !

yihui:
I think the tests need to be skipped on CRAN, i.e. `skip_on_cran()` just like the tests in test-formats.R, because we cannot guarantee Pandoc is available on CRAN.

PR #222, opened 2014-08-19
Append extra_dependencies after all other dependencies

jimhester:
Doing this allows extra_dependencies to be loaded last, this is
beneficial if a dependency depends on jquery for instance.

This also allows you to load both bootstrap 3 and bootstrap 2 in the same document.


jjallaire:
Closing as we've updated to Bootstrap 3 in rmarkdown.

PR #276, opened 2014-10-13
Update vignette preamble to use VignetteEncoding

jimhester:
This uses the VignetteEncoding rather than `\usepackage[utf8]{inputenc}`.  The check logic was changed in commit https://github.com/wch/r-source/commit/b1648464ee4c5fe3aa227a6151a9f23a9e868b40.  The previous template also failed R CMD check with non-ASCII vignettes because of the prefixed `%` in the encoding line, the vignette lines should have been

```r
vignette: >
  %\VignetteIndexEntry{Vignette Title}
  \usepackage[utf8]{inputenc}
```

### DIFF
--- a/inst/rmarkdown/templates/html_vignette/skeleton/skeleton.Rmd
+++ b/inst/rmarkdown/templates/html_vignette/skeleton/skeleton.Rmd
@@ -6,7 +6,7 @@ output: rmarkdown::html_vignette
 vignette: >
   %\VignetteIndexEntry{Vignette Title}
   %\VignetteEngine{knitr::rmarkdown}
-  \usepackage[utf8]{inputenc}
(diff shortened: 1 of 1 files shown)


jjallaire:
@yihui Should we merge this PR now?

jjallaire:
If we do this will it fail when building against R 3.0? If so should we wait until R 3.2 is released to merge?

yihui:
I think we can merge this now. I was saying R 3.1.2 instead of R 3.2. Actually R <= 3.1.2 has never worked in terms of identifying the vignette encoding and pass it to the vignette engine, and the bug was fixed in R 3.1.3. So merging this PR will be an improvement to R >= 3.1.3 users, and should not affect R <= 3.1.2 users.

PR #284, opened 2014-10-15
Fixes #283.

### LINKED ISSUE #283: Support `opts_template`
The new `opts_template` options are not supported in the existing `knitr_options`.  These can be set in the same way the other options can be set, using `$set()` and `$get()`.

Currently any templates defined before calling `render` work as expected but it would be preferable to have explicit support.

jamiefolson:
Adds support for knitr templates.


jamiefolson:
Are there any issues with this pull request?  I'd be happy to help address any concerns..

jjallaire:
It looks like there are merge problems with the PR. My best guess is that this is because you are using an older version roxygen. Could you re-roxygenize with roxygen2 4.1.0?

R/output_format.R
```diff
@@ -120,6 +120,7 @@ merge_output_formats <- function (base, overlay)  {
 #'   \code{\link[knitr:opts_chunk]{opts_chunk}})
 #' @param knit_hooks List of hooks for R code chunks, inline R code, and output
 #'   (see \code{\link[knitr:knit_hooks]{knit_hooks}})
+#' @param opts_template List of templates for chunk level knitr options (see \code{\link[knitr:opts_template]{opts_template}})
```
yihui: Please reflow comments here. If you are in RStudio, you can just press Ctrl+Shift+/

Issue #341, opened 2015-01-01
Something deletes cached images


gonulate:
I run the following very basic RMD file:

    ---
    output: html_document
    ---

    ```{r, cache=TRUE}
    a <- 2
    ```

    ```{r, echo=FALSE, cache=TRUE}
    plot(rnorm(1e5), rnorm(1e5))
    ```

When I run this first time it runs perfectly. But second time it tries to locate the cached images and it fails. Because something deletes them. After pressing the "knit HTML" in Rstudio, I watch the folder contents in another window, The folder "trial1_files" created and mysteriously deleted just after the html file produced (when I run it first time). Later, when I knit it again, it gives an error because it cannot locate the deleted folder. The "trial1_cache" folder is there intact. 

This is weird. The cache functionality run without error in Latex documents. 

I asked the same question in knitr forum. Yihui Xie offered following solution (updating the package), but it did not work. 
```
update.packages(ask = FALSE) 
devtools::install_github('rstudio/rmarkdown') 
```

RStudio version: 0.98.1091
rmarkdown version: 0.4.2
knitr version : 1.8

Here is the error:
```
  |................                                                 |  25%
  ordinary text without R code


  |................................                                 |  50%
label: unnamed-chunk-1 (with options) 
List of 1
 $ cache: logi TRUE


  |.................................................                |  75%
  ordinary text without R code


  |.................................................................| 100%
label: unnamed-chunk-2 (with options) 
List of 2
 $ echo :


processing file: trial1.Rmd
 logi FALSE
 $ cache: logi TRUE




[...]

PR #342, opened 2015-01-03
correctly detect existence of cache dir on windows

jjallaire:
We were not recognizing the existence of the cache_dir on windows because it ended in a trailing slash, and file.exists on windows requires no trailing slash. This resulted in our not properly recognizing that there was a cache directory and as a result removing the _files directory when we shouldn't (see bug https://github.com/rstudio/rmarkdown/issues/341)

### DIFF
--- a/R/render.R
+++ b/R/render.R
@@ -222,6 +222,10 @@ render <- function(input,
     cache_dir <-knitr_cache_dir(input, pandoc_to)
     knitr::opts_chunk$set(cache.path=cache_dir)
 
+    # strip the trailing slash from cache_dir so that file.exists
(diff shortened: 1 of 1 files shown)


yihui:
Although this fixes #341, the fundamental fix is to use `file_test('-d')` instead of `file.exists()` to check if a dir exists. `file_test('-d')` does not have the trailing slash problem. I meant to point this out a long time ago since I saw `file.exists()` was used in many places (not only in this package), but I forgot to do so.

jjallaire:
Ah, I didn't know about file_test ! Will be mindful to use that from here
on out.

Issue #346, opened 2015-01-08
sh: 1: /home/travis/opt/pandoc/pandoc: not found


audy:
Rmarkdown is trying to find pandoc instead of just using `$PATH`.

the [`find_pandoc`](https://github.com/rstudio/rmarkdown/blob/d4947c70cbbc7a588a18a71dfcbcabef0dba1412/R/pandoc.R#L487) function is supposed to find the Pandoc _executable_, which it does, using `Sys.which`.

Once found, `find_pandoc` tries to find pandoc's version [here](https://github.com/rstudio/rmarkdown/blob/d4947c70cbbc7a588a18a71dfcbcabef0dba1412/R/pandoc.R#L501) using the function [`get_pandoc_version`](https://github.com/rstudio/rmarkdown/blob/d4947c70cbbc7a588a18a71dfcbcabef0dba1412/R/pandoc.R#L527) which takes the location of the pandoc **binary** as first argument, but treats it as if it is the pandoc installation directory. It joins `pandoc` to the path of the binary (resulting in `pandoc/pandoc`) then tries to execute it crashing the program because the file does not exist.

I have a [`.travis.yml`](https://github.com/audy/make-rmarkdown/blob/5cb1d097b6c300fd436e3adc41ed256e62ee6f3c/.travis.yml) that reproduces this error.

Is there a way to override this or just use `$PATH`?

yihui:
To me, this looks more like a issue of @craigcitro's travis-tool.sh than an rmarkdown issue.

I do not use this travis script myself, and I just install pandoc from the marutter PPA:

    sudo apt-get install pandoc pandoc-citeproc

The version of Pandoc in his PPA is a little dated. If you really care about the latest version, here is another of my personal recipe: https://github.com/yihui/crandalf/blob/master/inst/scripts/install-pandoc

craigcitro:
`travis-tool.sh` puts it where @jjallaire recommended, see the [r-travis issue](https://github.com/craigcitro/r-travis/issues/136) for more detail. my understanding was that the path in question was one that RMarkdown always checked?

in any event, it sounds like the issue @audy mentions (of conflating the file path with a directory) is probably worth fixing?

yihui:
[...]

PR #347, opened 2015-01-09
Replace file.exists() with file_test('-d')

yihui:
The former does not work on directories with trailing slashes on Windows, whereas the latter does, e.g. file.exists('foo/') is FALSE, and file_test('-d', 'foo') is TRUE if the directory foo exists

BTW, this will also make the error message in #346 clearer.

### DIFF
--- a/R/pandoc.R
+++ b/R/pandoc.R
@@ -497,7 +497,7 @@ find_pandoc <- function() {
 
     # determine the versions of the sources
     versions <- lapply(sources, function(src) {
-      if (file.exists(src))
+      if (dir_exists(src))
         get_pandoc_version(src)
       else
         numeric_version("0")
--- a/R/render.R
+++ b/R/render.R
@@ -64,15 +64,10 @@ render <- function(input,
 
   # setup a cleanup function for intermediate files
   intermediates <- c()
-  on.exit(lapply(intermediates,
-                 function(f) {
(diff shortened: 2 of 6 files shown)


R/render.R
```diff
@@ -64,15 +64,10 @@ render <- function(input,
 
   # setup a cleanup function for intermediate files
   intermediates <- c()
-  on.exit(lapply(intermediates,
-                 function(f) {
-                   if (clean && file.exists(f))
-                     unlink(f, recursive = TRUE)
-                 }),
-          add = TRUE)
+  on.exit(if (clean) unlink(intermediates, recursive = TRUE), add = TRUE)
```
yihui: @jjallaire In case you wonder the rationale behind the change here: there is no need to use a loop to remove the dirs: `unlink()` can take a vector of files/dirs, and it removes them only if they exists, so there is no need to check `file.exists()`, either.

R/pandoc.R
```diff
@@ -497,7 +497,7 @@ find_pandoc <- function() {
 
     # determine the versions of the sources
     versions <- lapply(sources, function(src) {
-      if (file.exists(src))
+      if (dir_exists(src))
```
yihui: In the case of #346, if `$HOME/opt/pandoc` is the pandoc executable instead of a dir, this path will be skipped, hence less confusing error message.

Issue #353, opened 2015-01-13
The default LaTeX template does not work with multiple authors 


yihui:
A minimal reproducible example:

```yaml
---
title: "R Markdown v2 Demo"
author:
  - John
  - Joe
date: "2015/01/01"
output:
  pdf_document:
    keep_tex: yes
---

Hello world.
```

The error message:

```latex
! Misplaced \crcr.
\endtabular ->\crcr 
                    \egroup \egroup $\egroup 
l.71 \maketitle
```

I have no idea what this means. Perhaps @robjhyndman can help again? Thanks!

PR #354, opened 2015-01-13
Fixed bug with multiple authors in pdf output

robjhyndman:
Fixes https://github.com/rstudio/rmarkdown/issues/353

### DIFF
--- a/inst/rmd/latex/default.tex
+++ b/inst/rmd/latex/default.tex
@@ -149,7 +149,7 @@ $else$
   \posttitle{}
 $endif$
 $if(author)$
-  \author{$for(author)$$author$$sep$ \and $endfor$}
+  \author{$for(author)$$author$$sep$ \\ $endfor$}
   \preauthor{\centering\large\emph}
   \postauthor{\par}
 $else$


yihui:
Thanks!, But I guess this fixes the symptom (and changes behavior) without curing the disease, and I'd very much like to preserve `\and` since it is the correct thing to do in LaTeX when writing multiple authors.

robjhyndman:
I'm not sure what you mean by "correct". The definition of `\and` is

    \def\and{%                  % \begin{tabular}
      \end{tabular}%
      \hskip 1em \@plus.17fil%
      \begin{tabular}[t]{c}}%   % \end{tabular}

This normally works because LaTeX sets the authors within a tabular environment. But rmarkdown uses the titling package which does not use the tabular environment, and then the definition causes a problem. 

My fix was just to replace `\and` with `\\\\` which avoids the use of the tabular environment and so it does not cause a clash when the titling package is loaded. 

There is some discussion of this issue at https://tex.stackexchange.com/questions/109658/and-command-in-author-tag

robjhyndman:
Note that I also mimicked the formatting for html and doc outputs which places each author on a new line. Using `\and` without the titling package will put authors on the same line. Presumably we want the output to be as similar as possible for each output type.

yihui:
By "correct", I mean `\and` is what we normally use to separate authors. If you hard-code it to be a line break `\\`, there is no way to customize it. For example, the original layout was to put authors on one line, and that is what I'd prefer.

@jjallaire I feel we opened a can of worms by defining our own LaTeX template here. I looked at Pandoc's default template again, and the spacing does not look too bad:

---

image

---

This is what our template produced:

---

image

---

Anyway, since it is trivially easy to switch to Pandoc's default template in R Markdown, this issue does not really bother me too much.

jjallaire:
@yihui I hear what you are saying and am certainly tempted to go back to using pandoc's template. I have heard however from a ton of people that they don't like the default pandoc spacing at the top so it wouldn't be without significant detractors. I think we should leave it as-is for now and perhaps make it more obvious (via documentation and the IDE) how to switch to the default template.

PR #360, opened 2015-01-22
Added subtitle command to default.tex to allow for subtitle in yaml header

johnstantongeddes:
I think it would be nice to include subtitles in the default pdf layout (and apparently [I'm not alone](http://stackoverflow.com/questions/26043807/multiple-authors-and-subtitles-in-rmarkdown-yaml)).

This may have been planned since `\subtitle` already existed in the default.tex template, but this command is not defined in the default `article` documentclass in latex. I added `\newcommand{subtitle}` to default.tex so that a subtitle will render. 

Example yaml header:

~~~
---
title: "Title"
subtitle: "Subtitle Here"
author: "John Stanton-Geddes"
date: "January 22, 2015"
output:
  pdf_document:
    keep_tex: yes
---
~~~

This is my first pull request so I apologize for any mistakes, and would be happy to close/open as an issue if there's a reason not to implement this. 

Thanks for the work on rmarkdown!

### DIFF
--- a/inst/rmd/h/default.html
+++ b/inst/rmd/h/default.html
@@ -90,7 +90,7 @@ $if(title)$
 <div id="$idprefix$header">
 <h1 class="title">$title$</h1>
 $if(subtitle)$
-<h1 class="subtitle">$subtitle$</h1>
+<h3 class="subtitle"><em>$subtitle$</em></h3>
 $endif$
 $for(author)$
 $if(author.name)$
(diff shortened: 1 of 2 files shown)


johnstantongeddes:
I'm closing this and will be re-opening another request. This request fixed subtitle for the pdf only. My new request also modifies the `default.html` file so that the subtitle renders in html output.

PR #370, opened 2015-02-07
Tuft(ish) book template

elbamos:
This is an adaptation of the Tufte-Book latex class.  This is more flexible and has more features than Tufte-Handbook.  Many of these are shown in the skeleton.  Couple notes:

* I don't know how to make this show up in the rstudio template chooser
* There are a small number of nits I'm not yet completely happy with; I'm making the request now hoping that it may get others to pay a bit of attention to one of those issues, which is compatibility with xetex and luatex.  This is an issue inherited from the Tufte-Handbook template.  Its an issue because pdflatex can easily run out of memory with tikzDevice.  Making it work seems to be beyond my TeX skills.  Either way, the template is useable now.


jjallaire:
You've done everything you need to do to get the format to show up in the RStudio template chooser (it shows up for me). You may be having a problem related to RStudio auto-updating rmarkdown over the top of your version. If you get the latest version of RStudio (v0.98.1102) then you won't have this problem.

jjallaire:
I think we should consider creating a new package to host the tufte formats (we could then move the tufte_handout from rmarkdown to this package). Would you consider creating and maintaining this package?

jjallaire:
I think you should move these files from the resources directory to the skeleton directory:

tufte_book/resources/tufte-book.cls
tufte_book/resources/tufte-common.def
tufte_book/resources/tufte.bst

This is because (on Linux at least) pandoc doesn't seem to find the tufte class in the resources directory and thus has to fall back to getting it from the system installed latex packages (where it's not always found). If we move the classes to the skeleton directory (and change the create_dir variable to true) then these files will be copied alongside the user's Rmd file where they are always accessible.

elbamos:
@jjallaire 

1.  Ok, I will upgrade today, thanks for letting me know.

2.  I'm really not the right person to be maintaining a TeX package -- I know only the TeX I had to learn to make the class work. Less than that really.  

3. Thank you! Mac seems to have the same issue - that explains why some changes weren't getting picked-up.  I'll do that today.  I think I'm also going to rename it "tuftish-ebook" to avoid confusion.

jjallaire:
I'm not suggesting that you maintain a TeX package, rather an R package that hosts this template. Even though we have the tufte_handout format in the rmarkdown package, this is really more for the purpose of providing an illustrative example of a custom LaTeX format. i.e. we don't want to start collecting additional formats in this package but instead have them existing in other packages. 

The rticles package is an example of another package that hosts several LaTeX formats: https://github.com/rstudio/rticles. I'd say you should either create a new tufteformats package or perhaps submit this as a PR to the rticles package. I'll close this PR but look forward to working with you on getting this format refined and packaged elsewhere.

elbamos:
Meh... The reason I say that about TeX is the tufte formats are very sensitive to TeX issues.  If you try to create a table the wrong way inside a fullwidth environment, then you'll get all sorts of confusing latex compilation errors that I won't be able to help people with.  

I actually made the pull request after, when I said I couldn't get part of tufte-handout to work, someone at rstudio support suggested I do this and they'd be inclined to take it.  I certainly understand why you wouldn't want to proliferate templates, although I sort of have the opposite reaction, which is I couldn't figure out why rticles was a separate package when it seemed like it should be consolidated.

elbamos:
I implemented the changes from earlier, fixed what had been a nagging bug, and added some examples.  I think it would be great if we could get this rolled-into the main package rather than split-off.

jjallaire:
The reason is that there could in theory be dozens of templates and there
will be different levels and types of expertise required to maintain
distinct groups of them. We are not as a group LaTeX experts either so
don't want to imply that we can support and enhance sophisticated LaTeX
based templates as part of the core package (the main LaTeX template is
actually provided by pandoc rather than developed internally).

jjallaire:
Let's put it in rticles. Many of the templates in there currently are via
pull requests and at some point I'm hopeful we can find a LaTex expert
within the R community to maintain and enhance the package, including
submitting it to CRAN.

elbamos:
Oh I totally understand and respect your reasoning.  I just disagree about your conclusion.  It seems to me that you guys were, in a way, "asking for it" when you included the very complex tufte-handbook template.  The reason I went and did this, was because tufte-handbook doesn't work with a lot of standard rmarkdown stuff, and when I tried to get support for tufte-handbook, it wasn't available.  So I took a more flexible variant of the tufte class and modified it to be more consistent with the rest of rmarkdown.  

I think you may actually get fewer issues by including this than by including just tufte-handbook, because this provides documentation for many things, and it also relieves a lot of limitations on the implementation of tufte-handbook.  For example, there was no way to get tufte-handbook to produce a table of contents.  Or table of figures.  Or to number sections.  Or to compile with luatex.  There was no documentation for the figuretable feature.  Or making fullwidth sections.  Etc., etc., etc.

Similarly, the documentation on rstudio.com for making citations and bibliography entries is... well... 

Putting this in will, I think, help you in those respects.  

If you want to put it in rmarkdown, I will commit to try to spend time maintaining this part of the package.  If you decide to put it in rticles instead (is that your package as well?), I'll respect your decision.

jjallaire:
We were definitely asking for it by putting in tufte-handout! I'm sold
based on your analysis -- will do a final review and hopefully merge your
PR into rmarkdown tomorrow (and thanks for offering to help with it on an
ongoing basis, hopefully it's complete and robust enough that it won't
require this).

After merging I'll probably deprecate tufte_handout (seems to make sense
based on all of the limitations that your new format overcomes).

elbamos:
Thanks!  FYI the most recent version of this is at rticles.  I will try to move it back here.  I don't know if you really want to get rid of tufte-handout; for shorter, more lightweight things (i.e., handouts) it's a good template.  I'll also try to remove some limitations on tufte-handout, if I can.

jjallaire:
I'd be in support of keeping tufte-handout so long as we can remove some of
the headaches / limitations that you mentioned (I don't want users to keep
running into problems and we throw our hands up and say "just use the book
template" when that might be a misfit").

Let's include enhancements to tufte-handout in this PR (understood that
some of these enhancements might not even be possible and that you might in
turn throw up your hands and decide tufte-handout just is what it is).

R/tufte_book.R
```diff
@@ -0,0 +1,86 @@
+
+#' Tufte book format (PDF)
+#' 
+#' Template for creating a book according to the style of 
+#' Edward R. Tufte and Richard Feynman. 
+#' 
+#' @inheritParams pdf_document
+#' 
+#' @export
+tufte_book <- function(fig_width = 4,
```
jjallaire: Let's use the same order of parameters as used for pdf_document (e.g. toc options are first, pandoc_args are at the end, etc.).

R/tufte_book.R
```diff
@@ -0,0 +1,86 @@
...
+                          fig_crop = TRUE,
+                          highlight = "default",
+                          keep_tex = FALSE,
+                          includes = NULL,
+                          pandoc_args = NULL,
+                          toc = TRUE,
+                          toc_depth = 3,
+                          latex_engine = "xelatex",
+                          number_sections = TRUE,
+                          ...
```
jjallaire: Any argument we think is relevant / supported should be included in the signature (for documentation to end users of what is expected to work).

R/tufte_book.R
```diff
@@ -0,0 +1,86 @@
...
+    knitr_options$opts_knit <- list()
+  if (is.null(knitr_options$knit_hooks))
+    knitr_options$knit_hooks <- list()
+  
+  # set options
+  knitr_options$opts_chunk$tidy <- TRUE
+  knitr_options$opts_knit$width <- 45
+  
+  # set hooks for special plot output
+  knitr_options$knit_hooks$plot <- function(x, options) {
```
jjallaire: Share this code with tufte_handout?

inst/rmarkdown/templates/tufte_book/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,264 @@
+---
+title: "Tuftish Book"
+author: "John Smith"
+date: "August 13th, 2014"
+bibliography: "knitcitations.bib"
+output: 
+  tufte_book:
+    latex_engine: lualatex
```
jjallaire: If we think lualatex is preferred perhaps it should be the default? If not I don't think we should reset this here.

inst/rmarkdown/templates/tufte_book/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,264 @@
...
+library(ggplot2)
+library(papeR)
+library(pander)
+cite_options(citation_format = "pandoc",
+  					 cite.style= "authortitle",
+						 style="citation",
+						 hyperlink="to.doc",
+						 check.entries=FALSE)
+
+opts_chunk$set(   echo = FALSE,
```
jjallaire: I think it's fine to illustrate the setup chunk, but I don't think we should reset all of this default behavior (e.g. adding caching). Perhaps just a one-liner with a couple options or even a commented out version. If we really think these options are important then we should actually define them as part of the format.

inst/rmarkdown/templates/tufte_book/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,264 @@
...
+  								warning=FALSE,
+							 		error=TRUE,
+									message=FALSE,
+									tidy=TRUE,
+									cache=TRUE,
+							 		fig.keep = 'high',
+							 		dev = "tikz",
```
jjallaire: I don't think this should be the default as many user won't have this package and the document will therefore fail to compile.

inst/rmarkdown/templates/tufte_book/template.yaml
```diff
@@ -0,0 +1,5 @@
+name: Tufte Book (PDF)
+description: >
+  Template for creating a book according to the style of 
+  Edward R. Tufte and Richard Feynman.
+create_dir: false
```
jjallaire: I think this should be TRUE (since we have to copy a bunch of files side by side with the skeleton.Rmd)

jjallaire:
Also, could you send is a contributor agreement as described here: https://github.com/rstudio/rmarkdown/blob/master/CONTRIBUTING.md

elbamos:
Thank you for the comments, I'll implement today and move this back from rticles. I've gone back and forth and I think the default should be xetex. The reason this is complicated is there have been issues between xetex/luatex and one of the packages used for letter spacing. I've worked-around that for xetex, but with luatex the title page spacing is off. The reason to not pick pdftex is that it can run out of memory. This took me some time to diagnose, and more to fix, since I didn't have any latex experience. There's also an issue that some citation managers will produce high-bit characters that may need xetex/luatex, but that isn't a bug for here.  You're right about moving that code into tufte handout. There will also have to be some tinkers to the tufte handout latex to implement the workarounds, and I'm not sure they'll all work. I will have to experiment.  

What's your timing on the next cran release-not this week, right?

jjallaire:
That all sounds great.

We just did a CRAN release last week so the next one won't be for another
month or so.

elbamos:
Thanks; quick question -- is there any way to insert latex before and after the bibliography?  rmarkdown/knitr seems to take care of this outside the text and outside the template. I'd like to put the bibliography into full-width, two-column mode, and give it an appropriate entry in the TOC.

jjallaire:
If you look at the end of the default LaTeX template you can see how
bibliographies are handled there:

https://github.com/rstudio/rmarkdown/blob/master/inst/rmd/latex/default.tex#L206-L227

The include-after variable is indeed after the bibliography. If you want to
be able to insert arbitrary content before or after the bibliography you
could also use custom pandoc variables (you can add them with the
rmarkdown::pandoc_variable_arg function).

elbamos:
I *think* my latest commit should implement everything you requested for tuftish-ebook.  I have not been able to get the bibliography thing working right.  I see what's going on in default.tex, but tuftish-ebook.tex never calls \printbibliography or \bibliography.  I thought perhaps its embedded in \backmatter or \printindex, but after deleting those the bibliography still shows up!  If you have any suggestions let me know.

NAMESPACE
```diff
@@ -36,5 +36,6 @@ export(rmarkdown_format)
 export(run)
 export(slidy_presentation)
 export(tufte_handout)
+export(tuftish_ebook)
```
jjallaire: I don't like the "tuftish" designation nor the "ebook" designation. Is the book not intended to be printed? Why not tufte_book?

R/tufish_ebook.R
```diff
@@ -0,0 +1,86 @@
+
+#' Tuftish e-Book format (PDF)
+#' 
+#' Template for creating an e-book based on the style of 
+#' Edward R. Tufte and Richard Feynman. 
+#' 
+#' @inheritParams pdf_document
+#' 
+#' @export
+tuftish_ebook <- function(fig_width = 4,
```
jjallaire: We should order these arguments the same as pdf_document (e.g. toc args go first)

R/tufish_ebook.R
```diff
@@ -0,0 +1,86 @@
+
```
jjallaire: There are actually two versions of the tuftish_ebook.R file (with different spellings). I think this is the one that should be deleted but let me know.

R/tuftish_ebook.R
```diff
@@ -0,0 +1,91 @@
...
+#'
+#' @inheritParams pdf_document
+#'
+#' @param latex_engine LaTeX engine for producing PDF output. Options are
+#'   "pdflatex", "lualatex", and "xelatex".  Note that lualatex may have
+#'   problems with text spacing, and that pdflatex may have memory issues
+#'   when using tikzDevice.
+#'   
+#' @export
+tuftish_ebook <- function(fig_width = 4,
```
jjallaire: Same comment as above about order of parameters

inst/rmarkdown/templates/tufte_handout/resources/tufte-common.def
```diff
@@ -1,8 +1,8 @@
 %%
```
jjallaire: It seems like we should also move the tufte support files from resources to skeleton for tufte handout, do you agree?

inst/rmarkdown/templates/tuftish_ebook/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,276 @@
+---
+title: "Tuftish e-Book"
+author: "John Smith"
+date: "February 8, 2015"
+bibliography: "knitcitations.bib"
+output: 
+  tuftish_ebook:
+    latex_engine: xelatex
```
jjallaire: Let's remove the latex_engine argument here (too in the weeds for an example and will just encourage people to change it to pdflatex!)

inst/rmarkdown/templates/tuftish_ebook/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,276 @@
...
+
+```{r setup,echo=FALSE,message=FALSE,cache=FALSE,include=TRUE}
+# Its a good idea to start any rmarkdown document with a block like this one to set knit defaults, setup the bibliography, and so forth.
+library(knitcitations)
+library(knitr)
+library(xtable)
+library(ggplot2)
+library(papeR)
+library(pander)
+cite_options(citation_format = "pandoc",
```
jjallaire: If the cite_options and xtable options are important to the correct functioning of the template I wonder if they should become part of the format itself (we could set and reset them before/after the render). If you feel strongly that this would be better then I can even make changes in rmarkdown::render to make this possible.

elbamos:
I've made the first few of these changes and will commit shortly.  

* Yes about the parameters, yes about the misspelled tufish_ebook that got deleted.  

* Yes on the default engine, and I made a note about that in the text of the skeleton.

* On the citation format, citation_format has to be set for it to work.  That may be something to raise with the maintainers of RefManageR and knitcitations.  The other default-changes are just... well it can take someone a while to figure out why things aren't working as expected, so I tried to make that more visible.  

I'm not a fan of forcing altered-defaults into a template.  Its really confusing for the user, because something doesn't work, or changing some setting doesn't work, and then you have to hunt through someone else's source code to figure out why.  Some of the stuff I did here, was *un*-forcing forced defaults in tufte-handout.  I think if we're changing package defaults we should make that visible.  

* The xtable defaults do have to be set.  This is documented also in other parts of rmarkdown.

* To clarify the name, "tufte handout" and "tufte book" are latex classes built into the standard latex distribution.  I had to make changes to the class definition file to fix some things (the class hasn't been updated in 4 years), and then I made more changes for conformity.  I think retaining the name would be misleading for people who are familiar with the latex class. 

The reason for "ebook" is, I needed a different name than "book," and I wanted the format to be more consistent with the rest of rmarkdown.  For example, internal hyperlinks should look like hyperlinks.  And if you're distributing a document by e-mail, even-numbered pages shouldn't have a different format from odd-numbered pages, and you don't need three blank pages before and after every chapter heading.  

I realized that what I was doing was making the format more amenable to creating an e-document.  

Could you paper-print it?  Sure.  If you were making an actual *book*, then you'd probably go into the class.tex file, take [oneside] out of the top, and possibly add [symmetric].

jjallaire:
Okay, agreed with your analysis regarding forcing altered defaults into a
template (your method is much more transparent).

How about we name it "tufte_ebook"  ?  (rather than tuftish_ebook)

elbamos:
Works for me!  I just didn't want to be presumptuous.

jjallaire:
It looks like this might be ready to go. Is that correct? 

Could you also send a contributor agreement as described here: https://github.com/rstudio/rmarkdown/blob/master/CONTRIBUTING.md

elbamos:
Yes, its good to go.  I'll send the contributor agreement later today, thanks for reminding me.  I haven't taken the changes and put them back into tufte-handout, because I've still been tinkering, but I'd like of prefer that to be a separate pull.

jjallaire:
I'm doing some local testing of the PR and noticed that because of the knitcitations, papeR, pander, etc. dependencies that users can't knit without installing a bunch of other packages. 

This isn't by itself a problem, but a problem is created by the fact that rmarkdown gets installed by RStudio (when the user attempts to knit for the first time) and then the Tufte e-Book template will show up in the list Templates forever after. This means that we will end up doing quite a bit of tech support on this format (and answering questions about how the citations works, where can I get this or that package because my version of R doesn't seem to have it, etc.). This is a fairly complex format / workflow, and I feel like we won't be able to do a great job supporting it (believe me it's hard enough to support all of the other "balls in the air" created by working with R!).

For this reason I'm flip-flopping and thinking we are better off putting this in rticles. Your point about Tufte Handout is still well taken and I'm thinking we should move that to rticles as well. I am happy to do all of the legwork on this (unless you feel strongly about having the commits in rticles being from you rather than me -- up to you). For what it's worth, we're hoping to submit rticles to CRAN and promoting it on our blog, etc. so I don't think it will lack for visibility.

elbamos:
Those are very good points. One way to handle them is for me to put in code that tests for the packages and prints a note if they aren't installed. With knitcitations, though, the only option is to throw an error if the package is missing, because the .bib file is references in the Yaml header and pandoc I think will spit-up if it isn't found. I can test that. Anyway, however you'd prefer me to address it - test for packages, pull out the code, move to rticles, whatever - I'll defer to your judgment.

elbamos:
Oh and just to be clear -- while I wouldn't *mind* having the commits under my name, really whatever is easiest for you guys, I'm just very happy to contribute.

jjallaire:
Okay, let's put it in rticles (I'll also move the tufte-handout format over
there). If you want to submit the PR to articles that will give you
attribution for the work so that's probably best.

elbamos:
Done.

jjallaire:
Great, just merged!

Issue #270, opened 2014-10-07
Pandoc on Windows


stephenll:
Pandoc no longer requires miktex on windows to run.  I have been testing texlive under windows with rmakrdown/pandoc for the two past months on two differently configured system and haven't had a single problem.  I think it is safe to say that it is recommended to use miktex but not necessary.

PR #373, opened 2015-02-10
add ulem for strikeout and morefloats for complex layouts

braverock:
Pandoc markdown defines a ~~ syntax for strikouts, but the tufte-handout
template did not support it.  Adding [normalem]{ulem} supports strikeout
without adjusting other formatting, and should be supported by all standard
TeX distributions.

LaTeX by default supports only a limited number of floating elements.
The tufte-handout template uses multi-column layouts and makes liberal
use of floating elements to layout things like side-notes, graphics, and
equations.  If you have a large number of side-notes or equations on a
page, LaTeX cannot lay out the page, and will stop with an error.
Adding {morefloats} resolves this problem, allowing more complex pages.
It is a core package, and should be available with all main LaTeX
distributions.

### DIFF
--- a/inst/rmarkdown/templates/tufte_handout/resources/tufte-handout.tex
+++ b/inst/rmarkdown/templates/tufte_handout/resources/tufte-handout.tex
@@ -38,6 +38,12 @@ $endif$
 % lipsum
 \usepackage{lipsum}
 
+% strikeout
+\usepackage[normalem]{ulem}
+ 
+% morefloats
+\usepackage{morefloats}
+
 % title / author / date
 $if(title)$
 \title{$title$}


jjallaire:
@elbamos It looks like this PR has some of the same improvements that your PR includes. @braverock, there is another pending PR that improves tufte_handout and also adds a new tufte_book format here: https://github.com/rstudio/rmarkdown/pull/370. 

If it's easy to roll these changes into the existing PR that would be great. If that would be unwieldy for any reason let's wait until the first PR is merged then re-submit this one against the new master.

braverock:
tufte-handout and tufte-book serve somewhat different purposes.

I'm happy to look at the tufte-book PR and merge in the two packages I suggest adding to tufte-handout.

I also like your idea in the tufte-book report #370 to mrege it to rticles.  I think it will likely get the most traction there.

jjallaire:
Yes, understood. Our idea is to have keep both of them (as part of the
tufte-book PR there were also some improvements to tufte-handout which is
why I suggested we consolidating the work).



On Tue, Feb 10, 2015 at 9:01 AM, Brian G. Peterson <notifications@github.com

braverock:
I'll pull PR  #270 and see if I can merge.

elbamos:
Both these are fine in tuftish e-book and I'll include them in a commit shortly to make it easier.

elbamos:
@jjallaire just to be clear, though, there's zero overlap between the changes in the two pull requests.

jjallaire:
Okay, if that's true should I just take Brian's request now?

elbamos:
I don't see why not.  I took his change to tufte-handout, tested and pasted it into the last commit for tufte-ebook, so its already in that as well.

PR #374, opened 2015-02-12
Allow subtitles in both pdf and html output.

johnstantongeddes:
Following up on a previous pull request (#360) that I just closed. The previous request added the ability to include a subtitle in pdf output, but the html output had a subtitle that was not visually distinguishable from the title (both `<h1>`). This request also modifies default.html so that the subtitle is distinct from the title (`<h3>` and `<em>`). Those choices were my preference and I'd be happy to change.  Please let me know if I need to make any other changes to get this merged. Thanks!

### DIFF
--- a/inst/rmd/h/default.html
+++ b/inst/rmd/h/default.html
@@ -90,7 +90,7 @@ $if(title)$
 <div id="$idprefix$header">
 <h1 class="title">$title$</h1>
 $if(subtitle)$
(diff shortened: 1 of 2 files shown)


jjallaire:
Thanks for this! We also need a contributor agreement (https://github.com/rstudio/rmarkdown/blob/master/CONTRIBUTING.md) in order to take the PR.

johnstantongeddes:
Great - will send the agreement later today!

PR #390, opened 2015-03-04
Numbered sections should begin at 1 not 0

lmmx:
Otherwise tables of contents look like

> * 0.1 First header
> * 0.2 Second header
> * ...

### DIFF
--- a/R/html_document.R
+++ b/R/html_document.R
@@ -144,7 +144,7 @@ html_document <- function(toc = FALSE,
 
   # numbered sections
   if (number_sections)
-    args <- c(args, "--number-sections")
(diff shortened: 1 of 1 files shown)


jjallaire:
Thanks for this. Could you send a contributor agreement as specified here: https://github.com/rstudio/rmarkdown/blob/master/CONTRIBUTING.md

yihui:
That is just because all your sections use the second-level (`##`) headings. If you add a first-level heading (`#`), you will get 1.1, 1.2, ... as expected. If you do not want to do so, you may use `pandoc_args: "--number-offset=1"` in the YAML frontmatter under `html_document` to deal with this particular case.

PR #404, opened 2015-03-19
Add support for incremental graphs in ioslides presentations

jamesonquinn:
When I'm making a presentation, I often want to build up my graphs incrementally, so that the viewer can parse them a little at a time. It's especially useful when giving a "surprise" result to first show the graph without the surprise, then add the surprise.

It is, of course, possible to do this in ioslides by simply copying and pasting multiple similar slides. But that's poor style; it violates DRY, and so for every small format change you want to make on the graph, you have to copy and "re-merge" the change for each of the copied slides.

I've added two functions, gginc and stages, to make this easier to do cleanly. I've also tweaked the javascript and css for ioslides to support this. The same basic idea could be applied to the other presentation output formats, but I've just done ioslides for now.

I don't think it's really a good idea to have this change exist as a separate package, given that it must integrate at the level of javascript and css.

I expect that this pull request will not be accepted as-is, and I'm ready to fix any glitches people find in my coding style and/or architecture. However, I feel strongly that this is a worthwhile improvement overall.


jamesonquinn:
I've created a corresponding issue for this pull request at [405](https://github.com/rstudio/rmarkdown/issues/405).

jjallaire:
Terrific idea! We are planning on spending some time on rmarkdown fixes/PRs/etc. after we finish the current RStudio release. We'll dig into this PR then (should be more than a few weeks from now). We also need a contributor agreement as described here: https://github.com/rstudio/rmarkdown/blob/master/CONTRIBUTING.md

jamesonquinn:
I sent you a contributor agreement.

Meanwhile, I'd like to blog about this work I've done. It's not such a huge
pull request, but it did actually take me a couple of days to get up to
speed on all the ancillary tools I needed to code this (compass, roxygen,
the css tricks I needed, the abstruse combinations of "substitute" and
"eval" I needed, etc.) so I'd like to show it off now, not a month or more
from now. Obviously, I don't need your permission to publicize my fork of
rmarkdown. But if you have any suggestions for how I should talk about it,
I'd be happy to follow them. For instance: would it be presumptuous of me
to call this a "pending pull request", or is there some better wording I
should use?

2015-03-19 14:47 GMT-04:00 JJ Allaire <notifications@github.com>:

jjallaire:
Blogging about it sounds like a great idea. I think calling it a "pending
pull request" sounds perfectly appropriate.

J.J.

jamesonquinn:
In order to make "stages" work inside "aes", I had to resort to blasphemous incantations of foetid, non-Euclidean effulgence. So this pull request is NOW OBSOLETE; the target SHOULD BE THE LATEST VERSION of my fork. Which totally works. But if you value your sanity, peer not at my unseelie reparser, o reader, or at least give me a code review / pointers to better code examples. I was trying to use R like lisp (ie, no amount of bquotes, evals, and substitutes were enough to get what I wanted) and there has got to be a better way to do that than what I did.

jjallaire:
The pull request is automatically updated to include all commits to your
fork so it's still up to date.

One implication of this is that all commits you make on your master ranch
are included in the PR, so if you go on to make other changes unrelated to
this PR they'll be included as well. That might be fine if this is the only
work you want to do on your fork. If however you have other sets of changes
you need to make you could create a local branch for this work and then
base the PR off of that.

yihui:
While I absolutely love this new feature for ioslides, apparently this PR has been left here for way too long. Sorry about that. Apologies again that it will probably be left here indefinitely... On one hand, I (as a probable maintainer of the rmarkdown package) read the code a couple of times and don't feel this PR is something I'm able thoroughly digest. On the other, I feel ioslides is in an unfortunate state that the upstream JS library is no longer being developed, and we don't seem to have enough human resources to carry on the development here, either.

So I'd encourage you to take a look at other presentation frameworks that are more active and better maintained. [remark.js](https://github.com/gnab/remark) is currently [my personal favorite](https://yihui.name/en/2017/08/why-xaringan-remark-js/) library, and I just created an example to show you how incremental slides work in remark.js: https://slides.yihui.name/xaringan/incremental.html I guess it can achieve what you originally wanted.

PR #467, opened 2015-07-09
add support for table of contents in word_document

daattali:


jjallaire:
I'm thinking we should do this via pandoc_args for now because the version of pandoc most users will be running (the one bundled with RStudio) does not yet support table of contents for word documents. 

In general we like to stay as close to the latest version of pandoc as we can however the procedure for building a new version is a bit involved. Specifically in order to create a standalone binary that works on many Linux distros we build pandoc from source and this exposes us to risk in the form of regressions that may have occurred in underlying Haskell dependencies. We also work with an outside contractor on the RHEL 5 version. As a result of all of this we wait until there is a pandoc version that has been labeled stable for a period of time (i.e. has no major regressions). Haven't yet seen this in the 1.14/1.15 series so haven't yet tried to update.

daattali:
I see, didn't know about those considerations but it makes sense

jjallaire:
Just added this on master: https://github.com/rstudio/rmarkdown/commit/693b0b1ce2b679c411c4ad4553612e2c943d90f6

Note that the upcoming version of RStudio bundles the latest release of pandoc (v1.15.0.6).

PR #473, opened 2015-07-21
knit_params_ask and render(params=ask)

aronatkins:
Implement `knit_params_ask` - a function that takes either a filename or file content and runs a Shiny app to customize its parameters. Returns a named-list containing overridden parameters.

Change `render(params=XX)` to optionally take a value `params="ask"` which uses the Shiny app to customize parameters for that single rendering.


R/params.R
```diff
@@ -0,0 +1,329 @@
+
+knit_params_get <- function(input_lines, params) {
+  # check for recent enough knitr
```
jjallaire: Do you need to check for a more recent knitr here? (to avoid the issue of `class` not being available in the params structure).

jjallaire: Sorry I know realize that this was the refactor so the version is correct.

R/params.R
```diff
@@ -0,0 +1,329 @@
...
+    converter <- params_converters[[c]]
+    if (!is.null(converter)) {
+      return(converter)
+    }
+  }
+  ## FIXME: this probably should be an error, as it is an unsupported type.
+  return(function(value){ value })
+}
+
+params_controls <- list()
```
jjallaire: We should be able to use S3 here to make the types of parameter controls externally extensible. Something like this:

```r
param_control <- function(param) {
  
  # create a fake object do to S3 lookup on
  x <- list()
  attr(x, "class") <- param$class
  
  # use appropriate method for class
  UseMethod("param_control", x)
}

param_control.default <- function(param) {
  NULL
}

param_control.integer <- param_control.numeric <- function(param) {
  ## If min/max are specified, use a slider.
  if (is.null(param$min) || is.null(param$max)) {
    shiny::numericInput
  } else {
    shiny::sliderInput
  }
}

param_control.logical <- function(param) { shiny::checkboxInput }
param_control.Date <- function(param) { shiny::dateInput }
param_control.POSIXct <- function(param) { shiny::textInput }
param_control.character <- function(param) { shiny::textInput }
param_control.data.frame <- function(param) { shiny::fileInput }

# here's a fake invocation which selects the right function for data.frame
param_control(list(class = "data.frame"))
```

R/params.R
```diff
@@ -0,0 +1,329 @@
...
+#'
+#' @param file Path to the R Markdown document with configurable parameters.
+#' @param input_lines Content of the R Markdown document. If \code{NULL}, the contents of \code{file} will be read.
+#' @param params A named list of optional parameter overrides used in place of the document defaults.
+#' @param shiny_args Additional arguments to \code{\link[shiny:runApp]{runApp}}.
+#'
+#' @return named list with overridden parameter names and value.
+#' 
+#' @export
+knit_params_ask <- function(file = "index.Rmd",
```
jjallaire: I don't think we should default the `file` argument (there's no strong convention around using `index.Rmd` as a filename so this will seem random, less confusing to force the specification I think).

R/params.R
```diff
@@ -0,0 +1,329 @@
...
+    }
+  }
+  
+  ## Return the current value for a given parameter, which is either `params`
+  ## when overridden by our caller or `param$value` otherwise.
+  current_value <- function(param) {
+    if (!is.null(params)) {
+      override <- params[[param$name]]
+      if (!is.null(override)) {
+        return(override)
```
jjallaire: There's a pretty strong convention (which I honestly had a hard to time getting used to) to never use the `return` function unless you are doing an explicit early return. In this case you therefore wouldn't need it.

R/params.R
```diff
@@ -0,0 +1,329 @@
...
+  ## The default values given to the UI. Lets us detect if the user has
+  ## specified an override value or is letting it float to whatever default is
+  ## generated by the YAML (most useful for R expressions).
+  ##
+  ## Only items in `configurable` will be present here.
+  uidefaults <- list()
+
+  server <- function(input, output) {
+    param.ui <- function(param) {
+      inputId <- paste0(param$name)
```
jjallaire: Why `paste0` here?

R/params.R
```diff
@@ -0,0 +1,329 @@
...
+      shiny::stopApp(values())
+    })
+
+    shiny::observeEvent(input$cancel, {
+      shiny::stopApp(NULL)
+    })
+  }
+
+  contents <- shiny::tagList(
+      shiny::fluidRow(shiny::column(12,shiny::tags$h1("Configure Report Parameters"))),
```
jjallaire: Maybe use `h2` or `h3` here?

R/params.R
jjallaire: I think I'm wrong about how to do the dispatch here though (as my example might end up dispatching on the fake object). @wch We have a scenario where we have a character vector of class names and need to use that to select a method to dispatch on an object (i.e. the object has a class field we use for the dispatch). What's the best way to do this?

aronatkins: The `param$class` comes directly from `param$value`. We could try dispatching on `param$value` instead of using `param$class` at all...

R/params.R
aronatkins: Done.

R/params.R
aronatkins: This one (for `override`) is an explicit early return to prevent the final return of `param$value`. I used the `return` form for `param$value` to make both returned values explicit (it seemed to read better) but I can eliminate the return call for the default `param$value`. Is that your preference?

R/params.R
aronatkins: I had some problems with labels in early versions of this, and believed that this helped. But I can't see why any longer. Removing.

R/params.R
aronatkins: Stepped down this heading and the "cannot be configured" heading below.

jjallaire: The problem is that param$value can be NULL (which is why we have a special
param$class rather than using class(param$value). We may want to talk one
more time about whether allowing NULL is necessary/desirable)

jjallaire:
Some additional feedback based on getting this working in the IDE. Here's what it currently looks like: 

screen shot 2015-07-21 at 1 36 46 pm

A few things it would be nice to tweak:

1) Eliminate the header (since we already have a dialog header)

2) Right align the buttons and have some more space between them

3) Change the caption to "Knit" from "Save"

(3) Could obviously be done as a parameter to the function (e.g. `save_caption = "Save"`). 

A more sticky issue is keyboard dismissal of the dialog. When the Shiny app has input focus I'm unable to trap the Esc/Enter keys to dismiss/accept the dialog (all that is required is that the Shiny app terminates as I subscribe to the "disconnect" message to dismiss the dialog). @jcheng5  Is there any way for a Shiny app to handle Esc and Enter for this sort of use case?

jcheng5:
> Is there any way for a Shiny app to handle Esc and Enter for this sort of use case?

I think just use custom JS/jQuery to trap Esc and Enter at the page level and send a click to Cancel or Save, that ought to work.

aronatkins:
Ok. The requested UI tweaks are complete and we also have Enter/Escape handlers.

aronatkins:
Requires https://github.com/yihui/knitr/pull/1078

R/params.R
```diff
@@ -0,0 +1,355 @@
...
+  # Maps between value types and input: XXX
+  default_inputs <- list(
+      logical = "checkbox",
+      ## BUG: dateInput does not allow the user to not specify a value.
+      ##     https://github.com/rstudio/shiny/issues/896
+      Date = "date",
+      ## BUG: shiny does not support datetime selectors
+      ##     https://github.com/rstudio/shiny/issues/897
+      ##     we ask for string input for now.
+      POSIXct = "datetime",
```
jcheng5: Should this be `POSIXt`, so that `POSIXlt` is included as well?

jcheng5: I agree that Shiny ought to have datetime input, but in general, don't feel like you need to limit yourself to the inputs that Shiny supports out of the box--if you want to bundle your own custom controls with rmarkdown that would be fine IMHO.

jjallaire:
One more tweak I think we need -- the Rmd resource discovery function needs to be enhanced to discover file paths referenced in `value` when there is an `input: file`. Here's the function that would need to be modified: https://github.com/rstudio/rmarkdown/blob/master/R/html_resources.R#L162

aronatkins:
That's good (re: resource discovery). Now that we know something is a filename, we can process it. More what you would expect compared to the old `data.frame` processing.

aronatkins:
re: datetime picker - I was under the impression that we didn't want to burden `rmarkdown` with third-party JS components. My preference is to rely on core Shiny. If a date-time picker does not belong in Shiny, then maybe this is one of those exceptions.

PR #485, opened 2015-07-28
Allow YAML blocks to end with periods

noamross:
YAML blocks can end with `...` when closing ([Ref](http://www.yaml.org/spec/1.2/spec.html#id2760395)).  While Jekyll ends blocks with `---`, pandoc prefers `...`.  This change allow header blocks to end with either.  This means that rmarkdown will parse markdown files generated by pandoc (via `pandoc -t markdown -s`, which retains header metadata), and pandoc will parse *.Rmd files and retain their metadata.

### DIFF
--- a/R/output_format.R
+++ b/R/output_format.R
@@ -520,7 +520,9 @@ validate_front_matter <- function(front_matter) {
 partition_yaml_front_matter <- function(input_lines) {
 
   validate_front_matter <- function(delimiters) {
-    if (length(delimiters) >= 2 && (delimiters[2] - delimiters[1] > 1)) {
+    if (length(delimiters) >= 2 &&
+        (delimiters[2] - delimiters[1] > 1) &&
+        grepl("^---\\s*$", input_lines[delimiters[1]])) {
       # verify that it's truly front matter (not preceded by other content)
       if (delimiters[1] == 1)
         TRUE
@@ -532,7 +534,7 @@ partition_yaml_front_matter <- function(input_lines) {
   }
 
   # is there yaml front matter?
-  delimiters <- grep("^---\\s*$", input_lines)
+  delimiters <- grep("^(---|\\.\\.\\.)\\s*$", input_lines)
   if (validate_front_matter(delimiters)) {
 
     front_matter <- input_lines[(delimiters[1]):(delimiters[2])]


jjallaire:
@jmcphers I checked and it looks like all the places in the IDE where we parse YAML from R use the codepath that has been modified. Do we do any YAML extraction in JavaScript that would need to be updated to handle the terminating `...`?

Noam, could you send us a contributor agreement (if you haven't already, I couldn't find one in our internal database). See details here: https://github.com/rstudio/rmarkdown/blob/master/CONTRIBUTING.md

jmcphers:
Yes, we do YAML extraction with JavaScript, and yes, it would need to be updated. Here's where we presume that the YAML block ends with `---`:

https://github.com/rstudio/rstudio/blob/master/src/gwt/src/org/rstudio/studio/client/rmarkdown/model/YamlFrontMatter.java#L35

Happy to fix this up if needed.

jjallaire:
Okay, yes, let's change this to allow for --- or ...

Once that's done let me know and I'll take the rmarkdown PR

One last question though: if an older version of RStudio encounters front
matter delimited by ... what will happen? Is it just thinking there is no
front matter or would something (e.g. YAML options editing) break entirely?

jmcphers:
Change has been made:

https://github.com/rstudio/rstudio/commit/72c2328fdc0dadc538301e8e76bcf87bb3f8653b

An older version will act as though there's no front matter at all if it can't find the end of the front matter. Nothing will break, but operations that act on the front matter will generally just create a new front matter block at the top of the doc with the recognized delimiters.

aronatkins:
`rsconnect` parses front-matter: https://github.com/rstudio/rsconnect/blob/e01048deb08f9a0933d13e54903694f48d481092/R/bundle.R#L85
as does `knitr`: https://github.com/yihui/knitr/blob/1adf407c8ba9dd6c11b3a7d2c5f13c52e8f09672/R/params.R#L150

jjallaire:
@aronatkins Great catch! 

@noamross I'm amenable to this change however it does seem like it's going to break some things (at least until these other packages are updated). I guess however since it's opt-in and at first won't be widely known / documented that the breakage will be pretty limited in impact. 

Could you also put PRs into rsconnect and knitr that compliment this one?

aronatkins:
And here's the `packrat` parsing: https://github.com/rstudio/packrat/blob/17c894d4da7586d90362829033147487fd8875c8/R/dependencies.R#L132

jjallaire:
@noamross Note that packrat also would require a PR for this as pointed out by @aronatkins

jjallaire:
@noamross The v0.99.636 daily build of RStudio also includes a couple of changes related to ... (the options editing UI should respect that delimiter and code folding should respect that delimiter).

PR #495, opened 2015-08-12
always emit a named list

aronatkins:
Returning a named list, even when empty due to a lack of parameter customization, allows for consistent downstream serialization. If we instead returned `list()`, `jsonlite::toJSON` would serialize as `[]` and not `{}`; meaning a different JSON root data structure.

https://github.com/jeroenooms/jsonlite/blob/777c7215a0d77085065d70299132287cd1392fca/R/asJSON.list.R#L38

I am playing with consuming the output from `rmarkdown::knit_params_ask` and it felt like this change would allow more consistent downstream processing. It's possible that I will throw away `length(overrides)==0` output, but if it hangs around, at least it will always serialize as a JSON map.

Are there other/better ways of accomplishing this?

### DIFF
--- a/R/params.R
+++ b/R/params.R
@@ -175,6 +175,13 @@ params_configurable <- function(param) {
   length(param$value) <= 1 && !is.null(params_get_control(param))
 }
 
+# Returns a new empty named list.
+params_namedList <- function() {
+  empty <- list()
+  names(empty) <- character()
+  empty
+}
+
 #' Run a shiny application asking for parameter configuration for the given document.
 #'
 #' @param file Path to the R Markdown document with configurable parameters.
@@ -217,10 +224,10 @@ knit_params_ask <- function(file = NULL,
     ## version.
   }
 
-  ## If we happen to not have any knit_params, just return an empty list and
-  ## don't fire up the Shiny app.
+  ## If we happen to not have any knit_params, just return an empty named list
+  ## and don't fire up the Shiny app.
   if (length(knit_params) == 0) {
-    return(list())
+    return(params_namedList())
   }
   
   configurable <- Filter(params_configurable, knit_params)
@@ -229,7 +236,7 @@ knit_params_ask <- function(file = NULL,
   ## This set of published values is the raw set that came from the user.
   ## It does not include those values that cannot be configured or are
   ## left to use the default.
-  values <- list()
+  values <- params_namedList()
 
   server <- function(input, output) {
     param.ui <- function(param) {


R/params.R
```diff
@@ -175,6 +175,13 @@ params_configurable <- function(param) {
   length(param$value) <= 1 && !is.null(params_get_control(param))
 }
 
+# Returns a new empty named list.
+params_namedList <- function() {
+  empty <- list()
+  names(empty) <- character()
+  empty
+}
```
yihui: FYI, there is a `setNames()` function, e.g. you can just return `setNames(list(), character())`.

aronatkins: Is one approach preferred over the other? Should I switch?

yihui: No. What you did here is basically what setNames() does (take a look at its source code -- it is only two lines). As I said, it is only FYI. It is no big deal to reinvent a wheel of two lines :)

hadley:
This seems like a pretty nasty json serialisation issue to me. But maybe it's impossible to be more elegant

aronatkins:
I initially thought it was a JSON serialization issue as well, but then consider:

```
```

aronatkins:
I initially thought it was a JSON serialization issue as well, but then consider:

```
# very natural
list("x","y") # ==> ["x", "y"]
# also makes sense
list(a=1, b=2) # ==> { "a": 1, "b": 2 }
```

Thus, `list()` is ambiguous, as it maps to both a JSON array and map depending on how it is used.

I suppose one could argue that `list()` should always be materialized as a JSON map:

```
list("x","y") # ==> { "1": "x", "2": "y" }
```

but that's a little bizarre. Especially those are string keys and Javascript arrays are zero-based.

RJSONIO even has `RJSONIO::emptyNamedList`. Oy.

hadley:
Yeah, seems like this is just one of those fundamental ambiguities of translating R lists to JSON :/

R/params.R
hadley: I also appreciate the combination of snake case and camel case. That's how you make everyone happy :P

yihui: Haha. The naming convention world is so noisy that IdidNot_even_NOTICEit!

aronatkins: Wow. I am awesome. My brain didn't even pick up on that. I was, "oh. we prefixed all the params stuff with `params_` so i'll do that" and then, "oh. RJSONIO has `emptyNamedList` so i'll follow their lead".

hadley: Achievement unlocked: you are now such an experienced R programmer that you don't notice when using two or more coding styles in one function.

aronatkins: This should be the next RStudio sticker.

Issue #348, opened 2015-01-09
highlight.js reference in html document not using relative path


jrowen:
I'm attempting to create a website html document (see description [here](http://rmarkdown.rstudio.com/html_document_format.html#creating-a-website)) using the code below.  In the resulting html file I'm not seeing any code highlight and finding path errors related to the `highlight.js` library.  Looking at the html source, it doesn't appear that the relative path is being generated correctly.

Here is the R code used to render the html document

```
out_nm = "ISM_class_model"
out_path = paste0("~/ShinyApps/markdown/", out_nm)
rmarkdown::render("markdown/ModelResult.Rmd", envir = e1,
                  output_format = "html_document",
                  output_file = paste0(out_nm, ".html"),
                  output_dir = out_path,
                  output_options = list(self_contained = FALSE,
                                                   lib_dir = paste0(out_path, "/lib"),
                                                   css = "../include/htmlcustom.css"))
```

and this is the chunk at the top of the Rmd file

```

---
title: 'ISM Model'
output:
  html_document:
    css: ../templates/htmlcustom.css
    includes:
      in_header: ../templates/header.html

---
```

Here's a subset of the output file's header that includes the path error.  The path for the other external libraries is being set correctly.

```
<script src="lib/jquery-1.11.0/jquery.min.js"></script>
<meta name="viewport" content="width=device-width, initial-scale=1" />
<link href="lib/bootstrap-3.3.1/css/bootstrap.min.css" rel="stylesheet" />
<script src="lib/bootstrap-3.3.1/js/bootstrap.min.js"></script>
<script src="lib/bootstrap-3.3.1/shim/html5shiv.min.js"></script>
<script src="lib/bootstrap-3.3.1/shim/respond.min.js"></script>
<link rel="shortcut icon" href="images/favicon.ico" type="image/x-icon">
<link rel="icon" href="images/favicon.ico" type="image/x-icon">

<style type="text/css">code{white-space: pre;}</style>
<link rel="stylesheet"
      href="~/ShinyApps/markdown/ISM_class_model/lib/highlight/default.css"
      type="text/css" />
<script src="~/ShinyApps/markdown/ISM_class_model/lib/highlight/highlight.js"></script>
```

If I specify a highlight theme (e.g. `highlight: tango`), the error disappears, as the reference to highlight.js in the header is not in the output html.

yihui:
[...]

Issue #238, opened 2014-08-27
InvalidUrlException when absolute path specified for output_dir on Windows


jrowen:
When calling `render` and an specifying a absolute path for `output_dir` on a Windows machine

```
rmarkdown::render("c:/data/test.Rmd", output_dir = "c:/temp")
```

an `InvalidUrlException` exception is returned

```
pandoc.exe: Failed to retrieve c:/temp/test_files/figure-html/unnamed-chunk-2.png
InvalidUrlException "c:/temp/test_files/figure-html/unnamed-chunk-2.png" "Invalid scheme"
```

Relative paths work fine.

yihui:
In general I'd recommend the `file.rename()` approach. My experience is that the `output_dir` and `output_file` arguments are quite fragile and should be avoided if possible. I have seen problems several times, e.g. it just happened yesterday at https://github.com/rstudio/shiny/issues/1831.

PR #505, opened 2015-09-04
New byproducts_dir argument for render()

lionel-:
This adds a `byproducts_dir` argument to `render()` that allows to specify where to write the cache and figures folders.

This is useful to avoid cluttering the project folder.


jjallaire:
Thanks for this. There are *many* downstream assumptions about file locations that this has the potential to break (it might not break any, but it very likely could) so we will need some time to analyze and test before considering the PR. In the meantime I'd suggest just running your local fork assuming it works for the things you are trying to accomplish.

lionel-:
No problem. Thanks for the quick heads up.

yihui:
I think we have had enough headache with these three arguments: `output_file`, `output_dir`, and `intermediate_dir` (e.g. https://github.com/rstudio/rmarkdown/issues/587#issuecomment-168437646 #238 #348 #574 #1033 #861 etc). I tend not to add yet another argument to deal with output directories.

PR #506, opened 2015-09-04
Fixes #283.

### LINKED ISSUE #283: Support `opts_template`

jamiefolson:
Adds support for knitr option templates in the knitr options.

### DIFF
--- a/R/output_format.R
+++ b/R/output_format.R
@@ -131,8 +131,10 @@ merge_output_formats <- function(base, overlay)  {
 #'   \code{\link[knitr:opts_chunk]{opts_chunk}})
 #' @param knit_hooks List of hooks for R code chunks, inline R code, and output
 #'   (see \code{\link[knitr:knit_hooks]{knit_hooks}})
-#'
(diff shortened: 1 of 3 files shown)


jjallaire:
@yihui  does this look okay to you?

yihui:
@jjallaire Yes, it does. 

@jamiefolson It will be nice if you can avoid white space changes in the future to make the patch a little cleaner. You probably also need to upgrade your roxygen2 package. Thanks!

jjallaire:
@jamiefolson If we don't already have one could you send us a contributor agreement as described here: https://github.com/rstudio/rmarkdown/blob/master/CONTRIBUTING.md

jamiefolson:
It looks like I'm using the current version of Roxygen2 (4.1.1).  I'm not sure where the extra whitespace in the Rd files came from.  You should have a contributor agreement from last fall.

PR #527, opened 2015-10-07
Do not inline bootstrap.min.css when not self-contained

yihui:
There is no need to inline it when we don't want the output to be self-contained.

BTW, since the bug https://github.com/jgm/pandoc/issues/2224 seems to have been fixed, I'm wondering if we still need the workaround anyway.


jjallaire:
You are right that we longer need to do this. Fixed here: https://github.com/rstudio/rmarkdown/commit/8036e25c6a03d130023768cf1f5ad294465b6a74

PR #536, opened 2015-10-21
Use natbib or biblatex to process citations when keep_tex = TRUE

yihui:
This PR was motivated from https://github.com/hadley/ggplot2-book/issues/96. LaTeX has its own solid tools to process citations (e.g. `bibtex`), and `pandoc-citeproc` does not seem to be very reliable. When a user wants to generate the tex output, we use the `--natbib` or `--biblatex` option instead of `--filter pandoc-citeproc`. In this case, I didn't figure out how I can convert the tex document to PDF using Pandoc (it seems Pandoc always parses the tex document and creates a new tex document then tries to compile it to PDF, and I don't want it to do that), so I invented a helper function `latexmk()`. Let me figure out if Pandoc has a built-in way to convert LaTeX directly to PDF without going through the steps `tex -> tex -> PDF`. Please hold on for now.


R/pdf_document.R
```diff
@@ -13,6 +13,8 @@
 #' @param keep_tex Keep the intermediate tex file used in the conversion to PDF
 #' @param latex_engine LaTeX engine for producing PDF output. Options are
 #'   "pdflatex", "lualatex", and "xelatex".
+#' @param citation_package The LaTeX package to process citations, \code{natbib}
+#'   or \code{biblatex}.
```
jjallaire: Should we note that this only applies when keep_tex is TRUE?

jjallaire:
Would we have any motivation or desire to also do this when keep_tex is FALSE or would that be overstepping?

yihui:
Okay, I asked the question in the Pandoc mailing list, and @jgm told me Pandoc didn't have anything fancy to process LaTeX bibliography or index. It simply calls pdflatex/xelatex/lualatex to compile tex to PDF. Now I'm confident to use my `latexmk()` function to compile tex to PDF. The last thing I need to complete this PR is a way to parse and show the LaTeX error message. I guess you have done the work before when working on Sweave. Is there anything I can use to parse the LaTeX log and extract the error message(s)?

jjallaire:
The LaTeX log parser we have is a non-trivial piece of C++ code:

https://github.com/rstudio/rstudio/blob/master/src/cpp/core/tex/TexLogParser.cpp

It doesn't have too many dependencies on the rest of our codebase so it
could be pulled out but this would take some effort.

Here's another implementation written in Python for a Sublime Text plugin:

https://github.com/SublimeText/LaTeXTools/blob/master/parseTeXlog.py

And a JavaScript implementation:

https://github.com/jpallen/latex-log-parser

And a thread that has some more leads:

http://tex.stackexchange.com/questions/10558/use-latexmk-to-filter-the-log-file

Porting some variation of one of these to R might be the best bet. Let's
discuss realtime if need be.

yihui:
I have added a simple helper function `show_latex_error()` to find the error messages (lines that start with `!`). I have tested this PR on OS X/Ubuntu/Windows with/without latexmk, and it seems to be working well. We can certainly improve the `latexmk()` and `show_latex_error()` functions in the future if necessary.

One small caveat is that the `biblatex` package does not really work by default if a .bib database is used, because we need to `\usepackage[backend=bibtex]{biblatex}`, but this `backend` option was not set in Pandoc's template. The default backend for `biblatex` is `biber`, which is less widely known. We may change our LaTeX template, but I guess it is not worth the hassle for now and `natbib` should be much more popular among LaTeX users.

jjallaire:
Okay, is this good to merge from your standpoint?

yihui:
Yes.

PR #541, opened 2015-10-28
--natbib does not work with the default beamer template

yihui:
I discovered this issue yesterday and finally figured out the culprit: https://github.com/jgm/pandoc-templates/blob/de2e524e7df0a9d4159f2794deaa2e1847c87d46/default.beamer#L124-L140

I included the default beamer template and stripped off the problematic LaTeX commands (actually I don't see the difference with/without them).

### DIFF
--- a/R/beamer_presentation.R
+++ b/R/beamer_presentation.R
@@ -79,8 +79,10 @@ beamer_presentation <- function(toc = FALSE,
   args <- c()
 
   # template path and assets
-  if (!is.null(template) && !identical(template, "default"))
+  if (!is.null(template)) {
+    if (identical(template, "default")) template <- patch_beamer_template()
     args <- c(args, "--template", pandoc_path_arg(template))
+  }
 
   # table of contents
   if (toc)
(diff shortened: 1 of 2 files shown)


R/beamer_presentation.R
```diff
@@ -79,8 +79,11 @@ beamer_presentation <- function(toc = FALSE,
   args <- c()
 
   # template path and assets
-  if (!is.null(template) && !identical(template, "default"))
+  if (!is.null(template)) {
+    if (identical(template, "default"))
+      template <- rmarkdown_system_file("rmd/beamer/default.beamer")
     args <- c(args, "--template", pandoc_path_arg(template))
+  }
```
jjallaire: The problem with having our own copy of the default template is getting out of sync with the built-in pandoc templates (e.g .when pandoc starts emitting LaTeX that requires new packages that are included in templates updated along with pandoc). See this commit: https://github.com/rstudio/rmarkdown/commit/82c1dd98c6125f740fd7e2ae57bbaad7c75a10d4

Possibly forking the latest Beamer template gives us the most protection but it doesn't make things 100% future proof.

yihui:
I removed the template from the package, and applied a monkey patch to the default beamer template instead. Not sure if that makes it more future proof. In the meantime, I have also submitted a fix to pandoc: https://github.com/jgm/pandoc-templates/pull/145

PR #554, opened 2015-11-16
Bad UI element argument lists do not stop Shiny configuration app

aronatkins:
If a parameter block has keys that are incompatible with the input control (either inferred or explicitly chosen), note that in the UI instead of blindly running the UI builder.

screen shot 2015-11-16 at 10 32 38 am

This prevents the most likely cause for configuration app startup failures.

### DIFF
--- a/R/params.R
+++ b/R/params.R
@@ -291,8 +291,19 @@ knit_params_ask <- function(file = NULL,
       hasDefaultValue <- function(value) {
         identical(uidefault, value)
       }
-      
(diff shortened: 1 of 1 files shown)


aronatkins:
Here is the Rmd I used for the above screenshot:

```
---
params:
  rows: 24
  region:
    value: "east"
    choices: ["east","west","north","south"]
    multiple: True
  black: "white"
---

The region is: `r params$region`
```

Issue #516, opened 2015-09-20
rendering to md_document results in error if code generates HTML


dtenenba:
Hi,

I have a document `test.Rmd` like this:

```
---
output:
   md_document
---

# test

```{r}
library(DT)
datatable(cars)
    ```
```

(just pretend that last line is in the first column; i couldn't figure out how to escape the backticks)

When I render it with:

```
render("test.Rmd")
```

I get

```
Error: Functions that produce HTML output found in document targeting markdown_strict output.
Please change the output type of this document to HTML.
```

Maybe this is a feature, however, isn't HTML valid in markdown? In my particular use case, I am later going to convert the resulting markdown file to HTML using my own tools, but I really need markdown as the output of `rmarkdown::render()`. It would be great if there were a way to override this error and have rmarkdown go ahead and write the html to the output document.

@vjcitn

@vobencha

jjallaire:
The problem is that besides the generated HTML an htmlwidget like DT has additional dependencies (JavaScript and CSS) which in HTML rendering is automatically added to the document but which obviously can't be for markdown. Any publishing of your markdown will require manually reconstructing these dependencies (not impossible, but tricky, and not something we wanted to just leave as an undocumented exercise to the reader).

I created this branch which includes a workaround (an extra parameter to md_document that says it's okay to have HTML with dependencies): https://github.com/rstudio/rmarkdown/tree/feature/md-allow-html-dependencies. Not sure if this is the right answer though so probably won't merge this just yet.

dtenenba:
Great! How do I add the allow_html_dependencies option? I tried

```
options(allow_html_dependencies=TRUE)
```

But that didn't work.

jjallaire:
It's an md_document option, e.g.

---
output:
   md_document:
      allow_html_dependencies: true
---

dtenenba:
OK, I tried that but I was hoping the resulting md document would contain the required script and css tags. 

I don't understand why these "obviously can't" be added to the document for markdown. HTML is valid markdown and in my case the document will eventually be converted to HTML.

Can the behavior be modified so that if allow_html_dependencies is TRUE, then all relevant html is copied to the output document?

jjallaire:
It *can* be modified but it's unfortunately not a simple change and has some chance of breaking things so we aren't in a position to do that right now. Another possible workaround for you is this:

    ---
    output:
       html_document:
          self_contained: false
          keep_md: true
    ---

[...]

PR #559, opened 2015-11-30
Allow option to render html in non-html output

skranz:
Pull request for the discussion in issue https://github.com/rstudio/rmarkdown/issues/516

If in the yaml front matter the option

always_allow_html: yes

is set, then no longer an error is thrown if a rmarkdown file emits html
output in some chunk but the file is rendered to non-html output like
PDF or word.

### DIFF
--- a/R/render.R
+++ b/R/render.R
@@ -337,9 +337,15 @@ render <- function(input,
     if (!(is_pandoc_to_html(output_format$pandoc) ||
           identical(tolower(tools::file_ext(output_file)), "html")))  {
       if (has_html_dependencies(knit_meta)) {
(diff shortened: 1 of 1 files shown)


jjallaire:
Thanks!

nateGeorge:
Can you give an example of how this is supposed to look in the header?  Is it supposed to be like:

```
---
output:
  pdf_document
  always_allow_html: yes
---
```

or


```
---
output:  pdf_document
always_allow_html: yes
---
```

??

nateGeorge:
It should be the second version:
```
---
output:  pdf_document
always_allow_html: yes
---
```

PR #569, opened 2015-12-02
Retain tex source in pdf conversion when keep_tex = TRUE

aoles:
Includes the fix f8a1783 for running intermediate conversion to tex when `keep_tex = TRUE`, and 49ea937 to capture and merge all pandoc options when defining or extending formats.

### DIFF
--- a/R/output_format.R
+++ b/R/output_format.R
@@ -103,10 +103,7 @@ merge_post_processors <- function (base, overlay) {
 merge_output_formats <- function(base, overlay)  {
   structure(list(
     knitr = merge_lists(base$knitr, overlay$knitr),
-    pandoc = pandoc_options(
-      to = merge_scalar(base$pandoc$to, overlay$pandoc$to),
-      from = merge_scalar(base$pandoc$from, overlay$pandoc$from),
-      args = c(base$pandoc$args, overlay$pandoc$args)),
+    pandoc = merge_lists(base$pandoc, overlay$pandoc),
     keep_md =
       merge_scalar(base$keep_md, overlay$keep_md),
     clean_supporting =
--- a/R/output_format.R
+++ b/R/output_format.R
@@ -103,7 +103,10 @@ merge_post_processors <- function (base, overlay) {
 merge_output_formats <- function(base, overlay)  {
   structure(list(
     knitr = merge_lists(base$knitr, overlay$knitr),
-    pandoc = merge_lists(base$pandoc, overlay$pandoc),
+    pandoc = pandoc_options(
+      to = merge_scalar(base$pandoc$to, overlay$pandoc$to),
+      from = merge_scalar(base$pandoc$from, overlay$pandoc$from),
+      args = c(base$pandoc$args, overlay$pandoc$args)),
     keep_md =
       merge_scalar(base$keep_md, overlay$keep_md),
     clean_supporting =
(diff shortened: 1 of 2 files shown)


jjallaire:
@yihui Could you review this?

R/render.R
```diff
@@ -391,8 +391,10 @@ render <- function(input,
   perf_timer_start("pandoc")
 
   # compile Rmd to tex when we need to generate --bibliography
-  if (grepl('[.](pdf|tex)$', output_file) &&
-      ('--bibliography' %in% output_format$pandoc$args)) {
+  # or when keep_tex is TRUE
```
yihui: This commit does not appear to be necessary to me. I don't see what problem you are trying to solve here. When `keep_tex = TRUE`, you will get get a .tex file.

R/output_format.R
```diff
@@ -103,10 +103,7 @@ merge_post_processors <- function (base, overlay) {
 merge_output_formats <- function(base, overlay)  {
   structure(list(
     knitr = merge_lists(base$knitr, overlay$knitr),
-    pandoc = pandoc_options(
-      to = merge_scalar(base$pandoc$to, overlay$pandoc$to),
-      from = merge_scalar(base$pandoc$from, overlay$pandoc$from),
-      args = c(base$pandoc$args, overlay$pandoc$args)),
+    pandoc = merge_lists(base$pandoc, overlay$pandoc),
```
yihui: The new behavior does not seem to be equivalent to the old one when dealing with `args`, so it may be a breaking change, which may be reasonable or not. I'm not sure. Cc @jjallaire

R/render.R
aoles: The problem is that the intermediate tex file is not produced unless `--bibliography` is in `output_format$pandoc$args`, which is not always the case. For example, I run into this issue when compiling the sample rmd file provided by RStudio (File/New File/R Markdown.../Document/PDF) with added `keep_tex = TRUE`.

With rmarkdown release 0.8.1 the tex file is produced as expected.

R/output_format.R
aoles: You're right, I missed that. The issue with the current code is that only a subset of pandoc options is captured, with `keep_tex`, `latex_engine` and `ext` being discarded. I discovered this when implementing a format extending the base rmarkdown `pdf_document`.

I've reverted commit 49ea937 and extended the original approach in e4cf95e.

aoles:
Hi JJ, yihui,

thank you for considering my pull request and for reviewing it. I've addressed yihui's commnets in the inline notes. Briefly, f8a1783 restores the original behavior from release version 0.8.1, and e4cf95e adds the arguments of `pandoc_options()` which were previously missing to `merge_output_formats()`. Both commits are necessary for the option `keep_tex = TRUE` to work as expected in formats extending the base `pdf_document` format.

Cheers,
Andrzej

R/render.R
yihui: I see. That makes sense.

yihui:
Sorry. I have still been thinking about the first commit f8a1783765af8d5c5c8243529aaa4d81ffa5fc87. It did reveal a case that I failed to address, but the fix was not quite correct. I'll put together another PR soon.

jjallaire:
Okay, thanks!

PR #572, opened 2015-12-03
Fixes #555: `output_file` in render() can be a character vector when rendering multiple output formats

### LINKED ISSUE #555: render's document_format does not always use output_file name
If I have an Rmd file named `somefile.Rmd`, with the yaml header

```
---
title: Some title
output: 
    html_document:
        keep_md: true
---
```

Then using `rmarkdown::render('somefile.Rmd', output_file='newname.html')` will reports named `newname.html` and `newname.md`.

However, if I try to render the file using

```r
rmarkdown::render('somefile.Rmd', output_file='newname.html',
  output_format=c("html_document", "md_document")
```

it will produce reports names `somefile.html` and `somefile.md`.

Intuitively, I would have expected the above command to have produced `newname.html` and `newname.md`. Is this the intended behavior? If so, is there a way to rename the output **and** specify multiple output formats when using render? 

I'm happy to try and make changes if someone can point me in the right direction (and I can find some free time!).

(this is with Rmarkdown 0.8.1)

yihui:
e.g. 

```r
render(
  'foo.Rmd',
  c('html_document', 'pdf_document'),
  output_file = c('bar.html', 'bar.pdf')
)
```

### DIFF
--- a/R/render.R
+++ b/R/render.R
@@ -54,6 +54,10 @@ render <- function(input,
                        encoding = encoding)
       outputs <- c(outputs, output)
     }
+    if (length(output_file) > 1) {
+      file.rename(outputs, output_file)
+      outputs <- output_file
+    }
     return(invisible(outputs))
   }
 


R/render.R
```diff
@@ -54,6 +54,10 @@ render <- function(input,
                        encoding = encoding)
       outputs <- c(outputs, output)
     }
+    if (length(output_file) > 1) {
```
jjallaire: Should we verify that the length of output_file is appropriate for the number of output formats? (or does recycling take care of doing the right thing even when the number is different?)

yihui: Good point. I did think about it, and `file.rename()` will signal an error if `length(output_file) != length(outputs)`, so I decided not to add the check here.

PR #573, opened 2015-12-03
Fix after #569

yihui:
- simplified e4cf95e772d5f70b421cc5a342cf003ce7ec7ddb a little bit;
- included a correct fix for the bug I introduced in #536 (when `keep_tex = TRUE` and there is no `bibliography`, the intermediate tex file is not really preserved)


yihui:
@aoles I'll appreciate it if you can test the current development version of rmarkdown after this PR.

aoles:
Thank you @yihui for improving my initial implementation, I confirm that the current development version  works now as expected when `keep_tex = TRUE`.

PR #581, opened 2015-12-28
Tufte CSS

yihui:
This PR is for porting https://edwardtufte.github.io/tufte-css/ into rmarkdown. **It is not ready to be merged yet.** There are still quite a few things to be tweaked in the CSS before it works well with Pandoc.

@jjallaire Since tufte.css has some fonts dependencies, the `tufte_html()` format won't work well when `self_contained = FALSE`. Could you tell me how to declare these dependencies in the format function `tufte_html()`? Ideally, `render()` should copy tufte.css and the fonts (the `et-book` directory) to the output directory when render HTML output that is not self-contained. Thanks!

- [ ] call pandoc to convert markdown in `newthought()` and `marginnote()` to html/latex
- [x] convert references to margin notes
- [x] move margin figures before code blocks so they are placed side by side with code blocks instead of below them
- [x] table captions
- [x] default table styles are ugly


R/tufte_handout.R
```diff
@@ -56,26 +46,50 @@ tufte_handout <- function(fig_width = 4,
...
   format
 }
+
+#' @details \code{tufte_html()} provides the HTML format based on the Tufte CSS:
+#'   \url{https://edwardtufte.github.io/tufte-css/}.
+#' @param css One or more CSS files to include, and the Tufte CSS will always be
+#'   included in this list
+#' @rdname tufte_handout
+#' @export
+tufte_html <- function(css = NULL, ...) {
```
hadley: I think using `character()` instead of `NULL` is slightly better as it suggests the type of vector you should be supplying.

R/tufte_handout.R
```diff
@@ -56,26 +46,50 @@ tufte_handout <- function(fig_width = 4,
...
+#'   (using small caps)
+#' @rdname tufte_handout
+#' @export
+newthought <- function(text) {
+  if (is_html_output()) {
+    sprintf('<span class="newthought">%s</span>', text)
+  } else if (is_latex_output()) {
+    sprintf('\\newthought{%s}', text)
+  } else {
+    warning('newthought() only works for HTML and LaTeX output')
```
hadley: I think it's good practice to do `call. = FALSE` in `warning()`, but I'm not sure what the RStudio style is.

R/util.R
```diff
@@ -350,3 +350,26 @@ show_latex_error <- function(file) {
     stop(e, ' See ', logfile, ' for more info.', call. = FALSE)
   }
 }
+
+# import two helper functions from knitr (with fallbacks if the knitr version is
+# not sufficient)
+is_html_output <- function(...) {
+  fun <- tryCatch(getFromNamespace('is_html_output', 'knitr'), error = function(e) {
```
hadley: I think it would be cleaner if you switched behaviours based on `packageVersion("knitr")`. When you use a `tryCatch()` in this way, I worry about other errors being silently ignored.

R/tufte_handout.R
jjallaire: We don't have a consistent style on warning but I agree that it's good form
to use call. = FALSE.

jjallaire:
A few comments regarding fonts:

1. So long as your fonts are included in the main directory that the html_dependency points to they will in fact be copied over. I typically just include them in the "css" directory if there is one. If you write a tufte_html_dependency function like this one and make sure the fonts are in the main directory then you'll be good to go:

https://github.com/rstudio/rmarkdown/blob/8a8d3d58d57e36647ffd20aa5ec181228cc77b97/R/html_dependencies.R#L13-L25

2. I think you need to reduce the number of embedded fonts. My initial submission of rmarkdown to CRAN was rejected as being too large due to:

- Embedded MathJax (most of which was fonts). I got rid of this entirely.
- Embedded fonts for Bootstrap (I got rid of "woff" and just used "truetype")
- Embedded fonts for Reveal.js (I just removed Reveal.js from the package and hosted it in it's own package).

I think it's good to have this format be part of rmarkdown but we need to do what we can to slim the fonts down. For Bootstrap fonts I went with 100% truetype fonts:

https://github.com/rstudio/rmarkdown/blob/master/inst/rmd/h/bootstrap-3.3.1/css/flatly.min.css

Truetype rather than WOFF was necessary to be compatible with all browsers including FireFox, IE, and especially QtWebKit.

yihui:
@jjallaire So after I specify the Tufte CSS as an HTML dependency, how do I pass this dependency list to `html_document()`? Do I just use the `extra_dependencies` argument of `html_document_base()` (passed to the `...` argument of `html_document()`)?

I'll keep the truetype fonts and remove other types of fonts.

jjallaire:
I think you want your format to call html_document_base:

https://github.com/yihui/rmarkdown/blob/feature/tufte-html/R/html_document_base.R

You would indeed pass the dependency using the extra_dependencies argument
to that function.

jjallaire:
Is there yet an actual Rmd template provided as part of the PR? (i.e. one that shows up in the RStudio New R Markdown dialog and provides basic a basic example usage).  I didn't see it here but perhaps GitHub is obscuring it somehow. 

Also, how do you feel about your changes to tufte_handout in terms of including this PR in the update to be soon sent to CRAN? (i.e. do you think regressions are possible/likely, as we definitely won't want to be submitting again for at least 30 days after the patch).

yihui:
I'm still working on the Rmd template, and I'll include it once it is finished. The output will look like this: https://dl.dropboxusercontent.com/u/15335397/misc/skeleton.html

The changes to `tufte_handout()` are not critical. They can be included in the next major release. Once I figured out a fix to the file deletion bug, we can submit a patch release to CRAN.

jjallaire:
Okay, great! I'll wait on this PR until after we submit to CRAN.

jjallaire:
Let me know once you have skeleton.Rmd in and I'll merge.

yihui:
This PR is basically complete now. Preview of HTML and PDF examples:

- https://dl.dropboxusercontent.com/u/15335397/misc/skeleton.html
- https://dl.dropboxusercontent.com/u/15335397/misc/skeleton.pdf

Note it depends on the current development version of knitr, which I plan to release to CRAN by the end of this week.

I have one issue for discussion, and one question about CSS. The question is, I don't know why the responsiveness does not work on my iPhone, and it does not work with Chrome emulation, either. It only works on my computer screen when the window width is smaller than 760px. This is what I have in tufte.css:

```css
@media screen and (max-width: 760px)
```

For some reason, the original website works well on my phone: https://edwardtufte.github.io/tufte-css/ I don't know what made the difference.

The issue for discussion is whether we should hide the margin figures, figure captions, and table captions in the margin when the page width is small. The original author chose to hide them, but I feel it is not a good idea. I don't like figures or tables without captions by default (you can click the circled pluses to toggle the visibility).

jjallaire:
I agree completely that we shouldn't hide anything for small page widths.

For the CSS issue I'd check with Winston and/or Joe.

BTW I'm updating the embedded pandoc to 1.15.2 so the pandoc-citeproc
version requirement will be met in the upcoming version of RStudio.

jjallaire:
I am wondering if we shouldn't consider pulling both tufte formats and potentially the tufte_ebook format from rticles into a new standalone "tufte" package. The fact that we have introduced some new top-level functions (e.g. newthought) that are tufte format specific is what first got me thinking of this. Note that even if we did this we could still feature them in the IDE New R Markdown dialog (and just do a prompted install of the tufte package when the are selected). What do you think?

yihui:
I think it is a good idea to pull tufte_handout() and tufte_html() into a separate package. I'll do that. Thanks!

jjallaire:
Okay, I think it's fine if it's just called "tufte" but if you want to be
more specific you could consider "tuftedocs" or "tufteformats".

hadley:
Or if you wanted to be humorous, you could call it tufty :wink:

aronatkins:
tufter?

yihui:
Closing this PR. The new package is here: https://github.com/rstudio/tufte Once it is on CRAN, rmarkdown can import and re-export `tufte_handout()` from **tufte**.

PR #582, opened 2015-12-30
Make sure the original input directory is not used as the intermediate directory

yihui:


jjallaire:
This looks great. I will merge but I'd like one more look from @jmcphers before submitting to CRAN (hopefully first thing tomorrow). Jonathn could you look at this before EOD today?

jjallaire:
I went ahead and submitted without further review because I really want that file removal bug off of CRAN. To me the change looked straightforward enough that the changes of it introducing another problem looked extremely slim.

PR #584, opened 2015-12-31
Use normalizePath(winslash = "/") more consistently throughout the package

yihui:
I forgot to consider that `winslash = "\\"` on Windows for `normalizePath()` in #582, so the file deletion bug still exists on Windows. The first commit of this PR should fix this issue. To avoid potential future problems, I changed most instances of `normalizePath()` to `normalize_path()` since we are almost always using `winslash = "/"` throughout the package.


jmcphers:
Looks good to me. It looks like our existing usage of `normalize_path` is now able to emit warnings which were previously suppressed, but they seem appropriate in context.

https://github.com/rstudio/rmarkdown/blob/6cfcada7c5c7ace16b4fdcc875a086b9a94a9ba7/R/html_document.R#L184

yihui:
Good observation. Just added `mustWork = FALSE` there.

jjallaire:
My only concern here is that since this will be our 3rd submission of this package in ~ one week I definitely want no risk whatsoever of us requiring a 4tg submission due to a regression (which we've seen many of in the past due to the subtleties of / vs. \ on different platforms.

How about we confine this PR to the narrowest possible change that fixes the issue on Windows then put the rest of the normalize_path changes in a separate PR for the next release.

yihui:
Sure. I have removed the additional commits, and made the changes absolutely minimal.

jjallaire:
Thanks! I'll submit to CRAN first thing tomorrow. @jmcphers let us know if you have any concerns/amendments/etc

jjallaire:
Thanks, I just re-submitted to CRAN.

We can merge the other changes from a fresh PR once we are onto CRAN.

PR #589, opened 2016-01-07
Add the viewport setting to make the default HTML template work better with mobile browsers

yihui:
cf http://www.w3schools.com/css/css_rwd_viewport.asp https://developer.mozilla.org/en-US/docs/Mozilla/Mobile/Viewport_meta_tag


jjallaire:
@yihui Could you add a NEWS item for this?

PR #594, opened 2016-01-15
Knit an Rmd without running Pandoc

yihui:
If you want to review the code, you may only look at 6de386942680d54762afd9270c433fb6c6a242f8. The second commit was mostly cosmetic changes (indenting code and stripping white spaces).

This PR is to support rendering multiple Rmd's with each Rmd in a new R session in bookdown.


jjallaire:
Why not just have a run_pandoc flag added to the signature for `render`? As it stands if you pass output_format = "all" to render_one it will actually recursively call `render` back (rather than render_one) which is a bit confusing. 

I think adding a flag might be more clear but recognize there might be an important reason for it to be structured as you have done. Let's discuss.

I'd also love to see the whitespace commit happen first so that's all factored out of this diff (you could do one big commit to master than changes the option in .Rproj and then resaves all of the .R files in the package with the new setting).

yihui:
I do not really have an important reason. I was thinking this would only benefit the bookdown package at the moment (as I explained in the email last week), and might not have other use cases. It is indeed more clear to add the flag directly to `render()`, and I'll do that.

I will separate the white space changes in another PR.

jjallaire:
Okay, sounds good!

yihui:
I have moved `run_pandoc` and `knit_meta` to `render()`, and removed `render_one()`.

jjallaire:
I'm wondering if we can get rid of the `knit_meta` parameter as from what I can see it just enables easy appending to an existing knit_meta by the caller at the expense of adding a parameter that isn't very intuitive to understand. It seems like the caller could arrange for another way to accumulate knit_meta.

yihui:
It is possible to accumulate knit_meta without this argument, but it is difficult to render the accumulated knit_meta (e.g. think HTML dependencies) if it is not an argument of render(). I did think about it last week -- the (probably only) destination of knit_meta is the pre-processor. I did not figure out a better way to pass knit_meta to the pre-processor.

jjallaire:
Okay, that makes sense. I'd just change the order of render parameters then:

params = ...,
knit_meta = ....,
environ = ...,
run_pandoc = ...
quiet =
encoding =

yihui:
Done.

PR #595, opened 2016-01-19
wrap long lines of output in html_vignette CSS template

fmichonneau:
I moved the line that takes care of wrapping long lines to also include `code` such that long output fit in the gray box. 
This behavior is more consistent with what the default html template uses (but it breaks words with `word-break: break-all; word-wrap: break-word;`)

Before:
before

After:
after


jjallaire:
@yihui Does this look okay to you? If so feel free to merge (+ add a NEWS item).

PR #600, opened 2016-01-22
Add support for floating table of conents (via tocify) to html_document

jjallaire:
This PR adds a `toc_float` option to `html_document()` which when specified floats the table of contents to the left of the main content (it does this using a 3col / 9col fluid row). The table of contents remains visible even as the document is scrolled. This is all done via the tocify library (http://gregfranko.com/jquery.tocify.js/). 

@jcheng5 and @jmcphers, interested to get your feedback on whether my CSS and use of Bootstrap for this layout are sound (especially the `max-width: none` on the main column, without which the TOC and main column would end up mashing together at wider browser widths).

@jimhester, I know you implemented something like this within knitrBootstrap, let me know if there are subtleties I'm not aware of or additional capabilities that users might want.


jimhester:
I actually ended up removing tocify from knitrBootstrap. I ran into some bugs with the anchor generation code (IIRC it broke when you went from a `<H1>` to `<H3>` without an `<H2>` in between and also did not properly sanitize the anchors in all cases). It also seems to have a MIA developer / maintainer unfortunately. I was able to replace most of the functionality for knitrBootstrap with only a little custom js. https://github.com/jimhester/knitrBootstrap/commit/0e21181961cdd8f941f38be60bf9a9fca8d683e4 is the diff.

The main problem in general with a floating TOC is the behavior when the TOC is longer than the page content. Tocify has the `extendPage` option to work around this, but it is a somewhat hacky solution. I never resolved that problem satisfactorily with knitrBootstrap.

jjallaire:
Okay, thanks for the feedback!

It looks like you can provide a custom function for hash generation so I'm
now using yours:
https://github.com/rstudio/rmarkdown/commit/5b98ee15eae8a079e6b3929be56862cf461bdc35

In terms of requiring the H2 it looks like there is an unclosed bug for
that: https://github.com/gfranko/jquery.tocify.js/issues/65. This is a
bummer but not a complete disaster (we can document the requirement of
having at least an H2).

He is definitely MIA (hasn't updated in ~ 2 years!). That said, he did have
a quite a few releases to get things working well enough before abandoning
it and if you Google for JQuery floating table of contents libraries that
work with Bootstrap it still appears to be the only game in town :-\

He's got enough working well and looking good that I'm not thrilled about
re-implementing. I think I'll keep the use of tocify at a very high level
(i.e. don't expose any of it's underlying options
http://gregfranko.com/jquery.tocify.js/#OptionsAPI) and then if it it's too
sketchy in the field re-implement.

jmcphers:
With the `max-width: none` the lines can get really long on a wide display. I fiddled with this a bit--it's challenging because the TOC float is not really "inside" the column and thus doesn't derive its width from Bootstrap (`position: fixed` elements have width relative to the window). After the tweaks this is the behavior you should see (feedback welcome!):

1200px+ -  fixed width body and TOC (this gives the actual content a similar maximum width to what it had previously, ~940px)

992 - 1200px - TOC and body both scale proportionally, at 3 : 9 

768 - 991px - TOC and body move to a 4 : 8 ratio to reduce odds of TOC headers wrapping; proportional scaling continues

768px - TOC stops floating, moves above body

jjallaire:
Thank you, that's an excellent improvement!

jamesalmeida:
Still not sure how to use this toc_float feature... Has it been added to the readme on how to use it? Is it an option to enable?

jjallaire:
It's in the help topic for html_document. In short:

output:
  html_doucument:
    toc: true
    toc_float: true

awfrankwils:
Hi all,

This is a beautiful new feature. However, my TOC is much longer than the page and I am encountering some scrolling errors with the TOC: When i scroll down the entire HTML page moves, but the TOC remains static, even when I am now viewing content that is below (and out of view of) the items on the TOC list. Likewise, when I hover my cursor over the TOC and try to scroll up or down, the TOC list does not scroll. It will work with the UP and DOWN keys on the keyboard, so it is not completely useless, but not the more intuitive multi-touch pad on my macbook. Is this a known issue?

jjallaire:
You are describing one of the major problems with a floating TOC: if gets
so large that it needs to scroll it become unwieldy. The main think I can
suggest is to change your toc_level so it's smaller and/or to make use of
the collapsed: true option to only show sub-headings when you are inside a
section.

awfrankwils:
Thank you for that clarification JJ!

I will certainly think about my use of the headings in future applications. Right now I am knitting together an HTML report that provides an overview of 506 variables, and am not making use of the subheadings. The TOC is manageable with the UP and DOWN keys, just annoying that the cursor functions are buggered. Do you know if any work is being to correct this problem or are we stuck with it as is?

Cheers,

Andrew

jjallaire:
Our floating TOC is based on tocify (http://gregfranko.com/jquery.tocify.js/)
so any issue would have to addressed/resolved there.

jcheng5:
I haven't looked closely at the floating TOC but I wonder if you could improve the behavior with a simple CSS tweak in your Rmd: set a fixed max-height for the TOC container element and set overflow-y to auto.

PR #601, opened 2016-01-24
more work on toc_float

jjallaire:
- add support for collapsed and smooth_scroll options
- scroll to h1 when targeting first h2

### DIFF
--- a/R/html_document.R
+++ b/R/html_document.R
@@ -7,7 +7,9 @@
 #' @param toc_float \code{TRUE} to float the table of contents to the left of
 #'   the main document content. Note that when using \code{toc_float} you must
 #'   use level 2 headers (if your top level headers are level 3 the table
-#'   of contents won't display).
+#'   of contents won't display). Rather than \code{TRUE} you may also
+#'   pass a list of options that control the behavior of the floating table
+#'   of contents. See the \emph{Floating Table of Contents} section below for details.
 #' @param number_sections \code{TRUE} to number section headings
 #' @param fig_width Default width (in inches) for figures
 #' @param fig_height Default width (in inches) for figures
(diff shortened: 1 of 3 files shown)


davidcarslaw:
This looks very useful - welcome addition for longer documents. I realise you have just added the code but when I try it it only shows level 2 section headings (i.e. ##) and not level 1 (#)...

jjallaire:
That's correct, it shows only level 2 and below by design (Level 1 is
intended for the document title and thus doesn't get an entry in the TOC).

davidcarslaw:
Ah, right - thanks. In that case if I have a document where the toc is true, it gives section number 0.1, 0.2 etc - which seems a bit odd, or am I missing something? Is it that the floating TOC is meant to be used without section numbers and will only show level 2 section titles?

Let me take the opportunity to thank you all for the excellent, high quality software you provide to the community. It really has changed the way people work with R for the better.

jjallaire:
There may be something we can do about the section numbering. Could you
provide a simple example Rmd which I can work from?

davidcarslaw:
Here's a simple example based on the Rmd default template. In this case, # is used for the title (as advised) and the top-level section is ##. Although the advice is toc_depth should be 2, I think it can be 3 and still show ok in the floating toc. My feeling is that a section numbered 1, 2, 3 etc. should be associated with a single # and 1.1, 1.2, 1.3 should be ## etc. I would consider the title to be outside the definition of sections? A toc_depth of 2 would then show 1, 1.1 etc.

I've put a gist [here](https://gist.github.com/davidcarslaw/6a5ca92ef2dc68256e81) as I couldn't work out how to embed markdown in markdown...

jjallaire:
I see what you are saying. I tried playing with the --number-offset pandoc
arg to get something closer to what you were hoping for but that still ends
up numbering the first ## as "1.1" (rather than "1" as I think you were
hoping for). It just seems like pandoc section numbering in HTML flat out
doesn't work very well.

davidcarslaw:
Thanks. Pandoc itself seems OK (version 1.15.0.6). Given a simple markdown document:

```
# My top-level topic

## My first subtopic

## Another subtopic

### A sub-subtopic

## Another top-level topic
```

and then compile `pandoc --number-sections < markdownTest.md > out.html`, I get:

image

which seems OK...

jjallaire:
That's what I get as well. I think the underlying problem results from the
fact that:

1) Pandoc seems to expect that # will be used as a header within HTML
documents (they add class="title" to the document header so it can be
styled distinctly) and it's section numbering behavior is based on this
assumption.

2) In the R Markdown ecosystem we've been using ## as the standard top
level header since forever.

3) The TOC generation library we use (tocify) has a bug which breaks the
TOC if you include a header level in the "selectors" for the TOC and then
it isn't in the document the whole TOC fails to show. This is basically why
we can't just automatically include h1 in the TOC selectors.

So I think the story here just isn't very good for the time being. I'll
keep noodling on whether there is a way out of the box.

davidcarslaw:
Great - many thanks JJ; much appreciated.

jjallaire:
I ended up being able to fix the bug in tocify that prevented enabling h1
selectors for the table of contents. So now you can use # as a header
within your document and everything will work as expected (and you'll get
the section number behavior you want).

davidcarslaw:
Wow - quick work! This is excellent and much appreciated. It really helps to navigate long documents - very slick and effective. I just tested it on my document and I can confirm it works as expected.

jjallaire:
Okay, glad we figured that out!

PR #603, opened 2016-01-25
Extra dependencies in html_document()

yihui:
This is to fix rstudio/tufte#7. `tufte::tufte_html()` added a dependency to `html_document_base()`. This worked before #600 because `html_document()` did not use an explicit `extra_dependencies` argument when calling `html_document_base()`. Now it uses this argument to add tocify dependencies, so the tufte dependencies can no longer be passed to `html_document_base()` via `...` of `html_document()`.

### DIFF
--- a/R/html_document.R
+++ b/R/html_document.R
@@ -58,8 +58,8 @@
 #'   default definition or R Markdown. See the \code{\link{rmarkdown_format}}
 #'   for additional details.
 #' @param pandoc_args Additional command line options to pass to pandoc
-#' @param ... Additional function arguments to pass to the base R Markdown HTML
-#'   output formatter
+#' @param extra_dependencies,... Additional function arguments to pass to the
+#'   base R Markdown HTML output formatter \code{\link{html_document_base}}
 #'
 #' @return R Markdown output format to pass to \code{\link{render}}
 #'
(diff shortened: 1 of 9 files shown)


jjallaire:
My only concern here is that in theory `tufte_html` should be able to call `html_document_base` rather than `html_document`. If there's something we've left out of `html_document_base` that makes it vastly better to call `html_document` then perhaps we should try to fix that. Let me know what you think.

yihui:
Yes, I can call `html_document_base()`, and there are two possible ways to go (both are awkward):

1. Copy the full source code of `html_document`, and add the dependency here: https://github.com/rstudio/rmarkdown/blob/master/R/html_document.R#L256-L262 because I need almost everything from `html_document`, and the the only thing I want to add is an extra dependency;
    - Copying large amounts of source code is not only a bad idea in general, but also introduces problems of having to import internal functions like `html_dependency_tocify` (i.e. triple colons `:::`)
2. Use `html_document()` as the base format, and merge it with an `html_document_base(extra_dependencies = tufte_dependency())`; there are two drawbacks of this way:
    1. The high-level `tufte_html()` function has to sync its arguments with `html_document()`, e.g. when a new argument `toc_float` is added to `html_document()`, `tufte_html()` has to add it as well, otherwise it will be passed to `html_document_base()` via `...` (although in the case of `tufte_html()`, `toc_float` probably does not make much sense since the tufte package does not use Bootstrap);
    2. The pre-processor will generate duplicate items for Pandoc arguments, such as the Mathjax and Bootstrap variables;

So far I have only found myself in need of tweaking one or two parts of `html_document()`. To me, `html_document()` is the more pragmatic "base" format than `html_document_base()`, but it is not very extensible. For this specific case, I think it makes more sense to add the `extra_dependencies` argument to `html_document()`. #600 is a breaking change anyway -- previously one could use the `extra_dependencies` argument in `html_document()`, and now we cannot.

jjallaire:
Okay, that makes perfect, merging now!

PR #623, opened 2016-02-18
Remove the file extension .bib for bibliography files on Windows

yihui:
MikTeX's BibTeX always add the .bib extension to the bibliography filename, even if it has already got the extension

`\bibliography{foo.bib}` in foo.tex will generate `\bibdata{foo.bib}` in `foo.aux` after we run LaTeX, then bibtex in MikTeX will look for `foo.bib.bib`, which does not exist, so bibliography cannot be correctly generated.

It is different in TeXLive, though. Both `\bibdata{foo}` and `\bibdata{foo.bib}` work for `bibtex`. This PR should be safe enough. Even if a Windows user uses TeXLive (this should be rare), bibliography should still work.

### DIFF
--- a/R/render.R
+++ b/R/render.R
@@ -395,12 +395,16 @@ render <- function(input,
 
     perf_timer_stop("pre-processor")
 
+    need_bibtex <- grepl('[.](pdf|tex)$', output_file) &&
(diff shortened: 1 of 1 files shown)


jjallaire:
Could you add a NEWS item for this?

yihui:
Sure.

PR #653, opened 2016-03-23
render_site function for rendering collections of documents within a directory as a website

jjallaire:
This commit introduces the render_site function which provides a standardized (and extensible) mechanism for generating websites from a collection of files within a directory. While it has always been possible to create multi-document websites with R Markdown, this commit addresses the following additional concerns:

- A way for tools to discover that a directory contains a website (via the "site" metadata field in index.Rmd)

- A mechanism for tools to initiate the rendering of a site. By default this is just the rendering of each document within the directory however this can be overridden via custom site generators.

- A standardized way to expose where the output of site generation is written to (e.g. _book or _site) as well as a default name for the site (e.g. directory name for publishing on a server).

Here is an example of the custom site generation function used for bookdown (it does rendering via a _render.R script or a Makefile and knows it's output directory and name from it's internal configuration):

https://github.com/rstudio/bookdown/commit/64ac9a1a24da3049e01cb40c99c2b5d3de6fd95a

@jmcphers @aronatkins @jcheng5 @yihui @hadley @tareefk @trestletech


yihui:
Looks good to me!

R/site.R
```diff
@@ -0,0 +1,191 @@
...
+#' @rdname render_site
+#' @export
+default_site <- function(input, ...) {
+
+  # define render function (use ... to gracefully handle future args)
+  render <- function(output_format = NULL,
+                     quiet = FALSE,
+                     encoding = getOption("encoding"),
+                     ...) {
+    files <- list.files(input, pattern = glob2rx("*.Rmd"))
```
aronatkins: How does this approach interact with the various ways folks may include Rmd inside other Rmd? I don't know if this is the perfect example or all that common, but in just a couple moments of poking around, I found:
http://stackoverflow.com/a/25979000

I also wonder if `default_site` should worry about directory traversal rather than just looking at the current-directory. Although perhaps this is best left to an alternate site generator.

R/site.R
```diff
@@ -0,0 +1,191 @@
...
+  input <- input_as_dir(input)
+
+  # if we have an index.Rmd then check it's yaml for "site:"
+  index <- file.path(input, "index.Rmd")
+  if (file.exists(index)) {
+
+    # read index.Rmd and extract the front matter
+    index_lines <- read_lines_utf8(index, encoding)
+    front_matter <- parse_yaml_front_matter(index_lines)
+    if (!is.null(front_matter) && !is.null(front_matter$site)) {
```
aronatkins: Can `parse_yam_front_matter` ever return null? If yes, then we are missing similar null-checks at the other call locations.

R/site.R
jjallaire: I think it's okay if default_site is really basic. My guess is that a proper site framework will have a way of denoting actual top-level pages vs. includes (via metadata or through some sort of file naming convention). Our hope is to create something along the lines of jekyll which is R centric which would have just such a custom generator.

We could argue that due to these subtleties there should be no `default_site` function at all (to not lead people into a false sense of robustness). I could go either way on this, curious to hear what others think.

R/site.R
jjallaire: That was left over from when I was calling a different function that did return null. Just removed the null check.

R/site.R
jcheng5: Just to clarify, are you hoping (planning) to create jekyll, or hoping that someone else in the community does it?

jjallaire: We are hoping to!

hadley: I'm sure @jjallaire was thinking that we need a fully fledged CMS implementation in R :wink:

jcheng5: It seems awfully limiting that `list.files()` doesn't even get `recursive=TRUE`. I think just having subdirectories and maybe (to @aronatkins' point) excluding files that start with `_` would make `default_site` useful in more situations.

jjallaire: Okay, those are good ideas. I'll implement both.

jjallaire: One issue with `recursive = TRUE` is that there are many things that would be often used in websites (common `_output.yml` that references includes and CSS) that don't work with subdirectories. I was thinking that it would be nice if `default_site` worked for the simple rmarkdown-website template we have (https://github.com/rstudio/rmarkdown-website) and it certainly would with `recursive = TRUE`, however if you actually put Rmd files in subdirectories they won't work with the rest of the site unless we implement some sort of path patching/expanding for `_output.yml`.

jjallaire: Net: Putting Rmd files in subdirectories is broken without additional work by us (i.e. look in the parent for _navbar.html and _output.yml and fixup the paths) or asking the user to create custom per-directory versions of _navbar.html and _output.yml (barf!). So maybe we wait on `recursive = TRUE` until we have a better story for sites with subdirectories (which might end up being aforementioned jekyll type package).

jcheng5: Yeah, patching/expanding would be nice. Those options are schemaless though, right? So we can't just patch/expand automatically, without more information?

yihui: I intentionally avoided `recursive = TRUE` when I was working on bookdown for the same reason. I also excluded those Rmd's that start with `_` by default, so if users want to include child documents, they could use `_` or put them in sub-directories.

R/site.R
```diff
@@ -0,0 +1,191 @@
...
+#'
+#' This will result in the entire site being rendered each time Knit is
+#' executed within RStudio.
+#'
+#' @export
+render_site <- function(input = ".",
+                        output_format = NULL,
+                        quiet = FALSE,
+                        encoding = getOption("encoding"),
+                        ...) {
```
yihui: I think we need an argument `envir = parent.frame()` here, and pass it all the way to the site generator, otherwise `rmarkdown::render()` will be using an environment inside a certain function.

R/site.R
jjallaire: We know about includes and css (which are the most common options that
would refer to files) but we would need a hint of some kind for other
options

PR #655, opened 2016-03-24
set.seed(100) for default_site

jjallaire:
In the websites we've created with rmarkdown to date we've included set.seed(100) in the website Makefile to ensure that html output doesn't constantly churn based on the random number seed (this ends up creating much messier and hard to reason about diffs).

We also include that in our rmarkdown-webstie example: https://github.com/rstudio/rmarkdown-website/blob/master/Makefile#L9

Is there any reason we can think of not to do this here?

### DIFF
--- a/R/site.R
+++ b/R/site.R
@@ -157,11 +157,13 @@ default_site <- function(input, ...) {
     sapply(files, function(x) {
       # we suppress messages so that "Output created" isn't emitted
       # (which could result in RStudio previewing the wrong file)
-      suppressMessages(
+      suppressMessages({
+        set.seed(100) # for stable html output
         rmarkdown::render(x,
                           output_format = output_format,
                           envir = envir,
-                          quiet = quiet))
+                          quiet = quiet)
+      })
     })
   }
 


jjallaire:
@yihui @hadley

yihui:
I think this is fine.

hadley:
Hmmm, I'd worry a little about making it so hidden - it's going to be surprising if you're doing simulations. Could it at least be a parameter so it's documented and possible to turn off?

jjallaire:
Or I wonder if a better way of doing this in general is via .Rprofile
rather than in the Makefile or site generator? Would it work just as well
to do it in .Rprofile?

hadley:
Yeah, I think that makes more sense.

yihui:
I said it was fine because I thought very few people would actually end up using `default_site()` at all. For a whole website, I think the vast majority of users will customize the site generator. I'm okay if `set.seed()` is either included in `default_site()` or not.

jjallaire:
Okay, I'm going to close this PR (I do think default_site will be used frequently for simple 4 or 5 page websites based on html_document).

Issue #661, opened 2016-04-01
ioslides not handling sofbreaks


mtorchiano:
When generating ioslides from rmd soft-breaks (i.e. newlines added just to rearrange content are not handled correctly.

```
---
title: "Bug new line"
output: ioslides_presentation
---
## R Markdown

* If I write a bullet list element
  with a new-line in it is is not rendered correctly
```

In the output the word _element_ and the word _with_ are present **without** a separating space.

yihui:
Fixed by #661.

PR #662, opened 2016-04-01
Handles soft breaks. Resolve issue #661

mtorchiano:
Soft-breaks are added by pandoc when a new-line is found but no real
break is required, e.g. when wrapping long lines in lists.
If not handled they are rendered to empty strings, with the result of
merging the last word of a line with the first of the next one.


jjallaire:
Thanks!

Issue #672, opened 2016-04-15
Set option hooks


aoles:
Is there a way of setting option hooks `knitr::opts_hooks` in rmarkdown documents? I didn't find a corresponding argument to `rmarkdown::knitr_options`.

I would like to change certain chunk options according to the values of other chunk options, as described at http://yihui.name/knitr/hooks/#option_hooks. For example, to achieve the behavior described in the example presented  there.

jjallaire:
You can just use knitr_options$knitr_hooks, e.g.
https://github.com/rstudio/flexdashboard/blob/master/R/flex_dashboard.R#L144-L164

aoles:
I would like to override the default `fig.width` and `fig.height` options for the code chunks which have set a certain `fig.env`. In **knitr** I could achieve this by changing `fig.width` and `fig.height` in `knitr::opts_hooks`. Is there no similar mechanism in **rmarkdown** to change the device dimensions of the current code chunk according to the code chunk options?

jjallaire:
You can just set knitr hooks right inside your Rmd I believe (e.g. in the
setup chunk). @yihui am I missing something there? (i.e. rmarkdown is knitr
so anything you can do in knitr you should be able to do in rmarkdown)

yihui:
@jjallaire You are right. Just do it in the first code chunk in R Markdown, e.g.

    ```{r setup}
    knitr::opts_hooks$set(fig.width = function(options) {
      if (options$fig.width < options$fig.height) {
        options$fig.width = options$fig.height
      }
      options
    })
    ```

This does not really require special support in rmarkdown.

aoles:
@yihui @jjallaire thank you for your help!

I wasn't quite specific before. What I actually would like to achieve is to set `knitr::opts_hooks` in a new output format function which extends `pdf_document`, rather than from a code chunk within the document itself. So far I was able to set things via the interface provided by `output_format`. I was puzzled to find out that all knitr hooks but the `opts_hooks` can be provided in the `knitr` argument to `output_format`. It seems that this deficiency can be easily fixed, see https://github.com/rstudio/rmarkdown/pull/674.

[...]

PR #674, opened 2016-04-15
Add ability to set 'opts_hooks' in knitr_options()

aoles:
Addresses issue https://github.com/rstudio/rmarkdown/issues/672.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -61,7 +61,7 @@ Imports:
     tools,
     utils,
     grDevices,
-    knitr (>= 1.11),
+    knitr (>= 1.12),
     yaml (>= 2.1.5),
     htmltools (>= 0.2.4),
     caTools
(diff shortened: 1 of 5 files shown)


yihui:
Looks perfect to me. Had you brought this PR in the first place, there'd be no question asked in #672 :)

The only caveat is that this would require knitr >= 1.12, which I think is fine.

aoles:
Thanks @yihui , I didn't explain properly my intentions and the actual problem in https://github.com/rstudio/rmarkdown/issues/672 in first place either, so no worries! :) I also didn't expect that the fix would be so easy, as I assumed that there were some reasons not to include `opts_hooks` in `knitr_options`; but after looking into the code the solution seemed to be evident.

jjallaire:
Looks good! A couple of small additions to request:

1) Let's update the knitr version dependency to 1.12

2) Let's add an item to NEWS

aoles:
Thank you @jjallaire! I've made the requested updates.

I've already sent the Contributor Agreement to my previous PR last year (https://github.com/rstudio/rmarkdown/pull/569). My contact details didn't change since then. Do I need to sign a fresh copy or are you fine with the old one?

yihui:
The old copy of the agreement should be fine.

Issue #460, opened 2015-06-27
ioslides_presentation ignores the template option


ignacio82:
I have a presentation with a personalized template

```
output: 
    ioslides_presentation:
       template: ./templates/mytemplate.html
```

But rmarkdown is ignoring that option and using the default template

```
/usr/lib/rstudio-server/bin/pandoc/pandoc /tmp/RtmpP2Yo3D/EdCenter.utf8.md --to html --from markdown+autolink_bare_uris+ascii_identifiers+tex_math_single_backslash-implicit_figures --output /tmp/RtmpP2Yo3D/file6f62a7095b9.html --smart --email-obfuscation none --variable widescreen --variable transition=0 --css styles/MPR_Ioslides_Styles.css --template /usr/local/lib/R/site-library/rmarkdown/rmd/ioslides/default.html --mathjax --variable ioslides-url=file6f62a7095b9_files/ioslides-13.5.1 
/usr/lib/rstudio-server/bin/pandoc/pandoc /tmp/RtmpP2Yo3D/EdCenter.utf8.md --to ioslides_presentation.lua --from markdown+autolink_bare_uris+ascii_identifiers+tex_math_single_backslash-implicit_figures --output /tmp/RtmpP2Yo3D/ioslides-output6f66dbcb94b.html 
```

Is this a bug or am I doing something wrong?

Thanks a lot!

jjallaire:
It's because the template is a property of the ioslides_presentation format
(and therefore needs to be indented below it rather than at the same level).

ignacio82:
@jjallaire it is indented, makes no difference 

[Imgur](http://i.imgur.com/b7qeTgW.png)

jjallaire:
Okay, sorry about the mixup there.

ioslides_presentation doesn't actually support custom templates (it doesn't
have a template argument). This is mostly because ioslides is not a format
natively supported by pandoc so the built in template has a lot of special
sauce.

All of that said, you could workaround this by creating a new format that
just calls ioslides_presentation and then modifies the pandoc args to point
to your template.

ignacio82:
@jjallaire I think http://rmarkdown.rstudio.com/ioslides_presentation_format.html says that ioslide supports template: "You can also replace the underlying pandoc template using the template option:" 

If that is incorrect, can you point me in the right direction to create a new format? If I go that route, will I be able to upload the presentation to shinyapps.io ??

[...]

PR #681, opened 2016-04-21
add template option to ioslides_presentation

kiwiroy:
This has come up before in #460 and FWIW this achieves the result. I recognise that the slides themselves are not affected as the [Lua code handles their generation](https://github.com/rstudio/rmarkdown/issues/460#issuecomment-156434820), but there are other aspects of the template that can be altered - in my case I'm interested in the title slide. 

It would be good to see the [documentation](http://rmarkdown.rstudio.com/ioslides_presentation_format.html#custom_templates) updated to make note of this.

### DIFF
--- a/R/ioslides_presentation.R
+++ b/R/ioslides_presentation.R
@@ -18,8 +18,8 @@ ioslides_presentation <- function(logo = NULL,
                                   keep_md = FALSE,
                                   lib_dir = NULL,
                                   md_extensions = NULL,
-                                  pandoc_args = NULL,
                                   template = NULL,
+                                  pandoc_args = NULL,
                                   ...) {
 
   # base pandoc options for all output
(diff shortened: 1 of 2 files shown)


R/ioslides_presentation.R
```diff
@@ -18,6 +18,7 @@ ioslides_presentation <- function(logo = NULL,
                                   keep_md = FALSE,
                                   lib_dir = NULL,
                                   md_extensions = NULL,
+                                  template = NULL,
```
jjallaire: In the `html_document` format the `template` argument goes right after `mathjax`, could we do the same here?

jjallaire:
As you allude to, the `template` for ioslides is a bit special due to Lua generating the slides. It might be good to have a corresponding PR for the `gh-pages` branch that explains what it takes to create a custom template.

jjallaire:
Do we have a contributor agreement from you yet? (see https://github.com/rstudio/rmarkdown/blob/master/CONTRIBUTING.md).

Issue #230, opened 2014-08-24
add support for ioslides thank you and closing slide


jjallaire:

acoppock:
Hi all, just checking in on this!

yihui:
@acoppock You may consider [the xaringan package](https://github.com/yihui/xaringan) based on remark.js, which is a better maintained JS library. I think it is much more customizable than ioslides (especially in terms of CSS).

I'm closing this issue since it has been left here for too long and I don't think it will be resolved anytime soon. @kiwiroy made an attempt in #687 but the tricky thing was that the CSS classes were added to `<article>` instead of `<section>`.

Issue #195, opened 2014-07-29
Background image


jzen188:
Hi Rstudio team, 
I have a question about background image.

I have included this in my 7th slides when I am trying to use ioslide.presentation option.
slide > slide [data-slide-num="7"]
{
background-color: "black";
background-image: url("figures/xx.jpg");
}

The document I build would be black color but CANT FIND THE PATH of the xx.jpg...
Unlike shiny, we can set a www/ directory and put the things inside.
What can I do to find the figure/xx.jpg like calling it in the document itself?
Thank you.

daftmath:
Since this issue has been open for 2 years and doesn't have any easy fix...here is a **reckless and irresponsible**, but working solution for anyone trying to use a background img but rmarkdown can't find the path.

1) Create a dummy image in your rmarkdown file using regular <img> but set width and height to 0 px like so:
`<img src='bg.jpg' style='width:0px;height:0px;'>`

2) Run app and open in chrome. Right click webpage. Select "Inspect". Hit the "Sources tab.

3) Locate the temporary folder with a weird name. Your dummy image will be inside this folder.

![image](https://cloud.githubusercontent.com/assets/12780936/19042831/40b45e2e-8953-11e6-9b54-0b3d41915315.png)

4) Right click your dummy image and select "Copy Link Address"

5) Use the dummy image's address as your css url background image link. Example below, but your file address and name might be different than mine so follow the steps and don't just copy mine.

```
body {
    background-image: url(file1d1027701f9b_files/bg.jpg);
    background-size: 100% auto;
    background-repeat: no-repeat;
}
```

6. Run the application and rmarkdown will now find your background image. This solution might not be stable or easily reproducible, but it works for me.

ShKlinkenberg:
Jared had a solution but... well see for yourself. See [here](https://www.jaredlander.com/2017/07/fullscreen-background-images-in-ioslides-presentations/).

yihui:
@ShKlinkenberg Yeah, I have seen @jaredlander's solution before. I think once we merge #687, this will be much more straightforward and you don't have to hack at it by yourself any more.

jaredlander:
I look forward to not using my workaround and it looks like #687 even keeps the file self-contained. Plus we won't have to worry about loading in jquery and having it make changes at runtime.

yihui:
Just merged #687, and you can add the attribute `{data-background=???}` after a slide title, e.g.

```
## One Slide {data-background=#ffff00}

blah blah

[...]

Issue #140, opened 2014-06-25
knit preference button does not show up if yaml header already exists


fdschneider:
(Not sure if this actually is an RStudio issue.)

I encounter some difficulties when opening a file that was created with a previous version of RStudio (for a custom html parsing function): The little preferences symbol next to the knit button does not show up!
Apparently this happens with files that already have a yaml header, but no 'output' tag. Once I remove the header or add a minimal output specification, such as this:

```

---
output:
   html_document:

---
```

it reappears.

PR #687, opened 2016-05-02
f/ioslides background

kiwiroy:
## Change

Add more simple syntax for slide backgrounds and slide specific class.

Issues #195 and #230 are addressed, or partially so. The thank you slide is a little tricky as it breaks the mold somewhat.
## Usage

``` markdown
## Slide One {data-background="http://www.prolificbeing.com/wp-content/uploads/2015/02/Hello-World.png"}
## Another Slide | subtitle
Lorem ipsum dolor.
## Final content {slide-class="csstricks"}
`
slide > slide [data-slide-num="7"] {
  background-image: url("figures/xx.jpg");
}
`
## Thanks for listening {slide-class="thank-you-slide" class="flexbox vleft auto-fadein"}
```
## Testing

@alequartz's unit testing would be a really useful addition.

### DIFF
--- a/R/base64.R
+++ b/R/base64.R
@@ -117,10 +117,14 @@ process_html_res <- function(html, reg, processor) {
 }
 
 process_images <- function(html, processor) {
-  process_html_res(
+  html <- process_html_res(
     html,
     "<\\s*[Ii][Mm][Gg]\\s+[Ss][Rr][Cc]\\s*=\\s*[\"']([^\"']+)[\"']",
     processor)
+  process_html_res(
+      html,
+      "<[^>]*style=\"[^\"]*url\\(([^\\)]+)\\)",
+      processor)
 }
 
 base64_encode_images <- function(html, encoder) {
--- a/inst/rmd/ioslides/ioslides_presentation.lua
+++ b/inst/rmd/ioslides/ioslides_presentation.lua
@@ -41,6 +41,39 @@ local function attributes(attr)
   return table.concat(attr_table)
 end
 
+-- Helper function to split a string on spaces
+-- returns a table
+local function split(str)
+  local words = {}
+  for word in str:gmatch("%S+") do table.insert(words, word) end
+  return words
+end
+
+-- Helper function to remove duplicates
+-- returns a table http://stackoverflow.com/a/20067270
+local function uniq(t)
+  local seen = {}
+  local res = {}
+  for _,v in ipairs(t) do
+    if (not seen[v]) then
+      res[#res+1] = v
+      seen[v] = true
+    end
+  end
+  return res
+end
+
+-- Helper function to filter a list
+-- returns a table
+local function grep(f, l)
+  local res = {}
+  for _,v in ipairs(l) do
+    if (f(v)) then
+      res[#res+1] = v
+    end
+  end
+  return res
+end
 
 -- Blocksep is used to separate block elements.
 function Blocksep()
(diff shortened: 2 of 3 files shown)


kiwiroy:
Is there anything I can do to increase the merge ability of this?

jjallaire:
The data-background stuff looks great. On the `slide-class`, I'd rather see the syntax for this be something like:

Heading {.css-tricks}

To make this work I think you'd need to let pandoc apply the class to the section div and then use jquery at startup to move classes (that aren't section, level2, etc.) from divs onto their enclosed slide element.

alequartz:
I have updated my pull request. Maybe you could append you tests to the test-ioslides.R

kiwiroy:
@jjallaire thanks for the input. The `Heading {.css-tricks}` is already applied by pandoc, but the `id` and `class` attributes are applied to the `<article>` tag, by the [lua code](https://github.com/rstudio/rmarkdown/blob/master/inst/rmd/ioslides/ioslides_presentation.lua#L264-L266). I was not sure this duality could be resolved simply, but I like the simplicity of what you have suggested. In an attempt to capture that simplicity and build on it, I've augmented the PR. The new usage is thus:

```
## Heading {.css-tricks..}

The trailing `..` attempts to suggest parent tag as in parent directory :smile:

The trailing `..` are removed so the `.css-tricks` selector must be used in a stylesheet.

`
.css-tricks > hgroup {
  background-color: rgb(100, 100, 100);
  margin-top: -40px;
  padding-top: 40px;
  margin-left: -60px;
  padding-left: 60px;
  margin-right: -60px;
  padding-right: 60px;
}
`
```

If this syntax is acceptable and likely to be merged I can prepare a pull request against the gh-pages branch.

@alequartz excellent, I've refactored slightly to allow for easier additional testing.

jjallaire:
To be honest I don't love the .. syntax (it seems obscure to me). Standard
pandoc behavior is to attach classes and data attributes to the section tag
not to the header tag so I think this is the behavior that users will
expect. If it's too tricky to do this in Lua code then you can pretty
easily do it with jquery on page load.

kiwiroy:
@jjallaire it isn't that it is difficult to achieve in Lua, rather the syntax is a challenge and why my original choice was to add an attribute using the same technique for the `{data-background=#CCC}`. Whether jQuery or Lua achieve the assignment / movement of the class, the syntax for achieving this needs to be obvious for the user and easily differentiated for the code - yes I'd like that flexibility. As an example

```
## Heading One {.flexbox}
Lua sees `attr["class"] = "flexbox"`

## Heading Two {slide-class="thank-you-slide" class="flexbox vleft auto-fadein"}
Easily separated, Lua gets `attr["slide-class"] and attr["class"]`

## Heading Three {class="thank-you-slide.. flexbox vleft auto-fadein"}
Like the previous slide, easily separated, concise, but granted, obscure
```

347e518 removes this to focus on the `data-background` code as the branch name suggests.

kiwiroy:
Just realised the code was not honouring `self_contained: true` hence 26f0e07.

R/base64.R
```diff
@@ -117,10 +117,14 @@ process_html_res <- function(html, reg, processor) {
 }
 
 process_images <- function(html, processor) {
```
yihui: `process_images()` is called by `html_document_base()`, so the change in this function has potential risk of affecting all HTML output formats. I'd rather be a little conservative, and move the new `process_html_res()` call you added to the end of `base64_encode_images()` (line `#140`), so that this only affects `ioslides_presentation()`, because only `ioslides_presentation()` calls `base64_encode_images()`.

kiwiroy: @yihui sorry I didn't see this update. No worries on the timing, I've been reporting using a branch that includes it in the intervening time. Good to see it merged. Thanks.

yihui: No worries. Thanks!

jaredlander:
Been playing with this and it is a great solution. One additional request though. this sets `background-size: 100% 100%;` inline, for the slide so that global css can't override.

Can we have an option to set other `background-size` and `background-repeat` etc?

For example, my background images look better with `background-position: center; background-repeat: no-repeat; background-size: contain;`.

yihui:
@jaredlander Okay, 70a1847879ac10509c786a8fb649e5c14b490483 was my [wedding gift](https://www.nytimes.com/2017/10/22/fashion/weddings/rebecca-martin-jared-lander.html) to you. Now `background-size` defaults to `contain`, and `background-position` defaults to `center`. If you want to change them, just use the attributes `data-background-size` and `data-background-position`.

jaredlander:
Thank you @yihui this makes both me and @beckmart very happy.

jaredlander:
Not sure if this is related or if it was brought up elsewhere, but with the latest version of `rmarkdown` (installed from GitHub to get the background slide functionality), `knitr::include_graphics` no longer gets URI encoded. I doubt this is related to background images, but it's an interesting change? Was it by design or unintended?

yihui:
@jaredlander Could you provide a minimal reproducible example as a zip archive? You probably want to use `knitr::image_uri()` instead.

yihui:
@jaredlander Could you test the latest devel version here? I think your issue might be the same as #1197, which I just fixed. Thanks!

jaredlander:
@yihui I'm working on a talk now. Will update when I get in the air and report back.

jaredlander:
Looks good now! @yihui

yihui:
@jaredlander Glad to hear that!

PR #688, opened 2016-05-02
Make sure pandoc_version() is not NULL before testing it against version numbers

yihui:
Otherwise users may get a confusing error message like

> Error in if (version >= "1.15.2") latex_template <- "default-1.15.2.tex" else if (version >=  : argument is of length zero

This was originally reported in shiny-discuss https://groups.google.com/d/msgid/shiny-discuss/63de59d4-b208-4cd3-a98e-3f48de927511%40googlegroups.com and turns out to be due to an incorrect manipulation of the RSTUDIO_PANDOC variable.

### DIFF
--- a/R/pdf_document.R
+++ b/R/pdf_document.R
@@ -109,6 +109,7 @@ pdf_document <- function(toc = FALSE,
   # template path and assets
   if (identical(template, "default")) {
 
+    if (!pandoc_available()) stop("Pandoc not found")
     # choose the right template
     version <- pandoc_version()
     if (version >= "1.17.0.2")
--- a/R/pdf_document.R
+++ b/R/pdf_document.R
@@ -109,7 +109,7 @@ pdf_document <- function(toc = FALSE,
   # template path and assets
   if (identical(template, "default")) {
 
-    if (!pandoc_available()) stop("Pandoc not found")
+    pandoc_available(error = TRUE)
     # choose the right template
     version <- pandoc_version()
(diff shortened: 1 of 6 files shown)


R/pdf_document.R
```diff
@@ -109,6 +109,7 @@ pdf_document <- function(toc = FALSE,
   # template path and assets
   if (identical(template, "default")) {
 
+    if (!pandoc_available()) stop("Pandoc not found")
```
jspiewak: is it possible to say where we looked for pandoc to make it more obvious to the user that they have something misconfigured?

yihui: I think one has to work hard enough to be able to shoot oneself in the foot, since we have set up everything correctly in the IDE, Shiny Server, and RSC... It is possible to print out the places where we searched for pandoc, but I'm not sure if it is worth it.

jspiewak: From the perspective of a customer using Shinyapps.io, they will see "Pandoc not found" and likely open a support ticket asking if it is available, which will loop in both support and a developer.

Please consider making the error message as informative as is necessary for an end user to determine the nature of their mistake and correct it, without resorting to opening a support case.

yihui: Yep I definitely understand it. As I said, it should be very rare for them to run into this problem. Anyway, I'll change the error message and make it more clear.

yihui:
This PR should be ready to go.

PR #691, opened 2016-05-05
implement html_notebook + related routines

kevinushey:
This PR implements the R Notebook render format, + accompanying code to parse the generated `.nb.html` file. Most of the code is brought over from the RStudio side as-is, with some renaming to match e.g. style and so on.

## TODO
- [x] Export + document relevant methods
- [x] Create tests for some example `.Rmd` files
- [x] Think about access to the `evaluate` hook


DESCRIPTION
```diff
@@ -61,7 +61,9 @@ Imports:
     knitr (>= 1.12),
     yaml (>= 2.1.5),
     htmltools (>= 0.2.4),
-    caTools
+    caTools,
```
jjallaire: We'll need to add these (and their dependencies if any) to the rmarkdown dependencies here: https://github.com/rstudio/rstudio/blob/master/src/gwt/src/org/rstudio/studio/client/common/dependencies/DependencyManager.java#L159-L175

Note that we've already got caTools as a downstream dependency (perhaps for knitr?). 

We depend on jsonlite for Shiny here: https://github.com/rstudio/rstudio/blob/master/src/gwt/src/org/rstudio/studio/client/common/dependencies/DependencyManager.java#L268 but we'd still need to add it for rmarkdown. If possible let's list the same version number (0.9.16) for the sake of consistency.

NAMESPACE
```diff
@@ -41,6 +42,7 @@ export(pandoc_template)
 export(pandoc_toc_args)
 export(pandoc_variable_arg)
 export(pandoc_version)
+export(parse_notebook)
```
jjallaire: maybe parse_html_notebook?

R/html_notebook.R
```diff
@@ -0,0 +1,204 @@
...
+#' Format for converting from R Markdown to an HTML notebook.
+#'
+#' @inheritParams html_document
+#' @param knitr A list of \code{knitr} hooks, used to annotate notebook output.
+#'   See \code{knitr_options} for more information.
+#' @param ... Optional arguments to be passed to \code{\link{html_document}}.
+#' @export
+html_notebook <- function(code_folding = "show",
+                          highlight = "textmate",
+                          knitr = knitr_options_notebook(),
```
jjallaire: Does these need exposure here? (there are the yaml-level options, not sure if users would ever tweak these).

R/render_html.R
```diff
@@ -0,0 +1,56 @@
...
+notebook_render_html <- function(path = NULL, bytes = NULL, format) {
+
+  if (is.null(bytes))
+    bytes <- read_file(path, binary = TRUE)
+
+  encoded <- base64_encode_object(bytes)
+  sprintf(format, encoded)
+}
+
+noteobok_render_png <- function(path = NULL, bytes = NULL) {
```
jjallaire: Spelling

R/html_notebook.R
```diff
@@ -0,0 +1,204 @@
...
+notebook_annotated_knitr_hook <- function(label, hook, meta = NULL) {
+  force(list(label, hook, meta))
+  function(x, ...) {
+    output <- hook(x, ...)
+    meta <- if (is.function(meta)) meta(x, output, ...)
+    notebook_annotated_output(output, label, meta)
+  }
+}
+
+knitr_options_notebook <- function() {
```
jjallaire: I'm thinking this should definitely not be called from the signature of html_notebook (and therefore be override-able). Better we just call this within the implementation of html_notebook if that's practical.

kevinushey: I think you're right, although there's a chance that RStudio will need to be able to override any pre-existing hooks (ie, only setting the `evaluate` hook may not be enough). I'll need to play with this a little bit.

jjallaire:
I think for the evaluate hook we should wait until knitr v1.13 hits CRAN and then add an `evaluate` argument to the `html_notebook` function. This is mostly just to document the expectation that front ends may want to provide cached output in lieu of actually evaluating the chunks (we can still use our internal hook for the time being for compatibility with more versions of knitr).

DESCRIPTION
kevinushey: FWIW we list 0.9.19 here: https://github.com/rstudio/rstudio/blob/master/src/gwt/src/org/rstudio/studio/client/common/dependencies/DependencyManager.java#L437

Should we just update the `jsonlite` version throughout?

R/html_notebook.R
jjallaire: Okay, let's discuss after you've thought it through some more. If you absolutely need to make it all pluggable I'm okay with that (but we may want to have the default be `NULL` rather than calling a function so that we don't need to export that function as well.

DESCRIPTION
jjallaire: Sounds good!

PR #693, opened 2016-05-06
implement 'output_source' evaluation hook

kevinushey:
NOTE: not quite ready for merge; but ready for 'cross-examination'.

This PR implements the `output_source` hook, which will provide a mechanism for users to override the `evaluate` hook for e.g. generation of notebooks.

The general idea is this:

1. Output formats now accept an `output_source` option, which is a function to be called within an evaluate hook as needed.

2. This function gets the executing code, the R chunk context, and any extra arguments passed to `evaluate` (through `...`)

3. Implementers of `output_source` will be able to call a function `notebook_render_output()` to generate output in such a way that RStudio will be able to parse the generated document.


kevinushey:
I'm trying to figure out what we should enforce on the `output_source` return value. Basically, it needs to be something that conforms to the output of `evaluate::evaluate`, but I wonder if we should be strict (e.g. every output needs to go through an appropriate `notebook_output()` function) or more loose (e.g. figure out how to wrap whatever output as necessary)

R/render.R
```diff
@@ -24,7 +24,11 @@ render <- function(input,
                    envir = parent.frame(),
                    run_pandoc = TRUE,
                    quiet = FALSE,
-                   encoding = getOption("encoding")) {
+                   encoding = getOption("encoding"),
+                   on_exit = NULL) {
```
jjallaire: I was thinking that all of this logic could live inside the output format (i.e. the format itself would define an on_exit function in the same way it currently defines e.g. pre_knit and post_knit).

R/render.R
```diff
@@ -333,6 +337,76 @@ render <- function(input,
     }, add = TRUE)
     env$metadata <- yaml_front_matter
 
+    # hook up output_source to evaluate if specified
```
jjallaire: Again, I was thinking that the html_notebook format would have an output_source and the concept would remain scoped there (i.e. render wouldn't be aware of it). It very well could be that this isn't possible, but I think it's cleaner if we keep this concept associated with html_notebook. If you are uncertain we can just discuss realtime in Austin on Monday.

It should be the case that anything render can do in terms of hooks a format can do. If this isn't the case perhaps we could make whatever changes to `render` required to make this so.

jjallaire:
I'd vote for requiring the explicit calls so that we can have a known scope
of input data (you can always create a catch-all HTML type for arbitrary
rich output)

R/render.R
kevinushey: I see -- the idea then would be that the format can do any of the work it needs to in the `pre_knit` hook; does that sound correct? (Ie, we can move all of the code I added to `render` into the format's `pre_knit` hook). I think that should be doable.

jjallaire: Yes, exactly. It would all go in the pre_knit hook. The output_format
object would also gain another member (an on_exit hook). The html_notebook
format would implement on_exit to reverse out it's various global state
changes.

kevinushey:
I think this should be good to go, although we might have a couple opportunities for cleanup. Here's what's going on so far:

1. Formats can now specify an `output_source` function (with appropriate signature), to handle generation of chunk outputs as necessary.

2. Outputs can essentially be any regular R output (which will be handled by e.g. `knitr::knit_print()`), or special `html_notebook_render_*` functions for outputting certain kinds of content. So far, only `html_notebook_render_png` is implemented; it takes a path and generates a (base64 encoded) ``. (We implemented an `evaluate_output` wrapper that effectively wraps any output in a form expected by the normal `evaluate` hook)

3. Right now, we just the `pre_knit` hook to register these hooks + `on_exit` to clean up, but it's likely that we could just make these part of the regular `knitr` options. If this is the case we might not even need the `on_exit` hook.

Let me know what you think @jjallaire.

R/html_notebook.R
```diff
@@ -22,12 +27,76 @@ html_notebook <- function(toc = FALSE,
...
                           ...)
 {
-  # use a pre-knit hook to capture the original document
-  encoded_document <- NULL
+  # some global state that is captured in pre_knit
+  evaluate_hook <- NULL
+  exit_actions <- list()
+  on_exit <- function() {
+    for (action in exit_actions)
+      action()
```
kevinushey: `try()` to ensure all hooks get executed?

PR #703, opened 2016-05-19
handle spaces in output path

kevinushey:
This PR fixes an issue where attempting to render a file to a path containing spaces could fail (when that output path was provided explicitly by the user, rather than generated by rmarkdown).


jjallaire:
This looks good to me. @yihui Could you also take a look?

PR #706, opened 2016-05-24
Add toc_print_hide as a sub-option under toc_float

haozhu233:
Floating TOC is convenient on html page but sometimes it looks ugly on paper if users choose to print out the page, especially when the floating TOC is very long. So I added a sub-option called `toc_print_hide` under `toc_float`. It allows people to only print out the contents of a page while the TOC is still displayed on the webpage. The default value of this option is `FALSE` so users won't feel any differences unless they have the need. 

Also, I fixed a typo.. :)

screenshot 2016-05-24 13 48 43

### DIFF
--- a/R/html_document.R
+++ b/R/html_document.R
@@ -120,11 +120,12 @@
 #'  control the behavior of the floating table of contents. Options include:
 #'
 #'  \itemize{ \item{\code{collapsed} (defaults to \code{TRUE}) controls whether
-#'  the table of contents appers with only the top-level (H2) headers. When
+#'  the table of contents appears with only the top-level (H2) headers. When
 #'  collapsed the table of contents is automatically expanded inline when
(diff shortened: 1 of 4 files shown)


haozhu233:
Close and reopen PR to trigger another travis check. It seems that it failed for no reason last time

R/html_document.R
```diff
@@ -226,7 +227,8 @@ html_document <- function(toc = FALSE,
 
     # resolve options
     toc_float_options <- list(collapsed = TRUE,
-                              smooth_scroll = TRUE)
+                              smooth_scroll = TRUE,
+                              print_hide = FALSE)
```
jjallaire: How about we just call the option `print` and default it to `TRUE`?

jjallaire:
Could you add a NEWS item for this as well?

R/html_document.R
haozhu233: It makes a lot more sense now 👍

PR #708, opened 2016-05-27
Implement proj_download option to embed zipped project directory

noamross:
This PR is an experiment in implementation of a `proj_download` option to `html_document()` to embed and allow downloading a zipped archive of the entire project folder.  Currently, one provides a folder path as `proj_download`, and the folder is zipped, base64 encoded, and embedded in the document.

I'm currently having problem getting encoding right.  I don't fully understand the `saveAs` javascript function in `FileSaver.min.js`, so I'm not sure what I need leading the base64-encoded string in the HTML file (e.g., whether to use `dataURI()` at `html_document.R#301`), and what goes in `type:` at `navigation-1.1/projembed.js#6`. As a result the downloaded file is mangled somehow. 

I may implement the following future changes if you consider merging this worthwhile.
-  An option to automatically select the project folder rather than providing a path, using `rprojroot`
-  Sensible defaults and a way to specify whether to include dotfiles, files in `.gitignore`, etc.
-  Some DRYing, as there's some overlap between code for `code_download` and `proj_download`


jjallaire:
Even though it shifts a bit more work onto the user I might prefer if an explicit list of files had to be specified (because it's so easy to end up with junk / temporary files when you take a whole directory). We could recast this as a sub-argument of `code_download`, e.g.

    code_download:
      supporting_files:
        - data.csv
        - utils.R 

You might hate this because it shifts a lot of work onto the person sharing but I think this work is *well worth it* from the standpoint of everyone downstream that has to deal with the potentially considerable size of the archive embedded in the Rmd.

jjallaire:
Actually, there might be a much nicer path forward here. We already have code (used for `runtime: shiny`, @jmcphers knows the details) that can analyze the source code of an Rmd and determine what it's dependencies are (i.e. files referenced in YAML and/or R source code). So with that we could reduce the syntax to:

    code_download:
      supporting_files: true

Then the user could optionally specify additional files not picked up the source code scanning.

R/html_document.R
```diff
@@ -287,6 +293,16 @@ html_document <- function(toc = FALSE,
         base64enc::base64encode(input),
         '</div>')
     }
+    if (!is.null(proj_download)) {
+      proj_zip_file <- file.path(tempdir(), paste0(basename(proj_download), ".zip"))
+      zip(zipfile = proj_zip_file, flags = "-r9Xq", files = list.files(proj_download, recursive = TRUE, no.. = TRUE))
+      proj_zip <<- paste0(
+        '<div id="proj-zip">',
+        base64enc::base64encode(readBin(proj_zip_file, "raw",
```
kevinushey: FWIW I think the `base64enc` function also accepts file paths directly (so you don't need to read into R, and then encode)

noamross:
Thanks, I'll keep working on this.  @jmcphers, where would I find the code to find file dependencies? I've been poking around this package and **shiny** but haven't been able to find it.

jjallaire:
It's the find_external_resources function:

https://github.com/rstudio/rmarkdown/blob/master/R/html_resources.R

This will do auto-discovery using a variety of heuristics. In addition,
users can add a top level "resource_files" to the document YAML to name
files not picked up by auto-discovery (these references can include
directories and wildcard patterns).

jjallaire:
Based on the "resources" nomenclature here we might want the syntax to be something like:

```yaml
title: "My Document"
output:
  html_document:
    code_download:
      include_resources: true
resource_files:
  non_auto_discovered.csv
```

noamross:
The current update implements the auto-detection using @jjallaire's syntax above. Some notes:

-  The files are zipped from the lowest common parent directory of all resources including the Rmd.  This means that even if there are resources above the Rmd, the document should be reproducible as long as paths are relative.
-  Also, one can simply include the whole project directory with

```
title: "My Document"
output:
  html_document:
    code_download:
      include_resources: true
resource_files:
 -  "."
```

-  Invisible files are still ignored in the above case, but I've successfully added `.git`, `.gitignore`, and `*.Rproj`, and gotten the whole working project back.
-  `source_download_link()` allows users to insert the code download link manually. As such, using `code_download` without a theme now provides a warning, rather than an error.

jjallaire:
This is a big changelist that we'd definitely like to merge (along with whatever further refinements we come up with). That said we are flat out preparing for useR so will have to defer detailed review until July.

noamross:
Great. I may make a few more refinements in the meantime, primarily so that one has the option of using `include_resources` with R notebooks without breaking notebook parsing (as is currently the case).

yihui:
I just provided a much simpler and more flexible implementation in the **xfun** package:

```r
devtools::install_github('yihui/xfun')
```

Basically you can use one of the functions `xfun::embed_file()`, `xfun::embed_dir()`, or `xfun::embed_files()` in an R code chunk to embed one or multiple files or an entire dir in the HTML output, e.g.,

````
```{r echo=FALSE}
xfun::embed_files(c('source.Rmd', 'data.csv'))
```
````

Users have the full freedom to decide which files to embed, and the list of files can be programmatically obtained (e.g., `list.files('.', '[.](Rmd|csv)$')`).

I did it in a third-party package **xfun** because it can be generally useful (not only to `rmarkdown::html_document()`), e.g., https://github.com/rstudio/blogdown/issues/106

BTW, this implementation simply takes advantage of `<a download></a>` (i.e. only works for modern web browsers), and does not require JavaScript.

PR #711, opened 2016-05-31
add readme option to html_vignette to produce package README.md

jjallaire:
@yihui and @hadley I thought it would be useful to specify that a given vignette should also be used as the package README.md. This is my first pass at an implementation. Some questions:

1. Do you agree this is a good idea?

2. Is this done in the right place? (or should it e.g. be a property of the vignette builder)

3. I'm doing this only when not on CRAN. Is that the right thing or should we not care about whether we are on CRAN? (this imposed some cost as currently the Knit button in RStudio doesn't set NOT_CRAN=true although it certainly could).

### DIFF
--- a/R/html_vignette.R
+++ b/R/html_vignette.R
@@ -20,7 +20,8 @@
 #' @inheritParams html_document
 #' @param ... Additional arguments passed to \code{\link{html_document}}
 #' @param readme Use this vignette as the package README.md file (i.e. render
-#'   it as README.md to the package root)
+#'   it as README.md to the package root). Note that if there are image files
+#'   within your vignette you should be sure to add README_files to .Rbuildignore
 #' @return R Markdown output format to pass to \code{\link{render}}
 #' @export
 html_vignette <- function(fig_width = 3,
--- a/R/html_vignette.R
+++ b/R/html_vignette.R
@@ -41,7 +41,8 @@ html_vignette <- function(fig_width = 3,
                         output_format = "github_document",
                         output_options = list(html_preview = FALSE),
                         output_file = "README.md",
-                        output_dir = dirname(dirname(input)))
+                        output_dir = dirname(dirname(input)),
+                        quiet = TRUE)
     }
   }
 
--- a/R/html_vignette.R
+++ b/R/html_vignette.R
@@ -19,12 +19,15 @@
 #'
 #' @inheritParams html_document
 #' @param ... Additional arguments passed to \code{\link{html_document}}
+#' @param readme Use this vignette as the package README.md file (i.e. render
+#'   it as README.md to the package root)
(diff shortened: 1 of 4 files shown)


hadley:
I like it!

Adding @jennybc to the thread because I know she is interested in the problem of duplicate readme material in multiple places

jjallaire:
Spoke with Hadley about this realtime. 2 follow ups:

1. We probably don't need to worry about checking NOT_CRAN (as there is no harm in this file getting rebuilt by CRAN as far as we can tell)

2. This definitely belongs in the html_vignette format rather than the vignette engine because that's only run at vignette build time (which might occur in a temp directory) and this code needs to run at development time (e.g. during "Knit" in RStudio).

R/html_vignette.R
```diff
@@ -19,25 +19,42 @@
...
-                css = css,
-                theme = NULL,
-                highlight = "pygments",
-                ...)
+  pre_knit <- function(input, ...) {
+    if (readme) {
+      rmarkdown::render(input,
+                        output_format = "github_document",
+                        output_file = "README.md",
+                        output_dir = dirname(dirname(input)))
```
kevinushey: Is the intention here that this would only get called on vignettes in e.g. the `inst/doc` folder? (Ie, this would get called after e.g. `R CMD build` copies from `vignettes/` to `inst/doc/`?)

Just want to make sure that this wouldn't inadvertently leak out of the project folder for certain cases.

jennybc:
I tried it and it's quite nifty! In addition to `README.md`, I get `README.html` in the top-level directory, which I didn't quite expect with `github_document`.

R/html_vignette.R
jjallaire: I wonder if this whole thing might be better off as a utility function
(wondering if there might be too much magic here leaving us open to as you
pointed out files leaking out of the project directory, R CMD check
failures, etc.).

jjallaire:
That README.html is probably a bug. I'll take a look.

jjallaire:
I'm not seeing the REAME.html getting produced. Is this repeatable for you?

jennybc:
Yes. It's entirely possible I'm using it incorrectly because I wasn't sure how we are meant to use it? I installed `rmarkdown` from this PR. I have this yaml in an Rmd vignette in `vignettes/` (created via `devtools::use_vignette()`:

``` yaml
---
title: "Vignette Title"
author: "Vignette Author"
date: "`r Sys.Date()`"
output:
  rmarkdown::html_vignette:
    readme: true
vignette: >
  %\VignetteIndexEntry{Vignette Title}
  %\VignetteEncoding{UTF-8}
---
```

And I execute `render("vignettes/two-for-one.Rmd")`.

jjallaire:
Okay, I repro'd it and just submitted a fix. Thanks!

Issue #715, opened 2016-06-04
render_site cannot show the Chinese characters correctly


byzheng:
I have the similar problem with #712. The same document with Chinese characters can show Chinese characters with `render_html`, but cannot with `render_site`. 

My file is saved as UTF-8 format, and already try render_site(encoding="UTF-8").

byzheng:
Now `render_site(encoding="UTF-8")` can render Chinese characters correctly. But the build site button in RStudio still cannot correctly show Chinese characters.

yihui:
Sounds like RStudio didn't pass the encoding to `render_site()`? cc @jjallaire

yihui:
Scratch that. Just saw your fix. Thanks!

byzheng:
There is one more problem need to be fixed: L312 at html_document.R, but I am not sure how to fix it.

config <- site_config(input_file, encoding = ???)

yihui:
We can pass `encoding` to the `post_knit` function: https://github.com/rstudio/rmarkdown/blob/master/R/render.R#L358-L360

jjallaire:
One note here: since there are existing post_knit handlers in the wild that
don't take ... we'd be breaking those if we just added a parameter. As a
result, I think we should check the formals of the post_knit function
before calling and only pass the encoding if it takes > 3 arguments (or
better yet if it actually has an encoding parameter).

jjallaire:
Re-opening to confirm that we've in fact fixed the problem with the latest commits to master (we already updated RStudio Preview Release to pass encoding to render_site). If everything is now working as expected please close this issue. Thanks!

PR #716, opened 2016-06-04
Update render_site.R

byzheng:
fixed bug #715


jjallaire:
Thanks for this. Traveling this week but will take a further look next week.

yihui:
I think this PR is simple enough and should be safe, so I just merged it.

Issue #722, opened 2016-06-10
Scrambled accents in downloadable Rmd source 


aoles:
Consider the following minimal example Rmd file.

```

---
title: "Testfile"
author: "Andrzej Oleś"
output:
  html_document:
    code_download: yes

---

The document author name contains an accent.
```

When I render the document and download its source through the "Code>Download Rmd" button I get the following output:

```

---
title: "Testfile"
author: "Andrzej OleÅ"
output:
  html_document:
    code_download: yes

---

The document author name contains an accent.
```

aoles:
Fixed in https://github.com/rstudio/rmarkdown/pull/723

aoles:
Hi @jjallaire, @jcheng5,

I wondering whether you had a chance to look into the solutions discussed in https://github.com/rstudio/rmarkdown/pull/723. It's almost a year now since I've reported this problem, and I still observe it with the newest version of rmarkdown.

Thanks,
Andrzej

yihui:
I have proposed a fix at #1179. You may test it with:

```r
devtools::install_github('rstudio/rmarkdown#1179')
```

Your example works okay with this fix.

aoles:
@yihui, many thanks for taking lead on this. The fix works great for me!

PR #723, opened 2016-06-10
Byte-encode Rmd source to retain utf-8 encoding

aoles:
Fixes the issue that utf-8 encoded special characters in the downloadable Rmd source file get scrambled, see https://github.com/rstudio/rmarkdown/issues/722.

### DIFF
--- a/inst/rmd/h/navigation-1.1/sourceembed.js
+++ b/inst/rmd/h/navigation-1.1/sourceembed.js
@@ -3,7 +3,13 @@
 window.initializeSourceEmbed = function(filename) {
   $("#rmd-download-source").click(function() {
     var src = window.atob($("#rmd-source-code").html());
-    var blob = new Blob([src], {type: "text/x-r-markdown"});
+
+    var uint8 = new Uint8Array(src.length);
+    for (var i = 0; i < uint8.length; i++) {
+      uint8[i] = src.charCodeAt(i);
+    }
+
+    var blob = new Blob([uint8], {type: "text/x-r-markdown"});
     saveAs(blob, filename);
   });
 };


jjallaire:
LGTM. @jmcphers and @yihui any concerns here?

yihui:
I have no idea. I'm not an expert here.

jjallaire:
Okay, @jmcphers and @jcheng5 could you take a look and let us know if you see any problem with this (I confess it also out of my scope to know whether there are subtle problems which could be introduced here).

jcheng5:
It's weird that this works. Technically `charCodeAt` can return up to 65535, while we're only assigning into an uint8 array. However `window.atob` returns not a proper UTF-8 string (where `ś` would be a single character) but instead all the bytes are treated as individual characters.

In the example you gave, `ś` is represented like this in the `src` variable:

```
String.fromCharCode(0xC5) + String.fromCharCode(0x9B)
```

Instead the "correct" representation would be:

```
String.fromCharCode(347)
```

In other words, this code works but only because two "wrongs" cancel each other out. If we test this with some CJK characters and it works, I'm fine with shipping it, but let's keep on the lookout for further strangeness. We can use something like [this](https://github.com/beatgammit/base64-js/blob/master/lib/b64.js) to avoid the string at all, and just go straight from base64 to TypedArray, if necessary.

jcheng5:
Another way to go would be to forego JavaScript entirely and do something like:

```html
<a download="whatever.Rmd" href="data:text/x-r-markdown;charset=utf-8;base64,LS0tCnRpdGxlOiAiVGVzdGZpbGUiCmF1dGhvcjogIkFuZHJ6ZWogT2xlxZsiCm91dHB1dDoKICBodG1sX2RvY3VtZW50OgogICAgY29kZV9kb3dubG9hZDogeWVzCiAgICBzZWxmX2NvbnRhaW5lZDogbm8KLS0tCiAgCiAgVGhlIGRvY3VtZW50IGF1dGhvciBuYW1lIGNvbnRhaW5zIGFuIGFjY2VudC4K">...</a>
```

The download attribute is [supported in Chrome, Firefox, and Edge](http://caniuse.com/#feat=download). It isn't supported in Safari, but the results are no different than FileSaver.js (and note that I added the charset to the data URI, which improves matters for Safari).

jjallaire:
Thanks @jcheng5, I'll look at those alternatives.

@aoles <https://github.com/aoles> note that it may be a bit before I look
at this further (too many things to do before useR!)

aoles:
Awesome, thank you @jjallaire and @jcheng5 for looking into this!

aoles:
Hi, I just wanted to quickly check with you guys the status of this issue. It has already been a while since I filed https://github.com/rstudio/rmarkdown/issues/722, but it doesn't seem to be resolved yet.

Cheers,
Andrzej

PR #737, opened 2016-07-07
puts extra_dependencies at the end of format_deps

ijlyttle:
This way, custom formats can work "on top of" bootstrap, etc.

Please see https://github.com/ijlyttle/user2016docdemo/issues/1 for example and further rationale.

### DIFF
--- a/R/html_document_base.R
+++ b/R/html_document_base.R
@@ -83,7 +83,6 @@ html_document_base <- function(smart = TRUE,
     # resolve and inject extras, including dependencies specified by the format
     # and dependencies specified by the user (via extra_dependencies)
     format_deps <- list()
-    format_deps <- append(format_deps, extra_dependencies)
     if (!is.null(theme)) {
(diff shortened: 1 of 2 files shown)


jjallaire:
@yihui Could you confirm that this works with your package(s) that use extra_dependencies now (just want to make sure we don't have a lurking order dependency here).

@ijlyttle Could you add an item to NEWS as part of this PR.

yihui:
I'm using `theme = NULL` in tufte and bookdown, so this PR should not affect them. I guess most people are not even aware of the `extra_dependencies` argument (I just searched on Github and there are no results matching `extra_dependencies`), so it should be a reasonably safe change.

ijlyttle:
Gentle ping on this PR - if/when accepted and next CRAN release made, I can offer some tutorial materials on how to customize `html_document()` using packages and html dependencies, via https://github.com/ijlyttle/user2016docdemo

PR #745, opened 2016-07-19
Do not ignore the knit_meta passed to render()

yihui:
Currently knit_meta is always overwritten, and this causes a problem in bookdown: when using the Knit-and-Merge approach, knit meta data is collected from each chapter, merged, and finally passed to render(). If it is always overwritten, all meta data will be lost. `knit_meta` is only relevant to documents that need to be knitted, so move these lines back into the if statement.

The breaking change came from ff6019a2048d135dd27427dabe0acaf61aa0c07f

Original issue reported by @hadley.

### DIFF
--- a/R/render.R
+++ b/R/render.R
@@ -356,6 +356,16 @@ render <- function(input,
                          encoding = encoding)
 
     perf_timer_stop("knitr")
+
+    # pull any R Markdown warnings from knit_meta and emit
+    rmd_warnings <- knit_meta_reset(class = "rmd_warning")
+    for (rmd_warning in rmd_warnings) {
+      message("Warning: ", rmd_warning)
+    }
+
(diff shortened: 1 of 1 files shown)


jjallaire:
I think the reason the knit_meta processing got moved was to enable support for plain .md files within R Markdown sites:

https://github.com/rstudio/rmarkdown/commit/ff6019a2048d135dd27427dabe0acaf61aa0c07f

I confess though to not recalling exactly why this was key to supporting .md. I'll revert that commit locally and play with rendering sites that include .md to see what I find. Let me know if you have additional insight.

jjallaire:
The change looks fine to me visa vi .md files in sites, our site test (which includes .md) passes, and I rebuilt the rmarkdown website (which includes many .md files) with the change so I think we are in the clear!

Issue #744, opened 2016-07-19
default spacing between paragraphs in Beamer output has changed


kevinushey:
It appears that newer versions of `pandoc` do not render separate paragraphs in a Beamer PDF output with extra whitespace between paragraphs. For a markdown file with contents:

```
# Title

A paragraph.

Another paragraph.
```

We get the following renders:

#### Older pandoc (1.12.4.2):

![screen shot 2016-07-19 at 2 30 03 pm](https://cloud.githubusercontent.com/assets/1976582/16967346/554cce88-4dbd-11e6-98ac-425de901d0ff.png)

#### Newer pandoc (1.17.2):

![screen shot 2016-07-19 at 2 30 15 pm](https://cloud.githubusercontent.com/assets/1976582/16967354/5a194e00-4dbd-11e6-9a40-b258d4d84dd2.png)

I'm not yet sure if this should be considered a regression in pandoc, or if it's just something we could handle on the `rmarkdown` side with a default beamer template (or something like that).

Is this something we might want to handle on the `rmarkdown` side?

jjallaire:
We've tended to leave the pandoc beamer template alone and let these issues get addressed on the pandoc side. I'd say we could report it there and if it's not considered a bug we'd find out why.

kevinushey:
Sounds good. Filed an issue here: https://github.com/jgm/pandoc-templates/issues/207

kevinushey:
Sounds like this was indeed a minor regression in the default template; presumedly we'll get the fix with the next release of `pandoc`.

jjallaire:
Okay, it might be a while before we update pandoc though. We do have a
function already that patches the default pandoc beamer template here:

https://github.com/rstudio/rmarkdown/blob/master/R/beamer_presentation.R#L141

We'd just have to scope our patching to just the "regressed" versions of
the beamer template.

kevinushey:
Sounds good -- I'll submit a PR.

jjallaire:
Awesome, thanks!

J

PR #747, opened 2016-07-20
patch beamer template: restore paragraph spacing

kevinushey:
This resolves https://github.com/rstudio/rmarkdown/issues/744.

### DIFF
--- a/R/beamer_presentation.R
+++ b/R/beamer_presentation.R
@@ -138,15 +138,10 @@ beamer_presentation <- function(toc = FALSE,
   )
 }
 
-patch_beamer_template <- function() {
-  pandoc_available(error = TRUE)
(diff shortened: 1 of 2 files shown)


jjallaire:
@yihui could you take a quick look and merge if it looks okay to you? (Yihui created the original version of the patch code). See https://github.com/jgm/pandoc-templates/issues/207 for details on why the additional patch is necessary.

@kevinushey Could you add a NEWS item for this?

yihui:
Looks good. The current pandoc version in RStudio is 1.17.0.2, and I'm not sure if you really mean 1.17.2 or 1.17.0.2.

kevinushey:
I'm using the latest version of pandoc installed with homebrew, which is currently 1.17.2, so that check is intentional.

yihui:
Oh I just saw that. Yep, this makes perfect sense, then.

PR #751, opened 2016-07-26
Add paged tables to render data

javierluraschi:
Almost there, need to address the SQL comment from JJ...

screen shot 2016-07-26 at 2 42 00 pm

screen shot 2016-07-26 at 3 42 44 pm


javierluraschi:
@jjallaire:

1. Removed `jquery` dependencies. Seems some of the functionality was used to generate TOC, but I rendered one doc with it and is looking good. Let me know if there is something else I should check.

2. Ran notebook with SQL chunk. It's currently rendering as text, is this what you would expect from the comment you mention about the table coming from `kable`? Would be nice to render those as pageable tables as well, I can take a look if we are interested in this. The SQL chunks looks as follows (unchanged to my understanding):

Rendering SQL chunk from UI:
screen shot 2016-07-26 at 4 12 30 pm

Rendering SQL chunk with `rmarkdown::render("main.Rmd", "html_notebook")`:
screen shot 2016-07-26 at 4 12 58 pm

javierluraschi:
@jmcphers planning to check-in the pageable-table here and then integrate into the IDE the `pageable.js` and `pageable.css` files. However, might be easier to iterate here if you have design feedback.

Here is a live demo BTW, functionality is pretty basic at the moment: https://s3-us-west-2.amazonaws.com/jluraschi-temp/rmarkdown/main.nb.html - I could consider scrolling animations if that sounds interesting.

jjallaire:
Some UX comments:

1. In `kable` character columns are left-aligned by default. I'd repeat that behavior here.

2. I think we need to display the total number of records somewhere (and which "window" we are currently in). I also think that we should have the 1 2 3 >> type UI treatment for paging. e.g. something like we see here in DataTables:

screen shot 2016-07-27 at 8 22 31 am

DESCRIPTION
```diff
@@ -13,12 +13,6 @@ Authors@R: c(
   person("Aron", "Atkins", role = "aut", email = "aron@rstudio.com"),
   person("Rob", "Hyndman", role = "aut", email = "robjhyndman@gmail.com"),
   person(family = "RStudio, Inc.", role = "cph"),
-  person(family = "jQuery Foundation", role = "cph",
```
jjallaire: We are still using jquery just don't need the explicit paring with your JS file (i.e. it's already included elsewhere automatically). The reason we had it there for stickytableheaders was to ensure it was included in the correct order (i.e. before stickytableheaders).

R/html_document.R
```diff
@@ -241,9 +241,7 @@ html_document <- function(toc = FALSE,
 
     # dependencies
     extra_dependencies <- append(extra_dependencies,
```
jjallaire: These all need to be restored.

R/html_document_base.R
```diff
@@ -85,8 +85,7 @@ html_document_base <- function(smart = TRUE,
     format_deps <- list()
     format_deps <- append(format_deps, extra_dependencies)
     if (!is.null(theme)) {
-      format_deps <- append(format_deps, list(html_dependency_jquery(),
```
jjallaire: This needs to be restored.

R/knit_print.R
```diff
@@ -1,11 +1,19 @@
...
+  )
+}
 
 knit_print.data.frame = function(x, ...) {
   context <- render_context()
   if (!is.null(context$df_print)) {
     if (identical(context$df_print, knitr::kable)) {
-      res = paste(c('<div class="kable-table">', '', knitr::kable(x), '', '</div>'),
-                  collapse = "\n")
+      res = paged_table_html(x)
```
jjallaire: We can't do this here because this code runs for all rmarkdown formats (i.e. this would result in your paged_table_html running for PDFs!!!). Rather, for the `html_notebook` format we need to pass our df printing function for the `df_print` argument to `output_format` (i.e. where we currently pass the string "kable" in `html_notebook` we pass a function).

inst/rmd/h/jquery-AUTHORS.txt
```diff
@@ -1,217 +0,0 @@
-Authors ordered by first contribution.
```
jjallaire: Restore jquery as mentioned before.

inst/rmd/h/jqueryui-1.11.4/index.html
```diff
@@ -1,505 +0,0 @@
-<!doctype html>
```
jjallaire: Restore jqueryui as mentioned before.

jjallaire:
Note that when you do integrate into the IDE you can just read the .js and .css files via `system.file` (since we install an embedded version of rmarkdown we are guaranteed to always have the latest version).

R/html_document.R
javierluraschi: Reverted the jquery removal commit.

javierluraschi:
@jjallaire Thanks, following changes performed:

1. Reverted removal of jquery change, scoped removal to `html_notebook.R`
2. Scoped use of `paged_table_html` to only the `html_notebook`
3. Left-aligned `character` and `factor` columns
4. Added navigation bar with page indexes and row-count info box

screen shot 2016-07-27 at 1 28 29 pm

jjallaire:
Maybe capitalize "previous" and "next" ?

javierluraschi:
Sure!

screen shot 2016-07-27 at 2 08 42 pm

jjallaire:
We could implement the knit sql compatibility by allowing rmarkdown to set a knit sql render function (as a knit option, sort of like we do right here: https://github.com/rstudio/rmarkdown/blob/master/R/html_notebook.R#L128-L137). This option could be sql.print, e.g.

```r
knitr::opts_knit$set(sql.print = paged_table_html)
```

Then knitr would check for it here: https://github.com/yihui/knitr/blob/master/R/engine.R#L471

Alternatively, you could just look for `<div class="knitsql-table">` and automatically transform it to your structured/implementation. 

I don't have a strong opinion about which way is best.

@yihui

yihui:
I don't have a strong opinion, either.

jjallaire:
Providing a hook within knitr would allow for a more compact (JSON) representation of the data frame within the published document. @javierluraschi Let's put a PR into knitr to make the sql.print method pluggable and then use that to redirect to our data frame renderer.

javierluraschi:
@jjallaire @yihui I went for the `knitr` option following the option pattern from `sql.max.print`.

Here is the `knitr` PR: https://github.com/yihui/knitr/pull/1257

Sample output:

screen shot 2016-07-28 at 12 18 54 am

Sample notebook: https://beta.rstudioconnect.com/content/1369/

jjallaire:
A couple more suggested refinements:

1. Do you think we need a line that runs above the footer area? (hard to know until you actually see it I guess)

2. When there are less than 10 records I think we should: (a) Just display e.g. "3 records" (rather than "1-3 of 3"); (b) Get rid of the Previous/Next treatment entirely since it's a no-op.

javierluraschi:
I think it looks better, updated:

screen shot 2016-07-28 at 8 57 06 am

screen shot 2016-07-28 at 8 57 11 am

jjallaire:
Agree it looks better. A couple other small things I noticed:

- Column headers should reflect the alignment of data

- Scratch the page number as well when Next and Previous are omitted

javierluraschi:
Updated, fixed also 0-records case:

screen shot 2016-07-28 at 9 27 02 am

screen shot 2016-07-28 at 9 27 16 am

screen shot 2016-07-28 at 9 27 25 am

jjallaire:
One thing we did in SQL printing within knitr is detect when the first column is an ID column and in that case left align it (as right aligning an ID column looks odd, the latest screenshot is a case in point). Here's the code we used for this:

https://github.com/yihui/knitr/blob/master/R/engine.R#L454-L457

I'm going to go ahead and merge this and we can make further refinements on master. This feature looks great!

hadley:
If this is going to replace the default tibble display, I'd really like to surface the variable types somehow (maybe in small grey text beneath the names)

jjallaire:
Yes, definitely! @javierluraschi <https://github.com/javierluraschi> let's
do this

javierluraschi:
Noted! I'll finish the ide integration and then perform another pass at the comments in this pr.

javierluraschi:
Following up with https://github.com/rstudio/rmarkdown/compare/feature/pagedtable-types?expand=1 for the couple remaining feedback points.

PR #766, opened 2016-08-07
fix #760: the parameter can be an expression

### LINKED ISSUE #760: 'expr' option being added to parameters that use !r syntax
I noticed that when I create a parameter using the !r syntax, an extra option is automatically added to my list of choices when I go to knit the output with parameters. Here is my param:

```yaml
notes:                
      value: "yes"
      choices: !r c("yes", "no")
      input: select
      label: Notes
      multiple: yes
```

When I Knit with Parameters from within RStudio, I see the following options under Notes:

yes, no, expr

Where is the 'expr' coming from? If I create a param that does not include the !r syntax, there is no issue. Example:

```yaml
other: 
      value: first
      choices: [first, second, third]
      input: select
      label: Other
      multiple: yes
```

This displays the following options, as expected:

first, second, third

On a side note, I played around with the 'expr' option and noticed that if you select this option, what comes back for that parameter is all of the available choices you have specified (minus 'expr' itself). Is this a precursor to 'Select All' functionality (which would be a very useful feature)?

yihui:
Use its value instead of the parameter itself, which is a list of value/expression

### DIFF
--- a/R/params.R
+++ b/R/params.R
@@ -280,7 +280,9 @@ knit_params_ask <- function(file = NULL,
           }
         } else {
           ## Not a special field. Blindly promote to the input control.
-          arguments[[name]] <<- param[[name]]
+          arguments[[name]] <<- if (inherits(param[[name]], 'knit_param_expr')) {
+            param[[name]][['value']]
+          } else param[[name]]
         }
       })
 


aronatkins:
Parameter values have the knit expressions flatted within `resolve_params`: https://github.com/yihui/knitr/blob/6b01d276d3bb8b6aaf62de5897114214e99ca476/R/params.R#L238-L258

Should we have resolve the other properties within `resolve_params` as well?

I suppose there's not a functional difference, since only `knit_params_ask` uses those fields. Just trying to understand the responsibilities of the two packages.

PR #770, opened 2016-08-09
make use of tibble to format pagedtables and minor improvements

javierluraschi:
This change makes use of `tibble` to format paged tables as follows. Please notice that this introduces `tibble` as an import instead of a suggest.

screen shot 2016-08-09 at 3 55 49 pm
screen shot 2016-08-09 at 3 55 54 pm


jjallaire:
@javierluraschi One more change we need here is to enhance the html_document format's df_print argument to accept "paged" as a possible value alongside "default", "tibble", and "kable". We can also try to do the same with ioslides_presentation, etc. but we just need to make sure that the default 10 rows w/ header and footer fits cleanly onto a single slide. Let's discuss realtime before you work on this.

jjallaire:
I think the footer should include the text "rows" and "cols" for clarity sake, e.g.

1-10 of 1000 rows | 1-10 of 19 cols

PR #795, opened 2016-09-05
Make the "show code" buttons more CSS-friendly

steveharoz:
For code folding, the show/hide code buttons have some hard-coded styles, which limits the ability to use custom CSS. This PR moves those styles to CSS.


jjallaire:
pull-right has other implications besides just float: right (it combos with several other classes). How about just removing the bottom margin bit and keeping pull-right (you can always override float: right explicitly)

steveharoz:
pull-right frustratingly uses !important, but yeah, it's overridable. I've added it back.

PR #820, opened 2016-09-30
Clean temp files on exit

### LINKED ISSUE #819: Temp files are left on disk
Hi, I have not done a thorough review of the code, but we are using rmarkdown as part of a rendering engine for a live web application and are finding that the disk on the servers fill up with rmarkdown files. I have stepped through the code enough just to know that there are instances where tempfile (EG: https://github.com/rstudio/rmarkdown/blob/master/R/html_resources.R#L186) is called, but have not gone further to see if rmarkdown already attempts to clean them up. 

If rmarkdown does not clean up, I would be happy if the temp files were completely hidden from the user and cleaned up in the background, but would also be OK with the option of passing in the directory to use for all tempfile calls.

yihui:
This PR fixes #819: a few temp files are never cleaned up after `render()` is done. I went through the code that called `tempfile()`, and added `on.exit(unlink())` to clean up those temp files. The `as_tmpfile()` ones are a little tricky, since they cannot be immediately removed and we need to wait until `pandoc_convert()` is done, so I added `clean_tmpfiles()` in `render()`, hoping that the pattern `rmarkdown-str` is unique and it does not hurt to aggressively clean up files in `tempdir()`  that happen to have the same pattern but not created by `as_tmpfile()`.

### DIFF
--- a/R/beamer_presentation.R
+++ b/R/beamer_presentation.R
@@ -78,7 +78,12 @@ beamer_presentation <- function(toc = FALSE,
 
   # template path and assets
   if (!is.null(template)) {
-    if (identical(template, "default")) template <- patch_beamer_template()
+    if (identical(template, "default")) {
+      template <- patch_beamer_template()
+      # the default template could be a tempfile, which is a patched version of
+      # Pandoc's default template
+      if (!is.null(template)) on.exit(unlink(template), add = TRUE)
+    }
     if (!is.null(template))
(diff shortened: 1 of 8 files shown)


jjallaire:
We are getting ready to submit rmarkdown to CRAN so I'm going to hold off on merging this PR until then.

yihui:
Sure. No problem. I don't think this bug will affect many users.

PR #827, opened 2016-10-05
be more permissive of 'params' objects discovered

kevinushey:
This is part one of a fix for parameterized R Notebooks.

As part of the Notebook workflow, when parameters are discovered in the YAML header, we inject an object `params` into the global environment to ensure that it's available for code execution. However, this breaks subsequent renders, as `rmarkdown::render()` then detects a pre-existing `params` variable and bails.

The fix here is to permit `params` objects with class `knit_param` and `knit_param_list`, and ensure that RStudio injects the `params` object with that class. The `rmarkdown` package will then overwrite that variable as it is known to be 'safe' for overwriting during render.

Let me know if this sounds reasonable or if you think there's an alternate solution we should consider here.

(associated change on RStudio side, to be merged: https://github.com/rstudio/rstudio/commit/c8b0c6d786acd105e7858ee49d9a332e1738ca20)

cc: @jjallaire, @jmcphers

### DIFF
--- a/R/render.R
+++ b/R/render.R
@@ -368,8 +368,12 @@ render <- function(input,
       params <- knit_params_get(input_lines, params)
 
       # bail if an object called 'params' exists in this environment,
-      # and it seems to be an unrelated user-created object
-      if (exists("params", envir = envir, inherits = FALSE)) {
+      # and it seems to be an unrelated user-created object. store
+      # references so we can restore them post-render
+      hasParams <- exists("params", envir = envir, inherits = FALSE)
+      envirParams <- NULL
+
+      if (hasParams) {
         envirParams <- get("params", envir = envir, inherits = FALSE)
         isKnownParamsObject <-
           inherits(envirParams, "knit_param_list") ||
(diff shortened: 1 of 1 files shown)


jjallaire:
Let's hold this until after we release to CRAN (next week).

jjallaire:
Kevin just pointed out that we probably do need to take this to fix a pretty major bug with saving docs w/ params. This change looks good to me. @jmcphers  ?

kevinushey:
One more change to make -- the pre-existing `params` object is not restored post-render, so we should see if that's easy to fix.

kevinushey:
Alright, I think we should be good to go now -- confirmed that save works fine w/Notebook containing params with this PR alongside the

R/render.R
```diff
@@ -367,18 +367,34 @@ render <- function(input,
...
+      # and it seems to be an unrelated user-created object. store
+      # references so we can restore them post-render
+      hasParams <- exists("params", envir = envir, inherits = FALSE)
+      envirParams <- NULL
+
+      if (hasParams) {
+        envirParams <- get("params", envir = envir, inherits = FALSE)
+        isKnownParamsObject <-
+          inherits(envirParams, "knit_param_list") ||
+          inherits(envirParams, "knit_param")
```
yihui: FYI the second argument of `inherits()` can be a character vector, so you can get rid of `||` if you want.

PR #828, opened 2016-10-06
Place folding button inside code box

jeroenooms:
Places the folding button inside the code box (the same place the controls are placed in the Rmd), removing that ugly white space:

------

screen shot 2016-10-06 at 12 00 28 pm


jjallaire:
Fantastic, thank you!

Issue #829, opened 2016-10-08
data.table := causes printing on dev version of rmarkdown


shrektan:
### Issue Description

On the dev version of `rmarkdown`, the `:=` of `data.table` will print automatically. This phenomenon occurs only when dev version of  `rmarkdown` being loaded (the CRAN version is ok ).

Also, when using `knitr` to render the `rnw` files, if `library(rmarkdown)` (the dev version) was placed in the head of the document, the `:=` will start printing. However, if there's no `library(rmarkdown)` in the head, `:=` behaves normally.
### Reproducible Code

``` r
rmd <- tempfile(fileext = ".rmd")
download.file("https://raw.githubusercontent.com/shrektan/temporary/master/datatableprinting.rmd",
              rmd)
html <- tempfile(fileext =  ".html")
rmarkdown::render(rmd, output_format = rmarkdown::html_document(), output_file = html)
shell(html)
```
### My session info

```
sessionInfo()
## R version 3.3.1 (2016-06-21)
## Platform: x86_64-w64-mingw32/x64 (64-bit)
## Running under: Windows 7 x64 (build 7601) Service Pack 1
## 
## locale:
## [1] LC_COLLATE=Chinese (Simplified)_People's Republic of China.936 
## [2] LC_CTYPE=Chinese (Simplified)_People's Republic of China.936   
## [3] LC_MONETARY=Chinese (Simplified)_People's Republic of China.936
## [4] LC_NUMERIC=C                                                   
## [5] LC_TIME=Chinese (Simplified)_People's Republic of China.936    
## 
## attached base packages:
## [1] stats     graphics  grDevices utils     datasets  methods   base     
## 
## other attached packages:
## [1] rmarkdown_1.0.9016 data.table_1.9.6  
## 
## loaded via a namespace (and not attached):
##  [1] Rcpp_0.12.7     withr_1.0.2     digest_0.6.10   assertthat_0.1 
##  [5] chron_2.3-47    formatR_1.4     magrittr_1.5    evaluate_0.9   
##  [9] stringi_1.1.2   rstudioapi_0.6  devtools_1.12.0 tools_3.3.1    
## [13] stringr_1.1.0   yaml_2.1.13     memoise_1.0.0   htmltools_0.3.5
## [17] knitr_1.14      tibble_1.2
devtools::session_info()
## Session info --------------------------------------------------------------
##  setting  value                                              
##  version  R version 3.3.1 (2016-06-21)                       
[...]

PR #830, opened 2016-10-09
suppress printing of certain 'data.table' generating expressions (#829)

kevinushey:
This is a potential fix for #829.

### DIFF
--- a/R/knit_print.R
+++ b/R/knit_print.R
@@ -1,5 +1,22 @@
 knit_print.data.frame = function(x, ...) {
   context <- render_context()
+
+  # printing of certain expressions producing
+  # 'data.table' objects should be suppressed
+  printable <- TRUE
+  if ("data.table" %in% loadedNamespaces() &&
+      exists("shouldPrint", envir = asNamespace("data.table")))
+  {
+    shouldPrint <- get("shouldPrint", envir = asNamespace("data.table"))
+    printable <- tryCatch(
+      shouldPrint(x) || !inherits(x, "data.table"),
+      error = function(e) TRUE
+    )
+  }
+
+  if (!printable)
+    return()
+
   if (!is.null(context$df_print)) {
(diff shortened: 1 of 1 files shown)


R/knit_print.R
```diff
@@ -1,5 +1,21 @@
...
   context <- render_context()
+
+  # printing of certain expressions producing
+  # 'data.table' objects should be suppressed
+  printable <- TRUE
+  if ("data.table" %in% loadedNamespaces() &&
+      exists("shouldPrint", envir = asNamespace("data.table")))
+  {
+    printable <- tryCatch(
+      data.table:::shouldPrint(x) && inherits(x, "data.table"),
```
kevinushey: FWIW I think the ordering of checks here might be important, since it might matter whether the promise `x` has been forced by the time it enters the `shouldPrint()` function.

kevinushey:
Note that we might need a similar change in the IDE, to ensure that printing of such `data.table` generating expressions aren't printed.

jjallaire:
Okay, this definitely looks like the right direction. I do however think we need to hold it for the patch release given that it's a change to fairly low-level routine. Let's create an IDE bug targeted for v1.0-patch and link it to this bug.

kevinushey:
Sounds good. I think I'm also going to try writing a unit test here.

javierluraschi:
LGTM.

We should share the code with the IDE to avoid behavior differences between versions of `data.table`. The fix in the IDE is compatible with older versions of `data.table` but it had to use a non-exported function, this fix won't support older versions of `data.table` but the fix is cleaner.  Either way, I'm good, so updated the IDE fix with this fix to share this: https://github.com/rstudio/rstudio/pull/810

kevinushey:
Shall we merge this one now? (Looks like it was left behind when the IDE-related work went in)

jjallaire:
Yes, let's merge this one (could you add a NEWS item as well?)

PR #841, opened 2016-10-19
mark_utf8 should set encoding to the name of lists

shrektan:
or when read the yaml files, the Chinese var names display messes still.

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -87,6 +87,7 @@ mark_utf8 <- function(x) {
   attrs <- attributes(x)
   res <- lapply(x, mark_utf8)
   attributes(res) <- attrs
+  names(res) <- mark_utf8(names(res))
   res
 }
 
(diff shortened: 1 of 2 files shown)


jjallaire:
@yihui I'll let you review and merge this one since you are the foremost export on UTF-8!

@shrektan Could you add an item to inst/NEWS for this fix?

shrektan:
@jjallaire added.

@yihui Actually, I find this issue when I use your code in `bookdown` for reading utf-8 yaml files...

yihui:
@shrektan Great. Thanks a lot! Actually are you familiar with C? I really hope someone can fix the root problem in the yaml package instead of patching it over and over again.

shrektan:
@yihui unfortunately I'm not~

yihui:
@shrektan That is fine. I have found a C expert and will try his solution.

PR #853, opened 2016-11-02
remove dependency to tibble

javierluraschi:
Consider removing `tibble` dependency. Two PR are out for supporting `R 3.0.0` in `tibble` and `lazyeval`:

https://github.com/hadley/lazyeval/pull/87
https://github.com/tidyverse/tibble/pull/187

However, we could take the code out from `tibble` if needed as well.


R/render.R
```diff
@@ -802,8 +802,12 @@ resolve_df_print <- function(df_print) {
   if (!is.function(df_print)) {
     if (df_print == "kable")
       df_print <- knitr::kable
-    else if (df_print == "tibble")
+    else if (df_print == "tibble") {
+      if (!requireNamespace("tibble", quietly = TRUE))
+        stop("Printing 'tibble' without 'tibble' package available")
```
kevinushey: I wonder if instead of making this a hard error we could just use the default print method (and maybe warn once per session)?

kevinushey: A slightly cleaner implementation would be with a factory function e.g.

```
make_warning <- function(message) {
    message <- message
    warned <- FALSE
    function() {
        if (warned) return()
        warning(message)
        warned <<- TRUE
    }
}

tibble_warning <- make_warning("...")
tibble_warning()
```

R/html_paged.R
```diff
@@ -1,4 +1,83 @@
...
+
+  is_atomic <- function(x) {
+    is.atomic(x) && !is.null(x)
+  }
+
+  is_vector <- function(x) {
+    is_atomic(x) || is.list(x)
+  }
+
+  is_vector_s3 <- function(x) UseMethod("is_vector_s3")
```
jjallaire: @kevinushey and @hadley Are we okay to have a package private S3 method here or could that interfere somehow with the tibble exported method of the same name? We could just rename it to is_s3_vector and/or just implement it with a switch/case.

javierluraschi: @jjallaire just in case, renamed.

R/render.R
javierluraschi: kept as is.

jjallaire:
Yes, I'd say let's not use S3 here.

javierluraschi:
K, S3 removed.

krlmlr:
FWIW, I'm planning to support R 3.0.0 in tibble 1.3.0 which will be out soon.

PR #855, opened 2016-11-06
support knitr chunk and global option for pagedtables

javierluraschi:
Currently paged table options depend on `getOption` which does not respect setup or chunk options, use `knitr::opts_current$get` instead.

### DIFF
--- a/R/html_paged.R
+++ b/R/html_paged.R
@@ -85,11 +85,18 @@ paged_table_obj_sum <- function(x) {
   )
 }
 
+paged_table_option <- function(option, default = NULL) {
+  if (is.null(knitr::opts_current$get(option)))
+    getOption(option, default)
+  else
+    knitr::opts_current$get(option)
+}
+
 #' @import methods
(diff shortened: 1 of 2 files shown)


jjallaire:
This looks good. I'm realizing that for other table printing methods we have functions that can explicitly force them for a given print operation (`knitr::kabble`, `tibble::as_tibble`, etc.). I think we should also add a publicly exported`paged_table` function to rmarkdown that forces the paged table treatment. I'll merge this then we can handle `paged_table` as a separate PR.

Could you also send a PR to update the documentation here regarding the new options you've added: https://github.com/rstudio/rmarkdown/blob/gh-pages/html_document_format.Rmd#data-frame-printing

javierluraschi:
@jjallaire Sent these out to add `paged_table()` support:

https://github.com/rstudio/rstudio/pull/876
https://github.com/rstudio/rmarkdown/pull/857
https://github.com/rstudio/rmarkdown/pull/858

PR #857, opened 2016-11-07
add support for paged_table function

javierluraschi:
Adds support for:

```{r}
paged_table(iris)
```

to print data frames as paged tables.

The `pagedtable.js` script is added now by default which adds 45KB of uncompressed content and 13KB if compressed by the server, this seems reasonable compared to additional complexity we would have to introduce to detect the use of `paged_table()` while knitting and dynamically add the `pagedtable.js` script. The `pagedtable.css` style sheet is only 2KB uncompressed.


jjallaire:
You don't need to always add the extra dependencies. You can actually pass the dependency as an additional argument to knitr::asis_output and it will get automatically added.

javierluraschi:
That's much better! Updated. See also: https://github.com/rstudio/rmarkdown/pull/858

PR #859, opened 2016-11-08
Add support for paged tables for dplyr data frames

javierluraschi:
Currently, `grouped_df`. `rowwise_df` and `tbl_sql` do not print as paged tables when `df_print: paged`.

A while ago it was explored to, rather than override, provide an option in `dplyr` to print those through an override since those 3 methods print a bit of text and then a data frame. See https://github.com/hadley/dplyr/pull/2060.

However, not that the ide supports a gallery, this new option in dplyr does not help that much since the output would appear in a separate console gallery and a data gallery entry making a simple `collect()` more complex that what we would want.

Instead, the fix is to override these 3 directly when `df_print: paged`


jjallaire:
Could we add a NEWS item for this as well?

javierluraschi:
Added `NEWS` entry: https://github.com/rstudio/rmarkdown/commit/0c2f9199bf3a5eeb08734eb3bf98b4bfa290a50d

PR #866, opened 2016-11-11
Inject CSS styling into RMD parametrization app

frankandrobot:
Modified `knit_params_ask` so that it is possible to pass custom css/scripts, both as strings and as links to URIs. 

Connected to #3945

### DIFF
--- a/R/params.R
+++ b/R/params.R
@@ -186,6 +186,56 @@ params_namedList <- function() {
   empty
 }
 
+
+setup_html_head <- function(html_head_style = c(),
+                            html_head_script = c(),
+                            html_head_style_link = c(),
+                            html_head_script_link = c()) {
+
+  default_style <- shiny::tags$style(
+    # Our controls are wiiiiide.
+    ".container-fluid .shiny-input-container { width: auto; }",
+    # Prevent the save/cancel buttons from squashing together.
+    ".navbar button { margin-left: 10px; }",
+    # Style for the navbar footer.
+    # http://getbootstrap.com/components/#navbar-fixed-bottom
+    "body { padding-bottom: 70px; }"
+  )
+  ## Escape is "cancel" and Enter is "save".
+  default_script <- shiny::tags$script(shiny::HTML("$(document).keyup(function(e) {\n",
+                                           "if (e.which == 13) { $('#save').click(); } // enter\n",
+                                           "if (e.which == 27) { $('#cancel').click(); } // esc\n",
+                                           "});"
+  ))
+
+  html_head_style <- as.vector(html_head_style)
+  custom_styles <- lapply(html_head_style, shiny::tags$style)
+
+  html_head_script <- as.vector(html_head_script)
+  custom_scripts <- lapply(html_head_script, function(x) shiny::tags$script(shiny::HTML(x)))
+
+  html_head_style_link <- as.vector(html_head_style_link)
+  custom_style_links <- lapply(html_head_style_link, function(x) shiny::tags$link(href = x))
+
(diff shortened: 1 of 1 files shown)


R/params.R
```diff
@@ -186,6 +186,56 @@ params_namedList <- function() {
   empty
 }
 
+
+setup_html_head <- function(html_head_style = c(),
```
aronatkins: nit: all of our other param helper functions have "params" in their name. It's not the best namespacing, but it's something.

R/params.R
```diff
@@ -194,6 +244,12 @@ params_namedList <- function() {
 #' @param shiny_args Additional arguments to \code{\link[shiny:runApp]{runApp}}.
 #' @param save_caption Caption to use use for button that saves/confirms parameters.
 #' @param encoding The encoding of the input file; see \code{\link{file}}.
+#' @param html_head_style a string or a list/vector of strings representing CSS style that will be injected in the HTML HEAD.
+#' @param html_head_script a string or a list/vector of strings represeting JS scripts that will be injected in the HTML HEAD.
```
aronatkins: nit: spelling of representing

R/params.R
```diff
@@ -186,6 +186,56 @@ params_namedList <- function() {
...
+    # Our controls are wiiiiide.
+    ".container-fluid .shiny-input-container { width: auto; }",
+    # Prevent the save/cancel buttons from squashing together.
+    ".navbar button { margin-left: 10px; }",
+    # Style for the navbar footer.
+    # http://getbootstrap.com/components/#navbar-fixed-bottom
+    "body { padding-bottom: 70px; }"
+  )
+  ## Escape is "cancel" and Enter is "save".
+  default_script <- shiny::tags$script(shiny::HTML("$(document).keyup(function(e) {\n",
```
aronatkins: Any situation where we will want to _not_ inject the defaults?

R/params.R
```diff
@@ -203,7 +259,11 @@ knit_params_ask <- function(file = NULL,
                             params = NULL,
                             shiny_args = NULL,
                             save_caption = "Save",
-                            encoding = getOption("encoding")) {
+                            encoding = getOption("encoding"),
+                            html_head_style = c(),
```
frankandrobot: Sorry, when I replied in the email, it put the comments in the wrong place.
I initially had the same idea, but as per @trestletech , the long form is preferred.

jjallaire: You could indeed go either way. I agree with @trestletech that R tends to stylistically prefer long argument lists. Even though this isn't normally good form in more well decomposed systems in R where many operations tend to be top level "one button" ones it makes some sense (also forces documentation of the parameters in .Rd as a side benefit).

jjallaire:
That's fine. We are going to ship a new version of rmarkdown CRAN around
the end of this week (to support the forthcoming IDE patch release). We are
happy to do another rmarkdown release in a month's time if this work misses
the current release.

R/params.R
```diff
@@ -250,6 +250,7 @@ params_html_head <- function(html_head_style = c(),
 #' You must take care to unsure that the URL is absolute.
 #' @param html_head_script_link same as above except that these are interpreted as SRC attributes in SCRIPT tags.
 #' You must take care to unsure that the URL is absolute.
+#' @param disable_bootstrap if true, does not inject boostrap scripts and styles in the HTML HEAD.
```
trestletech: I don't know what the common formatting is for this package, but you could consider styling with `if \code{TRUE}, does not...` Probably more important to align with the conventions of the package but just making sure you're aware of the option.

PR #874, opened 2016-11-17
No longer need the mark_utf8() trick after yaml 2.1.15 has been released

yihui:
Unfortunately I forgot to do the same thing in knitr (https://github.com/yihui/knitr/commit/dd851513aa9bd88d5f1c4d455d0bf88d128bb833) when I released it last week, so knitr and rmarkdown have to go to CRAN at the same time next time.

### DIFF
--- a/R/params.R
+++ b/R/params.R
@@ -2,7 +2,7 @@
 knit_params_get <- function(input_lines, params) {
 
   # read the default parameters and extract them into a named list
-  knit_params <- mark_utf8(knitr::knit_params(input_lines))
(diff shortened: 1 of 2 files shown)


jjallaire:
I need to release rmarkdown to CRAN in the next few days to support our RStudio patch release (otherwise we'll have to hold the patch release). Should I take this PR now or wait until after the CRAN release?

yihui:
You can merge it, and I'll push a patch release of knitr to CRAN in the next few days as well.

PR #877, opened 2016-11-18
should update the yaml package version imported

shrektan:
ref #874

BTW, it's version 2.1.14, not 2.1.15. See https://cran.r-project.org/web/packages/yaml/index.html

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -63,7 +63,7 @@ Imports:
     utils,
     grDevices,
     knitr (>= 1.14),
-    yaml (>= 2.1.5),
+    yaml (>= 2.1.14),
     htmltools (>= 0.3.5),
     caTools,
     evaluate (>= 0.8),


jjallaire:
@yihui Could you review and merge this?

jjallaire:
@yihui I think I've identified a problem with updating the YAML requirement: this will effectively push the minimum R version for rmarkdown and R Notebooks to 3.2 (well ahead of the 3.0 that we advertise). This is because when a user goes to install rmarkdown it will attempt to install the latest yaml package, the binaries for which are only available for r-release and r-oldrel. We typically work around this for knitr and rmarkdown by specifying `type = "source"` in the RStudio auto-install profile, however with YAML we need to build native code which will make this fail on most Mac and Windows systems.

So I think the only choice is to conditionalize our behavior in both knitr and rmarkdown based on the version of yaml that we are running against, and to keep our minimal required yaml version where it is. Let me know what you think.

yihui:
Okay, I will use the workaround conditional on the yaml version.

PR #898, opened 2016-12-01
add stringi as dependency

bquast:
version 1.2 won't install without

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -70,7 +70,8 @@ Imports:
     base64enc,
     jsonlite,
     rprojroot,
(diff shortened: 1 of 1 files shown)


bquast:
related to this, it might be useful to add release tags, currently the last tag is v0.1.2

jjallaire:
We don't depend directly on either **stringi** or **stringr**, rather **knitr** depends on **stringr** which in turn depends on **stringi**. So in short the knitr dependency we already have covers this.

bquast:
Ah that makes sense. I should've looked more closely.

Issue #897, opened 2016-11-30
lang option in front matter not properly recognized


caesarivs:
I'm trying to produce a PDF file in spanish, so I'm using the following front matter according to the [documentation](http://rmarkdown.rstudio.com/pdf_document_format.html#latex_options).

```
---
title: "A title"
output: 
  pdf_document: 
    keep_tex: yes
lang: "spanish"
---
```

When I try to knit the file, I get the following error:

```
! Package babel Error: You haven't specified a language option.

See the babel package documentation for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.343 ...ry to proceed from here, type x to quit.}

pandoc: Error producing PDF
Error: pandoc document conversion failed with error 43
```

Checking the `.tex` file the following is produced (concerning the language):

```
\ifnum 0\ifxetex 1\fi\ifluatex 1\fi=0 % if pdftex
  \usepackage[shorthands=off,main=]{babel}
\else
  \usepackage{polyglossia}
  \setmainlanguage[]{}
\fi
```

If I manually edit the file and put `spanish` in the `main=` I get the PDF as expected.

I'm using latest RStudio v1.0.44 in the following system:

```
> sessionInfo()
R version 3.3.2 (2016-10-31)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 16.04.1 LTS

locale:
 [1] LC_CTYPE=es_CO.UTF-8       LC_NUMERIC=C               LC_TIME=es_CO.UTF-8        LC_COLLATE=es_CO.UTF-8    
 [5] LC_MONETARY=es_CO.UTF-8    LC_MESSAGES=es_CO.UTF-8    LC_PAPER=es_CO.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=es_CO.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
 [1] backports_1.0.4 magrittr_1.5    rprojroot_1.1   htmltools_0.3.5 tools_3.3.2     yaml_2.1.14     Rcpp_0.12.8     stringi_1.1.2  
 [9] rmarkdown_1.2   knitr_1.15.1    stringr_1.1.0   digest_0.6.10   evaluate_0.10  
```

caesarivs:
[...]

PR #901, opened 2016-12-02
Update pdf_document.Rd

caesarivs:
Included some examples for the "lang" variable, since it could lead users to get confused and use invalid values (e.g. LaTeX's babel language names) according to #897

### DIFF
--- a/R/pdf_document.R
+++ b/R/pdf_document.R
@@ -56,7 +56,7 @@
 #' Available metadata variables include:
 #'
 #' \describe{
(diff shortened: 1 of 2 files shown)


yihui:
Thanks! The Rd files are automatically generated by roxygen2 (as indicated in https://github.com/rstudio/rmarkdown/blob/master/man/pdf_document.Rd), so could you modify the R source? https://github.com/rstudio/rmarkdown/blob/master/R/pdf_document.R

caesarivs:
@yihui done

Issue #903, opened 2016-12-04
render_site with Rmd parameters


zeehio:
I started to use the `rmarkdown::render_site` function. At some point I had an `Rmd` file with parameters, and I wanted to render it several times with different options.

To do that, I adapted a custom site generator. I would like to know if you are interested in a pull request or at least export some functions so I don't need to use `:::`.

# render_site with rmarkdown parameters

I wrote a custom site generator named `site_with_params` based on [`default_site`](https://github.com/rstudio/rmarkdown/blob/master/R/render_site.R#L154). The main changes with respect to the default_site are:

- The yaml header of `index.Rmd` needs to have `site: site_with_params` to use this site generator.
- An additional `files` section to `_site.yml` is required, that states how html files need to be rendered. An example of `_site.yml` is shown here (note the "files" section):

```yaml
name: "my-website"
navbar:
  title: "My Website"
  left:
    - text: "Home"
      href: index.html
    - text: "Domestic results"
       href: domestic.html
    - text: "Foreign results"
      href: foreign.html
files:
  index.html:
    src: index.Rmd
  domestic.html:
    src: domestic_foreign.Rmd
    params:
      type_of_flight: "domestic"
  foreign.html:
    src: domestic_foreign.Rmd
    params:
      type_of_flight: "foreign"
```

# Questions

- Would you would be interested in a pull request offering this "site with parameters" feature?
- In case you are not interested, would it be possible to export `rmarkdown:::site_config`, `rmarkdown:::input_as_dir`, `rmarkdown:::knitr_files_dir`, `rmarkdown:::dir_exists`, `rmarkdown:::copy_site_resources` and `rmarkdown:::knitr_root_cache_dir` so it is easy to provide custom site generators in other packages?

# Actual code

The code I wrote is adapted from `default_site` and uses several non-exported rmarkdown functions. 

This could be submitted as a PR if there is interest:

```r
#' Generate site with Rmd files with parameters
#' To use this site generator set \code{site: rmarkdown::site_with_params}
#' in the yaml header of the \code{index.Rmd} file.
#' Then in the \code{_site.yml} file, add a files section:
#'
#' \code{
#' files:
#'   index.html:
#'     src: index.Rmd
#'     params:
#'       opt1: Val1
#'   about.html:
#'     src: about.Rmd
#'     params:
#'       opt1: Val2
#' }
#'
#' @param input
#'
#' @export
site_with_params <- function(input, encoding = getOption("encoding"), ...) {

[...]

PR #904, opened 2016-12-04
render_site compatibility with Rmd params.

### LINKED ISSUE #892: Use spin with render_site
Building a site from RStudio runs `rmarkdown::render_site(encoding = 'UTF-8')`, which renders all `.Rmd` files in the root directory. 

It will not render any `.R` files, which ordinarily is fine, since those don't typically produced HTML output. However, `rmarkdown::spin()` can knit any files—including `.R` files—and keeping R scripts as `.R` instead of `.Rmd` [can be advantageous](http://deanattali.com/2015/03/24/knitrs-best-hidden-gem-spin/).

Right now, if a `.R` script is included in a website project (like the minimal one below), it will not get knitted.

```r
#' ---
#' title: "Testing stuff"
#' output: html_document
#' ---

#' Here's some R code
2 + 2
```

It might be helpful to be able to specify a list of knittable files in `_site.yml` or somewhere else so that these non-`.Rmd` files are rendered when the site is built with `render_site()`. It is currently possible to include `.R` files in the built website with the `include:` YAML key, but including files only copies the raw file to the final site and does not render them.

zeehio:
This PR provides a solution to #903 and #892 as as a side effect.

@jjallaire I used "output_files" as you suggested. Feel free to suggest changes, although I do not have much more time to devote to this.

## Highlights:

 - `render_site` understands the `output_files` field in the `_site.yml`. If `output_files` is missing
   it uses the Rmd file with the same name. Example of `output_files` field:
 ````yaml
output_files:
  flights_domestic.html:
    src: "flights_analysis.Rmd"
    params:
      flight_type: "domestic"
  flights_international.html:
    src: "flights_analysis.Rmd"
    params:
      flight_type: "international"
````

- The `output_files` feature can be used as well to set as source for an html file a R script that will be spinned:

 ````yaml
output_files:
  flights_domestic.html:
    src: "flights_analysis.R"
````
- `render_site` also understands the `autospin` field in `_site.yml`. If `autospin: true` is set, `render_site` will render all R scripts spinning them first. (closes #892)

- Tests have been enhanced a bit to cover these features and documentation has been added as well to the `?render_site` man page.

I already signed the contribution agreement when I submitted my previous PR (#647)

Comments, patches and merges are very welcome

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,12 @@ rmarkdown 1.13
 
 - TOC items are not correctly indented when `toc_float` is enabled for the `html_document` format (thanks, @carolynwclayton #1235 and @RLesur #1243).
 
+- `render_site` can spin R source files. Use `autospin: true` in the `_site.yml`.
+  (#892, @zeehio)
+
+- `render_site` can render Rmd files with parameters. Use the new `output_files`
+  field in `_site.yml`. See `?render_site` for further details. (#903, @zeehio)
+
 
 rmarkdown 1.12
 ================================================================================
--- a/R/render_site.R
+++ b/R/render_site.R
@@ -91,6 +91,27 @@
 #' \code{MASS::filter}. The default behaviour of \code{render_site} is to use a
 #' common R session.
 #'
+#' \code{autospin: true} causes \code{.R} files to be spinned and rendered
+#' (as well as \code{.Rmd} files). If \code{autospin} is set to false (the default),
+#' it is still possible to spin specific R files by using the \code{output_files} field.
+#'
+#' \code{output_files} can be used to describe which Rmd file is used to render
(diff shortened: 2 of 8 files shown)


zeehio:
Hi,

Sorry for not coming back earlier, we are all a bit too busy I guess...

I rebased this rather old PR that offers the possibility to render Rmd files with parameters with `render_site`, as well as the option to `autospin` source R files into Rmd files.

Would you still be willing to review and merge this?

Thanks!

zeehio:
There is a breaking change in this PR, as it changes the `generator$render` function so the `input` argument can be interpreted as an output file. This may have many consequences, so I prefer to close this PR. The branch is there, so if you are interested in taking some things feel free to do so.

The autospin feature was nice, so I will try to create a new PR for that exclusively

PR #908, opened 2016-12-13
Support embedding raw RTF content in RTF output

yihui:
I didn't add a NEWS item for this PR because I guess only one of our customers is interested in it. There may be more, but let's ask them to test it more thoroughly before we publicize this feature.

Basically this allows users to embed raw RTF content generated by other packages (e.g. the rtf package) in R Markdown. The idea is similar to HTML widgets: first the raw content is extracted to bypass Pandoc, then reinsert it to the final output. Nathan has a full example here: https://github.com/rstudio/sol-eng-sales/tree/master/vignettes/rtf

### DIFF
--- a/R/html_document_base.R
+++ b/R/html_document_base.R
@@ -106,13 +106,7 @@ html_document_base <- function(smart = TRUE,
                                         lib_dir,
                                         output_dir))
 
-    # The input file is converted to UTF-8 from its native encoding prior
-    # to calling the preprocessor (see ::render)
-    input_str <- readLines(input_file, warn = FALSE, encoding = "UTF-8")
-    preserve <- extractPreserveChunks(input_str)
-    if (!identical(preserve$value, input_str))
-      writeLines(preserve$value, input_file, useBytes = TRUE)
-    preserved_chunks <<- preserve$chunks
+    preserved_chunks <<- extract_preserve_chunks(input_file)
 
     args
   }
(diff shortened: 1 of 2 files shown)


R/html_document_base.R
```diff
@@ -184,3 +178,13 @@ html_document_base <- function(smart = TRUE,
   )
 }
 
+extract_preserve_chunks <- function(input_file, extract = extractPreserveChunks) {
+  # The input file is converted to UTF-8 from its native encoding prior
+  # to calling the preprocessor (see ::render)
+  input_str <- readLines(input_file, warn = FALSE, encoding = "UTF-8")
+  preserve <- extract(input_str)
+  if (!identical(preserve$value, input_str))
+    writeLines(enc2utf8(preserve$value), input_file, useBytes = TRUE)
```
yihui: That was intentional. We should make sure `input_file` is always encoded in UTF-8, and `enc2utf8()` here should not be harmful. If `preserve$value` was converted to the native encoding for some reason on Windows, we would be doing the wrong thing by `writeLines(preserve$value, input_file, useBytes = TRUE)`.

kaigu1990:
Hello, 

I want to change orientation of RTF from rmarkdown, because I'd like to creat RTF report instead of SAS. I have tried to insert RTF code in the header of RTF file to change orientation, and it can be well. But it's not working when I use `knitr::raw_output()` to insert RTF code in rmarkdown. I think we should insert RTF code in the beginning of RTF, not  in the body. 
Cound you give some advise to deal with this question. 

Thanks!
Gu Kai

yihui:
@kaigu1990 This is not the best place to ask questions. Please follow [the issue guide](../issues/new) to if you have questions. Thanks!

Issue #910, opened 2016-12-15
bugs in pdf_document extra_dependencies


haozhu233:
Hi, I discovered a few bugs in the `extra_dependencies` option in `pdf_document`. I found them out in an afford trying to load a latex package from a R function while using rmarkdown + knitr. 

1. [Fixed in PR #911 ] First of all, in the current github version of rmarkdown (https://github.com/rstudio/rmarkdown/blob/master/R/pdf_document.R#L176-L187), for `pdf_documents`, the content of `extra_dependency` is passed to a list called `format_deps`. This `format_deps` will not be put into real use unless `has_latex_dependencies(knit_meta)` is `TRUE`.  I think this bug will prevent people from using `extra_dependency` in most cases. 

~~2. Also, if people use the results of `latex_dependency()` directly to `extra_dependency`, they will receive the following error message:~~
~~YAML header:~~
```
---
title: "Untitled"
output: 
  pdf_document:
    keep_tex: true
    extra_dependencies: "`r rmarkdown::latex_dependency('multirow')`"
---
```
~~Error:~~
```
processing file: Untitled.Rmd
Quitting from lines 2-8 (Untitled.Rmd) 
Error in vapply(x, format_sci_one, character(1L), ..., USE.NAMES = FALSE) : 
  values must be length 1,
 but FUN(X[[2]]) result is length 0
Calls: <Anonymous> ... paste -> hook -> .inline.hook -> format_sci -> vapply
Execution halted
```

3. [Fixed in PR #911 ] Even if people find a way around to get the results of `latex_dependency` to `pdf_document`, the results of `latex_dependency`, as a list of 3, should not be append to `format_deps` directly (https://github.com/rstudio/rmarkdown/blob/master/R/pdf_document.R#L177). We should put a list round it so the whole list can be passed as an element. 

I have worked around solutions to question 1 & 3 and I will submit a PR very soon. Also, I added a feature so people can use `extra_dependencies` in the YAML section more easily.

yihui:
For 2, instead of `` `r ` ``, you should use `!expr `. I know it is confusing, but `` `r ` `` is knitr's syntax for including inline output (typically it is supposed to be a character string). If you want to return an R value that is passed to the `pdf_document()` function, you need to use the yaml syntax, which is `!expr` followed by R code to generate an arbitrary value.

haozhu233:
@yihui 👍 Thanks for letting me know that! Now my problem is solved. :)

yihui:
[...]

PR #911, opened 2016-12-15
bug fixed for pdf_document extra_dependencies + a new feature

haozhu233:
I fixed bug 1 & 3 in #910. I didn't touch bug 2 as I'm not feeling confident enough to make changes to that part..

Also, as I said in the issue, I added a new feature that can translate a string of latex package names (but not options) to latex_dependencies so people can load latex packages (without further settings) in the follow way.
```
---
title: "Untitled"
output: 
  pdf_document:
    keep_tex: true
    extra_dependencies: multirow, threeparttable
---
```
In the future, the `string_to_latex_dependencies` function can also be extended to accept latex package options. I'm thinking a format below might be useful. However, if someone can fix bug 2 in #910, that might not be necessary. 
```
extra_dependencies: multirow[latex package options], threeparttable
```

### DIFF
--- a/R/latex_dependencies.R
+++ b/R/latex_dependencies.R
@@ -61,15 +61,20 @@ has_latex_dependencies <- function(knit_meta) {
   has_dependencies(knit_meta, "latex_dependency")
 }
 
-# Convert a string to a list of basic latex dependencies (no options)
-string_to_latex_dependencies <- function(x){
-  if(grepl("\\[|\\]", x)){
-    x <- gsub("\\[(.*?)\\]", "", x)
-    warning("You have to use `latex_dependency` to set options for latex",
-            " dependencies. Options set here in the string will be",
-            " ignored. ")
+# Convert a yaml structured input to a list of latex dependencies
+yaml_to_latex_dependencies <- function(x){
+  x_yaml <- try(yaml_load_utf8(x))
+  if(class(x_yaml) == "try-error"){
+    stop("Please check the yaml syntax of extra_dependencies.")
   }
-  latex_package_names <- strsplit(x, ",|\\s")[[1]]
-  latex_package_names <- latex_package_names[latex_package_names != ""]
-  lapply(latex_package_names, latex_dependency)
+  if(is.list(x_yaml)){
+    lapply(seq(length(x_yaml)), function(i){
+      latex_dependency(names(x_yaml)[i], x_yaml[[i]])
+      })
+  }else{
+    if(length(x_yaml) == 1 && grepl(",", x_yaml)){
+      x_yaml <- yaml_load_utf8(paste0("[", x_yaml, "]"))
+    }
+    lapply(x_yaml, latex_dependency)
+  }
 }
--- a/R/latex_dependencies.R
+++ b/R/latex_dependencies.R
@@ -60,3 +60,16 @@ validate_latex_dependency <- function(list) {
 has_latex_dependencies <- function(knit_meta) {
   has_dependencies(knit_meta, "latex_dependency")
 }
+
+# Convert a string to a list of basic latex dependencies (no options)
+string_to_latex_dependencies <- function(x){
+  if(grepl("\\[|\\]", x)){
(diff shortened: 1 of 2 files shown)


haozhu233:
Hello, bugs reported in #910 are still there. Could anyone review this PR? Thanks! ;)

jjallaire:
@yihui Could you review and merge if appropriate?

yihui:
I can only review it after rstudio::conf.

jjallaire:
Okay, no worries. @haozhu233 we'll take a look at this in a couple of weeks.

yihui:
In general, I'd avoid processing structured data using regular expressions, since it is not "bulletproof". In this case, I'd suggest you take advantage of the YAML syntax so that you don't really need `string_to_latex_dependencies` (splitting by comma), e.g. the syntax could be like

```yaml
---
title: "Untitled"
output: 
  pdf_document:
    keep_tex: true
    extra_dependencies: [multirow, threeparttable]
---
```

i.e. `extra_dependencies` is a character vector of LaTeX package names. If users want to specify options, they can use a _named_ list, e.g.

```yaml
---
title: "Untitled"
output: 
  pdf_document:
    keep_tex: true
    extra_dependencies:
      multirow: "foo, bar=false"
      threeparttable: "hello, world"
---
```

In this case, you can easily convert this list to `\usepackage[options]{name}`.

Does that make sense?

haozhu233:
@yihui Yes, it totally makes sense. I will see what I can do and try to make it happen.

haozhu233:
Hi, @yihui, I made some improvements based on your suggestions last night. Now `extra_dependencies` can take the following inputting formats:

```
extra_dependencies: [multirow, threeparttable]
extra_dependencies: multirow, threeparttable
extra_dependencies:
    multirow: "foo, bar=false"
    threeparttable: hello, world
```

yihui:
I feel this is not quite right yet. First, I would not support this:

```
extra_dependencies: multirow, threeparttable
```

since it looks hackish (you add `[ ]` to the string and parse it as YAML).

Second, I think a few things in `yaml_to_latex_dependencies()` can be improved, and a few things are not what I imagined.

Third, you moved some code from inside `if (has_latex_dependencies(knit_meta)) { }` to the outside in pdf_document.R. I feel that is not necessary.

If you don't mind, I think it will be much easier for me to rewrite it than asking you to change every bit.

haozhu233:
@yihui Haha, I knew it was a hacky approach. 😝

Sure, I'm looking forward to seeing how you are going to implement this function. :)

haozhu233:
close this pr as #910 is fixed by #944

R/latex_dependencies.R
```diff
@@ -60,3 +60,21 @@ validate_latex_dependency <- function(list) {
 has_latex_dependencies <- function(knit_meta) {
   has_dependencies(knit_meta, "latex_dependency")
 }
+
+# Convert a yaml structured input to a list of latex dependencies
+yaml_to_latex_dependencies <- function(x){
+  x_yaml <- try(yaml_load_utf8(x))
+  if(class(x_yaml) == "try-error"){
```
yihui: Just for future records, this is the wrong way to check classes. The correct way is `inherits(x, class)`, because `class(x)` can be a vector.

R/latex_dependencies.R
```diff
@@ -60,3 +60,21 @@ validate_latex_dependency <- function(list) {
...
+    stop("Please check the yaml syntax of extra_dependencies.")
+  }
+  if(is.list(x_yaml)){
```
yihui: In the future, if you submit pull requests, I'd be happy to see more spaces, e.g. after `if` and around `else`

R/latex_dependencies.R
haozhu233: Ah, I have been using class() directly in many other cases. Thank you for letting me know the correct way!!

R/latex_dependencies.R
haozhu233: Sure! ;)

PR #912, opened 2016-12-15
Better NULL support

trestletech:
Touches #3050

I'm not entirely clear whether this PR is worth taking, but the work is done so I'll let you make the call.

This avoids clobbering the parameter field when the default value is set to NULL. This likely isn't a pattern we would recommend at this point, but I'm opening this PR because I'd argue that it makes our behavior more consistent, even if this isn't an encouraged/documented practice.

Before this change, a document containing:

```
params:
  textField:
    value: NULL
    input: text
```

would be unable to "Knit with parameters" with a custom/overridden value of `textField`. If you attempted, you would get the following error:

```
Error in knit_params_get(input_lines, params) : 
  render params not declared in YAML: textField
Calls: <Anonymous> -> knit_params_get
Execution halted
```

because the parameter was being removed when we attempted to make a NULL assignment.

This avoids deleting the parameter entry from the list in the event that the parameter value is NULL. But I have not fully worked out the implications of allowing a NULL to propagate through the pipeline beyond a couple of simple tests like this.

### DIFF
--- a/R/params.R
+++ b/R/params.R
@@ -1,4 +1,5 @@
-
+# Get the combined list of parameters given the defaults provided in the lines
+# of the RMD file and the list of overridden parameters passed in.
 knit_params_get <- function(input_lines, params) {
 
   # read the default parameters and extract them into a named list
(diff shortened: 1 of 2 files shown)


trestletech:
@aronatkins I considered both TODOs that you'd suggested in #913.

Regarding ensuring that the numeric values be within the appropriate range, that would involve parsing out the parameters passed in in a completely different order, so it would be a pretty invasive change. Empirically the slider is able to handle values that are above/below the specified range and it gracefully handles such values. 

I agree that the refactor you suggested would be valuable if/when we go to inject another param into that function, bloating it further. I'm inclined to say that as it stands it might not be enriched much by this change.

So I'm voting this goes in as-is, but let's vchat through it if you have any reservations about these.

yihui:
The same fix was proposed in #1115 and merged. Apologies that we forgot this PR.

PR #913, opened 2016-12-15
Allow the suppression of default params in interactive configurations

trestletech:
(This will conflict with #912, but if we want to take both I'll resolve the conflicts.)

This adds a `show_default` config for parameters in an RMD. When set to `TRUE`, it will prevent the default from being shown during an interactive parameter configuration process. Non-interactive knits of the doc are not affected by this change.

The use case we'd like to support is to allow a developer to communicate that users should be responsible for configuring certain parameters from scratch, while still allowing them to iterate quickly when locally knitting the document on their machines using the defaults.

In my opinion, this change behaves well for:

 - Text
 - Date
 - DateTime (to the extent that it's supported at all)
 - Radio buttons
 - Select with `multiple: true`
 - file (no effect)

It behaves acceptably for:

 - checkbox (defaults to unselected, or `FALSE`)

And it does not perform as one might hope for: 

 - select with `multiple: false` (the first option is still selected)
 - numeric selector (set to 0 rather than leaving it empty)
 - slider (it attempts to set the default to 0)

As best I can tell, these are limitations in Shiny's input functions. We could consider adding an empty first choice in the "select w/ multiple: false" case, but that felt a little too invasive IMHO.

As an example:

```
params:
  numericField:
    value: 13
    input: numeric
    show_default: false
  textField:
    value: "abc"
    show_default: false
  selector:
    value: east
    choices: [east, west, north, south]
    input: select
    multiple: true
    show_default: false
  slider:
    value: 10
    min: 0
    max: 20
    input: slider
    show_default: false
  date:
    value: !r as.Date("2015-01-01")
    input: date
    show_default: false
  datetime:
    value: !r as.POSIXct("2015-01-01 12:30:00")
    input: datetime
    show_default: false
  checkbox:
    value: TRUE
    input: checkbox
    show_default: false
  radio:
    value: "option1"
    input: radio
    choices: [option1, option2]
    show_default: false
```

produces:

screen shot 2016-12-15 at 2 41 08 pm

### DIFF
--- a/R/params.R
+++ b/R/params.R
@@ -48,31 +48,54 @@ params_label <- function(inputControlFn, param) {
   label
 }
 
-params_value_to_ui <- function(inputControlFn, value) {
+params_value_to_ui <- function(inputControlFn, value, showDefault) {
+  if (is.null(showDefault)) {
+    showDefault <- TRUE
+  }
+
+  isNumericInput <- identical(inputControlFn, shiny::numericInput) ||
+    identical(inputControlFn, shiny::sliderInput)
+
   if (identical(inputControlFn, shiny::fileInput)) {
     NULL
   } else if (identical(inputControlFn, shiny::textInput)) {
     ## TODO: if long input, maybe truncate textInput values for display
 
-    classes <- class(value)
-    if ("POSIXct" %in% classes) {
-      as.character(value)
+    if (showDefault){
+      classes <- class(value)
+      if ("POSIXct" %in% classes) {
+        as.character(value)
+      } else {
+        value
+      }
     } else {
-      value
+      NULL
     }
   } else if (is.null(value)) {
-    numerics <- c(shiny::numericInput,
-                  shiny::sliderInput)
     # The numerics can't deal with a NULL value, but everything else is fine.
-    if (identical(inputControlFn, shiny::numericInput) ||
-        identical(inputControlFn, shiny::sliderInput)) {
+    if (isNumericInput) {
       0
     } else {
       value
     }
   } else {
-    ## A type/control that doesn't need special handling; just emit the value.
-    value
+    if (showDefault){
+      ## A type/control that doesn't need special handling; just emit the value.
+      value
+    } else {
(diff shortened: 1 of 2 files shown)


jjallaire:
We're just about to submit **rmarkdown** to CRAN (~ 3 days from now) and we've had very little activity on master for this release so it's in a known good state (we are motivated to release to CRAN ASAP b/c there was a regression in ioslides in the last CRAN release). Ping me back after the holidays and we'll review and merge this then.

trestletech:
Polite ping, @jjallaire

jjallaire:
I'm happy to merge this (and the other related PR) if @aronatkins gives the thumbs up (he is the author of the code which is being modified here).

R/params.R
```diff
@@ -48,31 +48,54 @@ params_label <- function(inputControlFn, param) {
   label
 }
 
-params_value_to_ui <- function(inputControlFn, value) {
+params_value_to_ui <- function(inputControlFn, value, showDefault) {
```
aronatkins: Would it be valuable to have two separate functions: 
* given an `inputControlFn`, return a "default" value for that control.
* given an `inputControlFn`, massage a value appropriately for that control (the current implementation).
this separation might make for easier-to-grok functions and simpler control structure in each (though some repetition).

R/params.R
```diff
@@ -48,31 +48,54 @@ params_label <- function(inputControlFn, param) {
...
     }
   } else {
-    ## A type/control that doesn't need special handling; just emit the value.
-    value
+    if (showDefault){
+      ## A type/control that doesn't need special handling; just emit the value.
+      value
+    } else {
+      if (isNumericInput){
+        0
```
aronatkins: This might be an existing bug, but if the numeric inputs have range specifications, we should ensure that the default value (both here and above) is within the range.

jjallaire:
Okay, I'll merge now. @trestletech Could you include the cleanup items in the PR that brings in the support for `null` parameters?

PR #926, opened 2016-12-27
New `citeproc` agrument in YAML header. Resolves #831.

### LINKED ISSUE #831: Applying a pandoc filter after pandoc-citeproc
Hi there, I have spent several hours looking for a solution but have come up empty. Any ideas would be greatly appreciated.

I'm creating documents that need citations in APA format. I'm using the APA citation style, which works great with one exception:

In APA guidelines 6th edition, Section 6.12 mention that:

> Precede the final name in a multiple-author citation in a running text by the word and. In parenthetical material, in tables and captions, and in the reference list, join the names by an ampersand (&).
> 
> as Kurtines **and** Szapocznik (2003) demonstrated
> as has been shown (Kurtines **&** Szapocznik 2003)

However, with pandoc-citeproc all "and" are rendered as ampersands.

> as Kurtines **&** Szapocznik (2003) demonstrated
> as has been shown (Kurtines **&** Szapocznik 2003)

I know this is because in-text citations are not supported by CSL and there is no way to define the in-text "and". I have written a simple filter that modifies the output of `pandoc-citeproc` to overcome this issue. The filter works on the AST and replaces all in-text `&` with `and`.

``` r
#!/usr/bin/env Rscript
intext_and <- function(x) {
  gsub("(\"citationMode\":\\{\"t\":\"AuthorInText\".*?\\})(.*?\\}\\],\\[.*?)(&)(.*?\\}\\]\\]\\})", "\\1\\2and\\4", x)
}

ast <- readLines(file("stdin"))
write(intext_and(ast), stdout())
```

The problem is, I need to call this filter after `--filter pandoc-citeproc`. From [another issue](https://github.com/rstudio/rmarkdown/issues/592) I get the impression the call to `pandoc-citeproc` is always appended at the end of the system call and there is no way to apply another filter after it. Is this indeed the case? Is there any way to apply a filter after `pandoc-citeproc`?

crsh:
As discussed in #831 this pull request adds a new option `citeproc` to the YAML front matter. If it is set to `TRUE` or omitted `pandoc-citeproc` is invoked as usual. Thus, the change should not break backward compatibility. If it is set to `FALSE` the `pandoc-citeproc`-filter argument is omitted from the call to `pandoc`. This makes it possible to call additional filters after adding the `pandoc-citeproc`-filter argument manually.

I deferred the automated check if `pandoc-citeproc` is needed via `citeproc_required()` to line [651](https://github.com/rstudio/rmarkdown/compare/master...crsh:master#diff-e446088a017e6c3093395c7f863509caR651) of `render.R`. Thereby the check is performed after the pre-processor of a format is invoked and enables changing the default setting for the `citeproc` option. I did not see any earlier reference to the `run_citeproc` variable so as far as I can tell this change should not have any side effects. Tests and checks are passing fine.

Let me know what you think.

### DIFF
--- a/R/output_format.R
+++ b/R/output_format.R
@@ -770,7 +770,12 @@ is_pandoc_to_html <- function(options) {
 }
 
 citeproc_required <- function(yaml_front_matter, input_lines = NULL) {
-  !is.null(yaml_front_matter$bibliography) ||
-  !is.null(yaml_front_matter$references) ||
-  length(grep("^references\\:\\s*$", input_lines)) > 0
+  (
+    is.null(yaml_front_matter$citeproc) ||
+      yaml_front_matter$citeproc
+  ) && (
+    !is.null(yaml_front_matter$bibliography) ||
+      !is.null(yaml_front_matter$references) ||
+      length(grep("^references\\:\\s*$", input_lines)) > 0
+  )
 }
--- a/R/render.R
+++ b/R/render.R
@@ -229,10 +229,6 @@ render <- function(input,
   }
   pandoc_to <- output_format$pandoc$to
 
-  # determine whether we need to run citeproc (based on whether we
-  # have references in the input)
-  run_citeproc <- citeproc_required(yaml_front_matter, input_lines)
-
   # generate outpout file based on input filename
   if (is.null(output_file))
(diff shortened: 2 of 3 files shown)


jjallaire:
@yihui Let me know what you think of this PR.

PR #932, opened 2017-01-07
Allow internationalization of code folding buttons using pandoc

walmes:
Edit default.html and codefolding.js to reveive pandoc
arguments and set text. This allow user set names for
header code folding button, his menu, and buttons before 
code chunks. User must define such variables using pandoc_args in YAML. 
A full example is below. It uses text in Portuguese and unicode
code for up and down triangles.

```
html_document:
  pandoc_args: [
      "--variable", "code_folding_button=Código",
      "--variable", "code_folding_show=Mostrar",
      "--variable", "code_folding_hide=Esconder",
      "--variable", "code_folding_show_btn='\u25BE'",
      "--variable", "code_folding_hide_btn='\u25B4'"
    ]
```

### DIFF
--- a/inst/rmd/h/default.html
+++ b/inst/rmd/h/default.html
@@ -247,6 +247,18 @@ $if(code_folding)$
   window.initializeCodeFolding("$code_folding$" === "show");
 $endif$
 });
+// Show/Hide button names internationalization.
+// Variables used in codefolding.js.
+$if(code_folding_show_btn)$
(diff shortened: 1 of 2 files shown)


jjallaire:
I think we should do this is a much more general fashion such that the rmarkdown package itself can include the appropriate translations and automatically include them based on the current locale. This might be based on the base mechanism for localizing R messages if it's straightforward enough to do so (I don't have experience with it so I'm not sure).

PR #947, opened 2017-01-27
Escape end tags in prerendered code contexts (fixes #943)

### LINKED ISSUE #943: Possible bug in shiny_prerendered runtime with selectInput
(I posted this as a question as a question on SO but am starting to think it may be a bug so posting issue here as well)
Link to SO post: http://stackoverflow.com/questions/41749076/error-in-shiny-prerendered-runtime

I have a functioning Shiny interactive document (.Rmd) file that works fine locally and on shinyapps.io. I am trying to prerender it using instructions here (http://rmarkdown.rstudio.com/authoring_shiny_prerendered.html) but am running into an unusual error when I introduce a selectInput().

**The error I am getting is:**
[lexical error: invalid character inside string.][1]

Here's a mininal code block that produces the error.

(app.Rmd)

    ---
    title: 'a title'
    output: html_document
    runtime: shiny_prerendered
    ---

    ```{r, context = 'setup', include=FALSE}
    library(shiny)

    ```

    ```{r, context='render'}
    selectInput('hello','Hi',choices=c('1','2'))
    ```

    ```{r, context='server'}

    ```

Here is my sessionInfo():

    R version 3.2.2 (2015-08-14)
    Running under: Windows 7 x64 (build 7601) Service Pack 1

    locale:
    [1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252    LC_MONETARY=English_United States.1252
    [4] LC_NUMERIC=C                           LC_TIME=English_United States.1252    

    attached base packages:
    [1] stats     graphics  grDevices utils     datasets  methods   base     

    other attached packages:
    [1] shiny_1.0.0

    [1] Rcpp_0.12.6     digest_0.6.10   rprojroot_1.1   mime_0.5        R6_2.1.2        xtable_1.8-2    backports_1.0.3
    [8] magrittr_1.5    evaluate_0.10   stringi_1.1.1   rmarkdown_1.3   tools_3.2.2     stringr_1.1.0   httpuv_1.3.3   
    [15] yaml_2.1.13     htmltools_0.3.5 knitr_1.15.1 

I've tried changing the encoding of my Rmd to UTF-8 with no luck. It looks like there is an invalid character being introduced, potentially in "input-select.R" in the shiny package (which references the "es5-shim.min.js" mentioned in the attached screenshot).

However, changing the runtime to "shiny" (from "shiny_prerendered") seems to fix the problem. Removing selectInput also fixes the problem. Is this a bug in the "shiny_prerendered" runtime or am I doing something wrong?

  [1]: https://i.stack.imgur.com/KGcOG.png

jjallaire:
@wch Could you review this pls?

### DIFF
--- a/R/shiny_prerendered.R
+++ b/R/shiny_prerendered.R
@@ -433,7 +433,7 @@ shiny_prerendered_extract_context <- function(html_lines, context) {
   }
 
   # unescape code, see https://github.com/rstudio/rmarkdown/issues/943
-  context_lines <- gsub("<\\u002", "</", context_lines, fixed = TRUE)
+  context_lines <- gsub("<\\u002f", "</", context_lines, fixed = TRUE)
   context_lines
 }
 
@@ -488,7 +488,7 @@ shiny_prerendered_append_context <- function(con, name, code) {
              paste0('<script type="application/shiny-prerendered" ',
                     'data-context="', name ,'">'),
              # escape code, see https://github.com/rstudio/rmarkdown/issues/943
-             gsub("</", "<\\u002", code, fixed = TRUE),
+             gsub("</", "<\\u002f", code, fixed = TRUE),
              '</script>',
              '<!--/html_preserve-->')
   writeLines(lines, con = con)
--- a/R/shiny_prerendered.R
+++ b/R/shiny_prerendered.R
@@ -431,6 +431,9 @@ shiny_prerendered_extract_context <- function(html_lines, context) {
     if (in_context)
       context_lines <- c(context_lines, html_lines[[i]])
   }
+
+  # unescape code, see https://github.com/rstudio/rmarkdown/issues/943
+  context_lines <- gsub("<\\u002", "</", context_lines, fixed = TRUE)
   context_lines
 }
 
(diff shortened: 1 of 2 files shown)


R/shiny_prerendered.R
```diff
@@ -431,6 +431,9 @@ shiny_prerendered_extract_context <- function(html_lines, context) {
     if (in_context)
       context_lines <- c(context_lines, html_lines[[i]])
   }
+
+  # unescape code, see https://github.com/rstudio/rmarkdown/issues/943
+  context_lines <- gsub("<\\u002", "</", context_lines, fixed = TRUE)
```
wch: I think it should be `"\\u002f"` (and same below).

jjallaire: Thanks, changed!

R/shiny_prerendered.R
```diff
@@ -431,6 +431,9 @@ shiny_prerendered_extract_context <- function(html_lines, context) {
     if (in_context)
       context_lines <- c(context_lines, html_lines[[i]])
   }
+
+  # unescape code, see https://github.com/rstudio/rmarkdown/issues/943
+  context_lines <- gsub("<\\u002f", "</", context_lines, fixed = TRUE)
```
wch: Just a note about this: I believe that unescaping is unnecessary, but it also shouldn't cause any harm.

wch:
Looks good to me.

R/shiny_prerendered.R
jjallaire: okay, I'll keep it in just to be symmetric.

PR #952, opened 2017-01-29
add sticky tabs to .tabset (url hash and active tab always correspond)

rubenarslan:
## Purpose
This should ensure that:

1. the URL hash changes when a tab is activated
2. this change is written to browser history
3. most importantly: calling up the page with this hash activates the right tab

I've been using this code in my own projects for a while now. I often generate complex analysis reports and it's very helpful to be able to link to specific sections.

It's an adapted version of Aidan Lister's stickyStuff plugin for bootstrap tabs. My changes just let it travel up the ancestors of the tabset, so that if a child tab (e.g. `.section.level3`) is activated, its parent tab is activated too.

## Test document
I've been using this simple document to test that it works, but I've tested in much bigger projects as well and I haven't noticed any problems.

```markdown
# Top heading {.tabset}

## Tab 1 {.tabset}

### Subtab 1
1

### Subtab 2
2

## Tab 2 {.tabset}

### Subtab 3
3

### Subtab 4
4
```

### DIFF
--- a/inst/rmd/h/navigation-1.1/tabsets.js
+++ b/inst/rmd/h/navigation-1.1/tabsets.js
@@ -1,3 +1,49 @@
+/**
+ * jQuery Plugin: Sticky Tabs
+ *
+ * @author Aidan Lister <aidan@php.net>
+ * adapted by Ruben Arslan to activate parent tabs too
+ * http://www.aidanlister.com/2014/03/persisting-the-tab-state-in-bootstrap/
+ */
+(function($) {
+  "use strict";
+  $.fn.stickyStuff = function() {
+    var context = this;
+    // Show the tab corresponding with the hash in the URL, or the first tab
+    var showStuffFromHash = function() {
+      var hash = window.location.hash;
+      var selector = hash ? 'a[href="' + hash + '"]' : 'li.active > a';
+      var $selector = $(selector, context);
+      if($selector.data('toggle') === "tab") {
+        $selector.tab('show');
+        // walk up the ancestors of this element, show any hidden tabs
+        $selector.parents('.section.tabset').each(function(i, elm) {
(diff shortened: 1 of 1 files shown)


jjallaire:
Really cool!

I don't know however whether this should be the default behavior for all tabsets in Rmd documents. Maybe this would be better as an htmltools based R function which injects the script into an existing page?

rubenarslan:
I like it as a default behaviour, because this is what I expected bootstrap tabs to do by default. I've read similar sentiments upvoted a lot [on Stackoverflow](http://stackoverflow.com/questions/7862233/twitter-bootstrap-tabs-go-to-specific-tab-on-page-reload-or-hyperlink). 

It's fairly lightweight (compared to rmarkdown's other dependencies), needs no configuration, does not disrupt existing workflows, and people would usually want to deploy it everywhere they use tabsets (if they want it). I see little harm :-)

If I understand htmltools correctly (I've never used it before), this would mean calling the commando to inject the dependency manually on every page, but the dependency would also be shared across files in a site? And htmltools is just the framework, so I'd have to make a htmlwidget-style package and release that? Probably wouldn't be noticed, would it?

I don't know, in my own work, I modified a template, so I don't have to drag stuff like this around, but maybe others find it less useful.

jjallaire:
I agree it would be too obscure if left in it's own package.

Here's an alternate formulation that includes it in the rmarkdown package but requires explicit enabling via the `tabset-sticky` attribute (sort of like we currently do with `tabset-fade`):

https://github.com/rstudio/rmarkdown/pull/953

rubenarslan:
Sounds like a good compromise! And easy to change, if people start clamouring for this to be the default (as if :wink:).

jjallaire:
Okay great, I'll merge the other PR!

PR #978, opened 2017-02-23
use 'default' theme for html_notebook

kevinushey:
This PR ensures that `"default"` is used as the highlight theme for `html_notebook`, to match that used with `html_document`. (Some users have been surprised / confused as to why the styles are different)

Not sure if we want to take this PR (I somewhat prefer `textmate` highlighting here + changing a default parameter value might be seen as a negative change), but opening this for discussion anyhow.


jjallaire:
My reason for using textmate was to match the default syntax highlighting of the IDE (which is what most see in the Rmd editor). I suspect we can't win here as now some users will be surprised that the notebook HTML has different highlighting than the HTML output.

kevinushey:
That's reasonable then -- let's just close this.

PR #979, opened 2017-02-24
avoid line breaks between non-print data.table expressions

kevinushey:
This PR fixes an issue where multiple consecutive non-printing `data.table` statements would print with spaces between the statements, e.g.:

screen shot 2017-02-24 at 11 28 54 am

There's a chance that we might want to change this downstream in `knitr`; e.g. by filtering out any `NULL` or empty outputs, but this should be sufficient for now.

### DIFF
--- a/R/html_document.R
+++ b/R/html_document.R
@@ -307,10 +307,11 @@ html_document <- function(toc = FALSE,
       try(action())
   }
 
-  # capture the source code if requested
   source_code <- NULL
   source_file <- NULL
   pre_knit <- function(input, ...) {
+
+    # capture the source code if requested
     if (code_download) {
       source_file <<- basename(input)
       source_code <<- paste0(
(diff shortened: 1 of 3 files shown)


jjallaire:
It seems like we might want this to be a part of render rather than html_document (as this leaves the issue unfixed in all but the html_document format). We could also consider adding it to html_document_base (which just about every other html format inherits from). The problem is that other formats may have their own evaluate hooks and I'm not sure if these stack. Net: this may need a fix in knitr rather than rmarkdown.

yihui:
Does the chunk option `collapse = TRUE` work?

kevinushey:
It looks like `collapse = TRUE` does indeed work here, but even more to the point, we can just return `invisible(list())` here instead (since we take over `knit_print`). Thanks!

PR #981, opened 2017-02-25
Mention that postscript probably should not be cropped

yihui:
Closes yihui/knitr#1365


jjallaire:
Could we auto-disable it for the postscript graphics device?

jjallaire:
I'll merge this then leave it to you whether and how we should do an auto-disable.

yihui:
I thought about it before submitting the PR. The problem is I don't know whether it is always a bad idea to crop postscript figures. I can disable cropping for postscript, but it will be difficult to enable it if cropping actually works (will require a separate chunk option like `crop.eps`), so I'd rather just document it. I believe the number of people using postscript should be fairly small.

jjallaire:
Okay, sounds good.

PR #990, opened 2017-03-08
Expose --section-divs pandoc argument

### LINKED ISSUE #987: Expose --section-divs pandoc option when generating html_document or html_fragment
Currently the --section-divs option is hardcoded in html_document.R, which means that there isn't a simple way to remove the divs that wrap each h1.

This is particularly relevant for html_fragments which might be imported into various places that the divs might cause issues with.

I'll attempt a pull request for this.

mungojam:
Exposes the section-divs argument so that it can be optionally excluded
when calling pandoc. This allows for generating HTML output without
surrounding h1s with div tags.

Resolves #987


jjallaire:
Thanks!

Issue #1026, opened 2017-04-30
 asis_output overrides header-includes in .Rmd file


hughjonesd:
(Moved over from yihui/knitr#1397)

In a .Rmd file, if you include a list of `rmarkdown::latex_dependency` objects in the `meta` argument to `asis_output`, this will override any `\usepackage` statements in the `header-includes` argument. See the attached mwe. This came up in hughjonesd/huxtable#7.

[knitr-bug-mwe.Rmd.zip](https://github.com/yihui/knitr/files/965951/knitr-bug-mwe.Rmd.zip)

A workaround is to use the `includes` argument to `pdf_document`.

The issue is that if a latex_dependency is present, pandoc is called with argument --include-in-header XXX, where XXX is a temporary file containing the stringified latex dependency. When this happens, it overrides the header-includes variable in the YAML front matter. 

I am trying to work out how to merge values from the command line and the YAML front matter, but I am not sure there is a way. In which case, one approach would be to take values from the YAML front matter and write them to the same file with the latex dependencies.

Another approach would be just to use `includes` - I get the impression that `header-includes` is a bit deprecated?

haozhu233:
https://github.com/rstudio/rmarkdown/blob/e672d4146bc3341a53a57a24b76ade02b07ce2ad/R/pdf_document.R#L184-L185

@hughjonesd is absolutely right on this. These two lines above are where things get wrong. I'm also trying to find out a solution for this issue.

haozhu233:
Ah, Just noticed @hughjonesd's new PR after I pushed my fix. I think it also works. Haha ;)

PR #1027, opened 2017-05-01
include header-includes in file for pandoc --include-in-header

hughjonesd:
Possible fix for rstudio/rmarkdown#1026


jjallaire:
@yihui I know less about this that you do. Could you review?

PR #1032, opened 2017-05-03
close the session prior to stopping the server

aronatkins:
Close the session when finalizing parameter choices and preparing to stop the
Shiny app; this should close the websocket gracefully in stead of leaving the
connection to be severed when the httpuv server is stopped.


jcheng5:
LGTM

PR #1035, opened 2017-05-04
Add test samples for inline code

rgaiacs:
**Note:** This is my first bug report to RStudio. Any comment will be welcome.  

When using RStudio 1.0.143 and R 3.3.3 for R Notebooks I noticed a bug with inline code so I decided to submit this pull request with the minimal example than I come across.

# Description

Inline code should use the value of variables taking in consideration only the code blocks before it. For R Notebooks, inline code use the value of variables after all code blocks were executed. **R Markdown (knit) doesn't have this bug.**

# Expectation

Have a look at `tests/testthat/resources/inline-code-notebook.nb.html`, included in this pull request, for the expected HTML generated from `tests/testthat/resources/inline-code-notebook.Rmd`.

# Current Outcome

Instead of

~~~
<!-- rnb-text-begin -->
<!-- rnb-text-end -->
<!-- rnb-chunk-begin -->
<!-- rnb-source-begin eyJkYXRhIjoiYGBgclxuYSA8LSBUUlVFXG5gYGAifQ== -->
<pre class="r"><code>a &lt;- TRUE</code></pre>
<!-- rnb-source-end -->
<!-- rnb-chunk-end -->
<!-- rnb-text-begin -->
<p><code>a</code> is TRUE.</p>
<!-- rnb-text-end -->
<!-- rnb-chunk-begin -->
<!-- rnb-source-begin eyJkYXRhIjoiYGBgclxuYSA8LSBGQUxTRVxuYGBgIn0= -->
<pre class="r"><code>a &lt;- FALSE</code></pre>
<!-- rnb-source-end -->
<!-- rnb-chunk-end -->
<!-- rnb-text-begin -->
<p>Now <code>a</code> is FALSE.</p>
<!-- rnb-text-end -->
~~~

we have

~~~
<!-- rnb-text-begin -->
<!-- rnb-text-end -->
<!-- rnb-chunk-begin -->
<!-- rnb-source-begin eyJkYXRhIjoiYGBgclxuYSA8LSBUUlVFXG5gYGAifQ== -->
<pre class="r"><code>a &lt;- TRUE</code></pre>
<!-- rnb-source-end -->
<!-- rnb-chunk-end -->
<!-- rnb-text-begin -->
<p><code>a</code> is FALSE.</p>
<!-- rnb-text-end -->
<!-- rnb-chunk-begin -->
<!-- rnb-source-begin eyJkYXRhIjoiYGBgclxuYSA8LSBGQUxTRVxuYGBgIn0= -->
<pre class="r"><code>a &lt;- FALSE</code></pre>
<!-- rnb-source-end -->
<!-- rnb-chunk-end -->
<!-- rnb-text-begin -->
<p>Now <code>a</code> is FALSE.</p>
<!-- rnb-text-end -->
~~~

For the full outcome see [output.zip](https://github.com/rstudio/rmarkdown/files/975679/output.zip).


kevinushey:
How are you generating the HTML notebook here? I do not see any issue when attempting to render the document from the console; e.g. by

    rmarkdown::render("notebook.Rmd")

I suspect that what you're seeing is the result of what you get previewing this in the RStudio IDE. Unfortunately, it's hard to improve the experience here -- we want to generate a Notebook preview based on the current state of the document as seen in the RStudio IDE, but we want to avoid running arbitrary R code in larger R code chunks.

Inline code chunks are the odd one out. Because we don't collect outputs for inline code chunks, we instead evaluate those chunks based on the current state of the global environment -- this implies that the values you see in e.g. `` `r a` `` depend on whatever value `a` is currently set in the global environment, not necessarily based on what it should be as from a top-to-bottom execution of that document.

rgaiacs:
> How are you generating the HTML notebook here?

@kevinushey Thanks for the reply. I'm using the "Preview Notebook" option on the Rmd tab menu as you suspect.

screenshot_2017-05-17_09-54-03

> Unfortunately, it's hard to improve the experience here -- we want to generate a Notebook preview based on the current state of the document as seen in the RStudio IDE, but we want to avoid running arbitrary R code in larger R code chunks.
>
> Inline code chunks are the odd one out. Because we don't collect outputs for inline code chunks, we instead evaluate those chunks based on the current state of the global environment -- this implies that the values you see in e.g. `r a` depend on whatever value a is currently set in the global environment, not necessarily based on what it should be as from a top-to-bottom execution of that document.

Thanks for the great explanation. Did I miss it some where in the documentation? Could I suggest to add a "Knit to Notebook" option on the menu so users that want take advantage of the Notebook format but don't mind that arbitrary R code in larger R code chunks are run before the Notebook is generate?

jjallaire:
Currently you could do a "Run All Chunks" or a "Restart and Run All Chunks"
to emulate this behavior.

rgaiacs:
> Currently you could do a "Run All Chunks" or a "Restart and Run All Chunks"
> to emulate this behavior.

Thanks. Fee free to close this pull request.

PR #1041, opened 2017-05-10
Fixes rstudio/bookdown#394: system2(stdout = F) may not work

yihui:
I tried to debug https://github.com/rstudio/bookdown/issues/394 this afternoon, and it turns out that `system2(stdout = F)` failed straight on his Windows. I don't know why (sounds like redirecting stdout to `/dev/null` doesn't work), but it worked without `stdout = F`, so I want to fall back to the default `stdout = ""` in this case.

### DIFF
--- a/inst/NEWS
+++ b/inst/NEWS
@@ -6,6 +6,9 @@ rmarkdown 1.6 (unreleased)
 
 * Support code folding for bash, python, sql, Rcpp, and Stan chunks.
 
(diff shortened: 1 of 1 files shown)


jjallaire:
It looks like this branch has a bunch of earlier master commits from rmarkdown. Is there a way to reduce it to just what has changed?

yihui:
I was rebasing my branch (worked on the fix in my master branch by mistake). Will open a new PR in a minute.

PR #1052, opened 2017-05-16
expand markdown image paths

kevinushey:
This PR expands Markdown images paths of the form e.g.

    hello

replacing `~` with the actual home directory. Doing this allows `pandoc` to successfully convert the  resulting Markdown file.

This is implemented as a `knitr` hook as it seemed the most natural place to perform this transformation in the current `render()` infrastructure, but let me know if there's a better spot (or if we should avoid this altogether).

The underlying motivation: currently, image paths prefixed with `~/` can be previewed from the IDE; this change is required to ensure such documents can be rendered by `pandoc` as well. (We could also just turn off preview for images of that form on the IDE side instead, if preferred)

### DIFF
--- a/R/hooks.R
+++ b/R/hooks.R
@@ -1,0 +1,55 @@
+md_document_hook <- function(hook) {
+  force(hook)
+
+  # regular expression that can capture elements of an image link
+  re_link <- paste0(
+    "^\\s*",        # initial whitespace
+    "!",            # starts with an '!'
+    "\\[(.*)\\]",   # '[...]'
+    "\\((~/.*)\\)", # '(~/...)'
+    "(.*)"          # trailing content (e.g. attributes)
+  )
+
+  function(x) {
+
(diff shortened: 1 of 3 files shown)


jjallaire:
I think we should leave this alone for now b/c that syntax is inherently non-reproducible from machine to machine.

If we do want to do it at some point, the right way I think is with a pandoc JSON filter (see: http://pandoc.org/scripting.html#json-filters) which would work for all formats (essentially becoming a local extension of pandoc markdown) and wouldn't require messing with knitr hooks (which can end up clobbering/masking each other even when you do the "old hook" thing as you have done here.

kevinushey:
Ah, I wasn't aware that Pandoc allowed users to provide arbitrary filters as well -- that's pretty awesome. I agree that if we want to tackle this we should go the route of writing a Pandoc filter then.

PR #1056, opened 2017-05-24
Couple fixes to paged tables

javierluraschi:
1. Using `df_print: paged` as an option followed by `tibble::tibble(x = 1, y = list(list(list(1, 2, 3))))` causes an error instead of printing the paged table. The problem here is that we are missing to port this fix: https://github.com/tidyverse/tibble/commit/275e465a3b5f4e72f2c44d1d8c3ac6cd77f610b7
2. Columns with small names and medium types get cut off, the problem was that we were only using the column name and data, but not the column type label, to calculate the column width.

### DIFF
--- a/R/html_paged.R
+++ b/R/html_paged.R
@@ -78,9 +78,13 @@ paged_table_obj_sum <- function(x) {
     paste0(" [", dim_desc(x), "]" )
   }
 
+  obj_sum.default <- function(x) {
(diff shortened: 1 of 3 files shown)


jjallaire:
Could you also add a NEWS entry?

javierluraschi:
For sure, updated.

PR #1061, opened 2017-06-01
Fixes rstudio/bookdown#394: system2(stdout = F) may not work

yihui:
Moved from #1041.


jjallaire:
NEWS entry for this one as well would be great

PR #1062, opened 2017-06-01
Also support biblatex when emulating latexmk

yihui:
Currently only `natbib` is supported in the `latexmk` emulation mode (via `bibtex`). When the citation package is `biblatex`, we need to run `biber` on `*.bcf`.

Cc @earowang


jjallaire:
Could you also add an entry to NEWS? (fine to just commit directly)

yihui:
Sure.

PR #1065, opened 2017-06-02
Provide rmarkdown.pandoc.id_prefix as Knit option

jjallaire:


jjallaire:
@yihui Could you review and merge?

R/render.R
```diff
@@ -799,6 +801,21 @@ merge_render_context <- function(context) {
   context
 }
 
+
+id_prefix_from_args <- function(args) {
```
yihui: I'll probably not use the for loop:

```r
i <- head(which("--id-prefix" == args), 1)
if (length(i) == 0 || i == length(args)) "" else args[[i + 1]]
```

jjallaire: Okay, thanks for the review. I'll avoid the for loop next time! ;-)

Issue #729, opened 2016-06-24
When param default is NULL, error when trying to pass parameter value


wch:
For example, this document and code produce an error:

``````

---
output: html_document
params:
  x: NULL

---

```{r}
params$x
```
``````

``` R
rmarkdown::render("test.Rmd", params = list(x=1))
# Error in knit_params_get(input_lines, params) : 
#  render params not declared in YAML: x 
```

But if you change the default in the YAML header to `x: NA`, it works fine.

stevekm:
I had the same problem, but even changing to NA did not fix it. I had to fill in some default value.

yihui:
Fixed via #1115.

stevekm:
@yihui which version of knitr did this fix go into?

yihui:
@stevekm rmarkdown 1.7, released in Nov last year: https://github.com/rstudio/rmarkdown/releases/tag/v1.7

stevekm:
Thanks. Not sure if its too off-topic an issue to post here, but I am having trouble getting that version installed with conda; 

base install:
```
conda install -y -c r r-base \
r=3.3.2 \
r-base=3.3.2 \
pandoc==1.15.0.6 \
r-reshape2 \
r-knitr=1.15.1 \
r-rmarkdown \
r-ggplot2
```
Available versions:
```
root@521e956ad803:/# conda search r-rmarkdown
Loading channels: done
Name                       Version                   Build  Channel
r-rmarkdown                0.5.1                  r3.1.3_0  defaults
r-rmarkdown                0.5.1                 r3.1.3_0a  defaults
r-rmarkdown                0.7                    r3.2.0_0  defaults
r-rmarkdown                0.7                   r3.2.0_0a  defaults
r-rmarkdown                0.7                    r3.2.1_0  defaults
r-rmarkdown                0.7                   r3.2.1_0a  defaults
r-rmarkdown                0.8                    r3.2.1_0  defaults
r-rmarkdown                0.8                   r3.2.1_0a  defaults
r-rmarkdown                0.8                    r3.2.2_0  defaults
r-rmarkdown                0.8                   r3.2.2_0a  defaults
r-rmarkdown                0.8.1                  r3.2.2_0  defaults
r-rmarkdown                0.8.1                 r3.2.2_0a  defaults
r-rmarkdown                0.9.6                  r3.3.1_0  defaults
r-rmarkdown                1.2                    r3.3.1_0  defaults
[...]

PR #1115, opened 2017-08-09
Fixes issues with NULL/null/empty parameters

### LINKED ISSUE #762: Please allow us to disable param defaults
This is more of a request than an issue.

Currently, when a parameterized .Rmd is created, a default value must be coded for each param. For example, if I have the following param in the YAML...

``` yaml
other: 
      value: first
      choices: [first, second, third]
      input: select
      label: Other
      multiple: yes
```

...then I must specify a default option to go into the 'value' spot. If I leave this spot blank, it throws an error*. If I try knitting this doc with the above code, then backspacing this param widget so that no value is showing, it returns the first available value (from my choices) in the rendered doc.

It seems that RMarkdown requires a default value for every param in order to render. Is there a way I could create a parameter with no defaults (i.e., a param that passes through to the output as NULL)?

*There seems to be some NULL capability built into params in RMarkdown, but it is unpredictable (or I am using it incorrectly). For example, when I leave the 'value:' spot blank (or I make it NULL) and knit the doc with parameters, I can sometimes get an option to appear in the select widget that says "Unspecified (NULL)" and another option that says "Use a custom value." Selecting the first option _does_ pass a NULL value through to the output. Selecting the second option resets the widget to display my original choices, but throws an error during the knit. It seems like the more coder-friendly alternative would be to simply allow RMarkdown to interpret these NULL values (or blank widgets) as such, without displaying the extra options.

Thank you,

Diana

bearloga:
- Fixes how the list of “render params not declared in YAML” is
formatted
- Adds support for NULL/null/empty parameters
- Adds related tests

Additionally, I'm _pretty sure_ this fixes #762.

### DIFF
--- a/R/params.R
+++ b/R/params.R
@@ -6,11 +6,7 @@ knit_params_get <- function(input_lines, params) {
   if (packageVersion('yaml') < '2.1.14') knit_params <- mark_utf8(knit_params)
   default_params <- list()
   for (param in knit_params) {
-    if (is.null(param$value)) {
-      default_params[param$name] <- list(NULL)
-    } else {
-      default_params[[param$name]] <- param$value
-    }
+    default_params[param$name] <- list(param$value)
   }
 
   # validate params passed to render
--- a/R/params.R
+++ b/R/params.R
@@ -6,7 +6,11 @@ knit_params_get <- function(input_lines, params) {
   if (packageVersion('yaml') < '2.1.14') knit_params <- mark_utf8(knit_params)
   default_params <- list()
   for (param in knit_params) {
-    default_params[[param$name]] <- param$value
+    if (is.null(param$value)) {
+      default_params[param$name] <- list(NULL)
+    } else {
+      default_params[[param$name]] <- param$value
+    }
   }
 
   # validate params passed to render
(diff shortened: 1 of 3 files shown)


R/params.R
```diff
@@ -6,7 +6,11 @@ knit_params_get <- function(input_lines, params) {
   if (packageVersion('yaml') < '2.1.14') knit_params <- mark_utf8(knit_params)
   default_params <- list()
   for (param in knit_params) {
-    default_params[[param$name]] <- param$value
+    if (is.null(param$value)) {
+      default_params[param$name] <- list(NULL)
+    } else {
+      default_params[[param$name]] <- param$value
+    }
```
yihui: You can use `default_params[param$name] <- list(param$value)`

bearloga: Thank you!

jjallaire:
@aronatkins You are (for better or worse) the expert on the implications of NULL parameter values. What's your take here? Are there other downstream implications that need accounting for?

aronatkins:
Also fixes #729

bearloga:
Thank you everyone for a positive experience!

PR #1117, opened 2017-08-11
Add standard include block to html_vignette template

hadley:
Let me know if you think this is too opinionated.


jjallaire:
I'm good with this. @yihui ?

yihui:
I don't have an opinion.

PR #1126, opened 2017-08-23
Better error messages when unable to prerender a document

### LINKED ISSUE #1125: With shiny_prerendered, unhelpful error message when directory is not writable
In deployment scenarios, it's not uncommon for users to have the wrong owner or permissions for their directories. With a `shiny_prerendered` doc, when the target directory isn't writable, it gives this unhelpful message:

```
> rmarkdown::run("test.Rmd", render_args = list(output_dir="/"))
Error in file(con, "r") : cannot open the connection
In addition: Warning messages:
1: In normalizePath(path, winslash = winslash, mustWork = mustWork) :
  path[1]="//test.html": No such file or directory
2: In file(con, "r") :
  cannot open file '//test.html': No such file or directory
```

rstudio/learnr#112 is an example of this. The error message didn't help at all for diagnosing the problem.

I think the culprit is that when the directory isn't writable, it assumes that the prerendering doesn't need to happen, that the prerendered file already exists:
  https://github.com/rstudio/rmarkdown/blob/6def14c/R/shiny_prerendered.R#L79-L81

When it gets to this line, it expects the prerendered file to exist, but it doesn't, raising a warning. Later on it tries to read the file and errors out.
  https://github.com/rstudio/rmarkdown/blob/6def14c/R/shiny_prerendered.R#L131

I think the solution is to check if the file exists above line 131 and error if it's not there. Does that seem reasonable? I just want to make sure that I'm not overlooking some use case here.

wch:
This fixes #1125. I added errors in two separate places because the unwritable directory case will probably be a common one, and deserves its own error message.

### DIFF
--- a/R/shiny_prerendered.R
+++ b/R/shiny_prerendered.R
@@ -77,6 +77,9 @@ shiny_prerendered_html <- function(input_rmd, encoding, render_args) {
   prerender_option <- tolower(Sys.getenv("RMARKDOWN_RUN_PRERENDER", "1"))
 
   if (file.access(output_dir, 2) != 0) {
+    if (!file.exists(rendered_html))
+      stop("Unable to write prerendered HTML file to ", rendered_html)
+
     prerender <- FALSE
   }
   else if (identical(prerender_option, "0")) {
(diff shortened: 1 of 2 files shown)


jjallaire:
Thanks!

PR #1130, opened 2017-08-29
More resilient icons dependencies detection

sgrubsmyon:
I wanted to use fa-icons in a navbar, but rmarkdown would not inject their html dependencies. I looked into the code and found that only if a <span> tag is used for the icons, it works. But I have used the function `shiny::icon()` that returns an <i> tag.

I decided to change the code. One can now use both <span> and <i> and also it does not matter whether one uses double or single quote. Even more resilience is added by also ignoring some whitespace.

### DIFF
--- a/R/html_dependencies.R
+++ b/R/html_dependencies.R
@@ -107,7 +107,7 @@ navbar_icon_dependencies <- function(navbar) {
   source <- readLines(navbar)
 
   # find icon references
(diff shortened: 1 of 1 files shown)


jjallaire:
@yihui This seems like a good fix to me. Could you take a look and merge if you don't see issues?

Issue #1110, opened 2017-08-04
theme not available for html_vignette


ThierryO:
Below is a minimal example. The `html_document` compiles. The `html_vignette` returns an error: `formal argument "theme" matched by multiple actual arguments`. 

    ---
    output: 
      html_vignette:
        theme: "readable"
      html_document:
        theme: "readable"
    ---
    
    ## R Markdown

jjallaire:
That's correct, themes are not supported for html_vignettes (that's part of
the notion of the html_vignette templates is that by having a more
lightweight theme smaller html files are generated).

ThierryO:
Then I would suggest to update the help file of `html_vignette`. It currently reads `... | Additional arguments passed to html_document`. And maybe ignore the `theme` (with a warning).

yihui:
@ThierryO That is a good suggestion. We will do it, unless you want to submit a pull request :)

PR #1136, opened 2017-09-06
ignore theme, highlight and fig_retina in html_vignette()

ThierryO:
The arguments are silently overwritten. This behaviour is documented in the help file. 

I've tried to set a warning but that wasn't displayed. Hence I removed the warning again.

solved issue #1110

### DIFF
--- a/R/html_vignette.R
+++ b/R/html_vignette.R
@@ -22,7 +22,7 @@
 #' @inheritParams html_document
 #' @param ... Additional arguments passed to \code{\link{html_document}}. Please
 #'   note that \code{theme}, \code{fig_retina} and \code{highlight} are hard
-#'   coded. They will be silently overwriten when set.
+#'   coded. Setting them will yield an error.
 #' @param readme Use this vignette as the package README.md file (i.e. render
 #'   it as README.md to the package root). Note that if there are image files
 #'   within your vignette you should be sure to add README_files to .Rbuildignore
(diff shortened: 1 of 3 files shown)


R/html_vignette.R
```diff
@@ -48,20 +52,21 @@ html_vignette <- function(fig_width = 3,
...
     keep_md = keep_md,
-    base_format = html_document(fig_width = fig_width,
-                                fig_height = fig_height,
-                                dev = dev,
-                                fig_retina = NULL,
-                                css = css,
-                                theme = NULL,
-                                highlight = "pygments",
-                                ...)
+    base_format = do.call(html_document, dots)
```
yihui: Since you are not going to issue warnings, the rewrite here is no longer meaningful (I mean `do.call(FUN, list(...))`). Could you restore the changes in the function?

ThierryO: The old version of the function cause an error when `theme` is supplied. The version in this pull request just ignores `theme` set by the user and compiles.

Do you prefer the error?

yihui: Yes. I think it will be confusing if the `theme` option is silently ignored.

R/html_vignette.R
```diff
@@ -18,7 +20,9 @@
 #' documentation} for additional details on using the \code{html_vignette} format.
 #'
 #' @inheritParams html_document
-#' @param ... Additional arguments passed to \code{\link{html_document}}
+#' @param ... Additional arguments passed to \code{\link{html_document}}. Please
+#'   note that \code{theme}, \code{fig_retina} and \code{highlight} are hard
+#'   coded. Setting them will yield an error.
```
yihui: Typo `them` -> `theme`.

Actually setting any of these hard-coded arguments will yield errors (it is not only `theme`).

R/html_vignette.R
```diff
@@ -48,7 +52,6 @@ html_vignette <- function(fig_width = 3,
                         quiet = TRUE)
     }
   }
-
```
yihui: Could you restore this blank line?

PR #1165, opened 2017-10-19
Fix #1162: use parent.frame() of run() instead of new.env() inside rmarkdown_shiny_server() to evaluate code chunks

### LINKED ISSUE #1162: Error with data.frame syntax with runtime:shiny and renderPlot
This is originally from #1160 and rstudio/shiny#1872, but I'm filing here with a more detailed example.

There's some sort of interaction between data.table, rmarkdown, and shiny, where the data.table syntax doesn't work correctly in a `renderPlot()`. Oddly, it does work if you use `eval(quote(<expr>))`, but not `evalq(<expr>)`.

````
---
output: html_document
runtime: shiny
---

```{r}
library(data.table)
mydata <- data.table(A=c("a", "a", "b", "c", "c"), val=c(1:5))

renderPlot({
  # Works - see console for output
  try(
    print(eval(quote( mydata[A %in% c("a", "c")] )))
  )
  
  # Error - see console for message
  try(
    print(evalq( mydata[A %in% c("a", "c")] ))
  )
  
  # Error - see console for message
  try(
    print( mydata[A %in% c("a", "c")] )
  )

  plot(1:3, 1:3) # Just an arbitrary plot
})


# All forms work fine in renderPrint
renderPrint({
  print(eval(quote( mydata[A %in% c("a", "c")] ))) # Works
  print(evalq( mydata[A %in% c("a", "c")] ))       # Works
  print( mydata[A %in% c("a", "c")] )              # Works
})

```
````

### PR DESCRIPTION
**Pull request by yihui**
This PR fixes #1162. BTW, I believe this should also be the correct fix to #619 (I have tested it), instead of the PR #624, which introduced `new.env()`.

### DIFF
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -138,6 +138,8 @@ run <- function(file = "index.Rmd", dir = dirname(file), default_file = NULL,
     else
       render_args$encoding
 
+  if (is.null(render_args$envir)) render_args$envir <- parent.frame()
+
   # determine the runtime of the target file
   target_file <- ifelse(!is.null(file), file, default_file)
   if (!is.null(target_file))
@@ -273,8 +275,7 @@ rmarkdown_shiny_server <- function(dir, file, encoding, auto_reload, render_args
                                output_dir = dirname(output_dest),
                                output_options = output_opts,
                                intermediates_dir = dirname(output_dest),
-                               runtime = "shiny",
-                               envir = new.env()),
+                               runtime = "shiny"),
                           render_args)
       result_path <- shiny::maskReactiveContext(do.call(render, args))
 
--- a/R/shiny_prerendered.R
+++ b/R/shiny_prerendered.R
@@ -123,10 +123,7 @@ shiny_prerendered_html <- function(input_rmd, encoding, render_args) {
   if (prerender) {
 
     # execute the render
-    args <- merge_lists(list(input = input_rmd,
-                             encoding = encoding,
-                             envir = new.env(parent = globalenv())),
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by wch** [comment]
I'm not sure this is fully correct, but maybe the fix should be like the one for #1124?

For example:

````
---
output: html_document
runtime: shiny
---

```{r}
args
```


```{r}
pryr::parenvs()
pryr::rls()
```
````

Here's what it looks like:

image

image

**Comment by yihui** [comment]
Oh #1124. That was the issue I was looking for. I vaguely remember someone filed a similar issue.

I feel the fix for #1124, `new.env(parent = globalenv())`, is not fully correct, though. We should always use `parent.frame()` instead of `globalenv()` or its child environments when evaluating code chunks. In most use cases, they are pretty much equivalent (i.e. `parent.frame()` is `globalenv()`), but in rare cases, the latter can blow up, e.g.

```r
f1 = function() {
  eval(expression(x), envir = parent.frame())
}

f2 = function() {
  eval(expression(x), envir = new.env(parent = globalenv()))
}

with(list(x = 1), f1())  # works
with(list(x = 1), f2())  # error

f3 = function() {
  x = 1
  f1()
}
f4 = function() {
  x = 1
  f2()
}
f3()  # works
f4()  # error
```

You can imagine `f2()`/`f4()` as `run()`.

BTW, my fix in this PR actually gives users the freedom to specify the environment if necessary instead of hardcoding it to `new.env()`.

wch:
That sounds reasonable to me.

PR #1177, opened 2017-10-26
Do not use output_options in default_site()

### LINKED ISSUE #793: Error for multiples outputs in render_site
I create my webpages with rmarkdown, function `render_site`. But I have problems when I use more outputs in yaml. For example:

**In _site.yml file**

```
name: ""
output_dir: "."
output:
  html_document: default
  pdf_document: default
navbar:
  left:
    - text: Home
      href: index.html
```

**In index.Rmd file**

```
% My title
% EERJ

## R Markdown

This is an R Markdown document. Markdown is a simple formatting syntax
for authoring HTML, PDF, and MS Word documents. For more details on
using R Markdown see <http://rmarkdown.rstudio.com>.
```

Thus, the function `rmarkdown::render_site()` generates the error below 

``` r
Error in (function (toc = FALSE, toc_depth = 2, number_sections = FALSE,  : 
  unused arguments (lib_dir = "site_libs", self_contained = FALSE)
```

I suppose the problem is related to https://github.com/rstudio/rmarkdown/blob/master/R/render_site.R#L195, because the list in `output_options` is unused in different outputs of`html_document`.

yihui:
Fixes #793.

Because these options (`lib_dir` and `self_contained`) are not supported by all output formats, e.g. `pdf_document`. Instead, temporarily patch the `html_document` format options in `_site.yml` during `render()`.

### DIFF
--- a/R/render_site.R
+++ b/R/render_site.R
@@ -173,8 +173,9 @@ default_site <- function(input, encoding = getOption("encoding"), ...) {
                      quiet,
                      encoding, ...) {
 
-    site_yml <- patch_html_document_options(config, encoding)
-    on.exit(file.rename(site_yml, '_site.yml'), add = TRUE)
+    site_yml  <- file.path(input, '_site.yml')
+    site_yml2 <- patch_html_document_options(config, encoding, site_yml)
+    on.exit(file.rename(site_yml2, site_yml), add = TRUE)
 
     # track outputs
     outputs <- c()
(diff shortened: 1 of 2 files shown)


jjallaire:
This fix effectively forces render_site() to always render an html_document based version of the site (since the patching introduces an `html_document` format), which in turn make render_site incompatible with other HTML based output formats (because the html_document then conflicts with them).

I would say that the original bug that this fixes is actually not in the design of R Markdown Websites (rendering HTML and PDF from a website) whereas supporting formats other than html_document is definitely in the design!

I will see if I can sort out a fix.

jjallaire:
I honestly feel like the best answer is to revert this fix. We could only do the patching when html_document is present within the site's _config.yml. This however might be problematic if sites don't specify an output format (i.e. rely on the default of html_document). Another problem with this solution is that all other HTML based formats that want to be published as sites now need to instruct their users to explicitly set lib_dir and self_contained.

We've effectively broken sites for non html_document for the purpose of supporting sites which want to have alternate non-html renderings (something we never said we could do).

PR #1178, opened 2017-10-26
Upgrade highlight.js from v1.1 to v9.12.0

### LINKED ISSUE #988: highlight.js eating certain numbers
Reported from http://rmarkdown.rstudio.com/authoring_knitr_engines.html#comment-3191518865 Reproducible example:

    ---
    title: "Test"
    output: 
      html_document: 
        theme: null
    ---
    
    ```{bash, eval = FALSE}
    R-3.3.2 RHOME
    ```

Output (`.2` was eaten for some reason):

image

The HTML source is actually correct:

```html
<pre class="bash"><code>R-3.3.2 RHOME</code></pre>
```

It works fine with syntax highlighting disabled:

image

Or use Pandoc's syntax highlighter:

image

I think we are lagging too far behind the current version of highlight.js, and our highlight.js v1.1 is starting to bite us now (also see #907).

yihui:
Fixes #988.

I picked these languages from https://highlightjs.org/download/:

```js
var langs = [
  'awk', 'bash', 'coffeescript', 'cpp', 'css', 'diff', 'fortran', 'go',
  'ini', 'java', 'javascript', 'json', 'julia', 'makefile', 'markdown',
  'perl', 'python', 'r', 'ruby', 'sql', 'stan', 'tex', 'xml', 'yaml'
];
document.querySelectorAll('input').forEach(function(el) {
  el.checked = langs.indexOf(el.name.replace(/\.js$/, '')) > 0;
});
```

Let me know if you want to add more or remove any.

I have mixed and matched the highlight themes `default.css` and `textmate.css`, and the output should look similar to before.

I don't truly understand the issue in #89, but [`hljs.initHighlighting` has been defunct since v3.0](https://github.com/isagalaev/highlight.js/blob/master/CHANGES.md):

> The function `initHighlighting` that was used to initialize highlighting instead of initHighlightingOnLoad a long time ago no longer works. If you by chance still use it — replace it with the new one.


jjallaire:
Actually, we have a special version of highlight.js that Joe patched to do a better job at highlighting constants and R operators. We have a script somewhere that can be used to create this version, I believe @kevinushey may recall the details as he last updated it to add more languages.

jjallaire:
Our fork is here: https://github.com/rstudio/highlight.js/tree/r  (it's the `r` branch of our fork).

The `tools/build.sh` script will build you a new highlight.js with a different set of languages.

If you were doing the update to get features from the more recent releases of highlight.js then I don't think that's practical at this point as we are thousands of commits behind the upstream.

yihui:
That sounds very sad. Was there a reason why we could not get our changes merged into the official highlight.js in the beginning?

jjallaire:
I believe that @jcheng5 tried to get both of our changes merged (one had to do with finer-grained highlighting and the other had to with highlight.js not automatically guessing languages, which caused over-highlighting of code blocks in the Rmd). IIRC the author of highlight.js did not agree with either of these changes and wouldn't take our PR.

We could consider accepting less granular highlighting of R symbols as a tradeoff for syncing to the newer version. If highlight.js has backed off on the notion that they want to automatically highlight every `<pre><code>` then this is more legit, but if they insist on this I don't know if we'll like the results in Rmd output.

yihui:
When I use highlight.js, I always turn off the automatic guessing of languages, because it rarely makes much sense in the case of R Markdown output (code blocks have already got proper classes). To turn off guessing, you can set `hljs.configure({languages: []})`.

I'll take a closer look at the granularity issue. If Joe's changes look simple enough, I can try to make another PR to highlight.js. I guess it is not really big deal.

jjallaire:
Okay, sounds good!

yihui:
I compared our `src/languages/r.js` with the current official version in highlight.js:

- https://github.com/rstudio/highlight.js/blob/0710243a8ea53f2b2d34cebd58b7fa82b7760863/src/languages/r.js
- https://github.com/isagalaev/highlight.js/blob/master/src/languages/r.js

In terms of granularity, our version highlights these extra tokens:

- The special arguments `..1`, `..2`, etc. They should be rare (I have never seen them for a single time in any R code I have read).

- Parentheses `()` `[]` `{}`, and operators like `+`, `-`, `%%`. highlight.js no longer support them: http://highlightjs.readthedocs.io/en/latest/css-classes-reference.html

I think the visual differences between the two versions are nearly unnoticeable:

With our highlighter:

image

With highlight.js's highlighter:

image

The textmate theme:

image

image

So I'd suggest retire our fork of highlight.js and keep up with the official release (which has also included the official stan support).

kevinushey:
I'm personally a fan of dropping our fork and moving to the official release, but I will really miss the operator highlighting. At least for me, without separate highlighting of operators, the identifiers on lines with lots of code tend to blur together and I have to squint to separate them.

FWIW this issue suggests that it's at least possible, even if it might not be recommended:

https://github.com/isagalaev/highlight.js/issues/1268

yihui:
Okay, let me try to add support for operators and submit a PR to highlight.js.

yihui:
I thumbed through all other language definitions, and I didn't find any of them support syntax highlighting operators, which makes me feel not optimistic about getting this supported for R, but I just submitted a PR any way: https://github.com/isagalaev/highlight.js/pull/1629.

On the other hand, operators do not seem to be distinct in either the default or textmate theme (see screenshots above, e.g. `+`), which means even if they can be syntax highlighted, it is still difficult to notice the differences.

jjallaire:
@yihui I'm okay with switching to the newer highlight.js in spite of the lack of operator highlighting.

kevinushey:
Perhaps we should take this PR as-is, and see whether we're able to get upstream support for operator highlighting in a future release of `highlight.js`?

yihui:
Okay, let's work on operators in the future.

PR #1199, opened 2017-11-13
Attach metadata to render()'s return value via attach_metadata()

yihui:
1. The metadata is to be specified in the field `rmd_output_metadata` in YAML. The name `rmd_results` (in the Google doc) sounds too vague to me, but I'm open to other suggestions.

2. Users call `output_metadata$set()` to update arbitrary elements. The API is the same as `opts_chunk` in **knitr**, e.g., `output_metadata$set(subject = "Hello World")`.

    - If users feel `output_metadata` is too long to type, this object is mutable in the sense that you are free to create copies and the `$set()` method still works in these copies, e.g., `m <- output_metadata; m$set(subject = ...)` and `subject` will be modified in `output_metadata`.

3. The attribute name is `rmd_output_metadata`. Again, I'm open to suggestions. Naming is hard.

4. The documentation is intentionally brief, since I cannot think of another use case other than sending emails from RSC. The implementation is general and not tied to the case of sending emails. You can document it in RSC with specific examples.


yihui:
@trestletech @aronatkins This PR is ready for review now.

trestletech:
Code change LGTM. I can get behind `output_metadata` but `rmd_output_metadata` feels like a mouthful to me. I know that you're worried about Pandoc collisions so the more unique, the better. I'd vote for `rmd_metadata` or something a little simpler, but I'm not strongly opinionated.

yihui:
Yes, it is mouthful, but I could think of a better name (I spent more time on coming up with a proper name than writing the code). I feel any shorter names will almost surely be ambiguous, e.g., `rmd_metadata` sounds too general. Remember that we have already got `rmarkdown::metadata` in the package, and we need to differentiate the new type of metadata with the existing general `metadata` (i.e., the full YAML frontmatter).

yihui:
Merged, but the names of the function and YAML field are still open to changes if necessary. Just let me know before the next rmarkdown release (in mid or late January).

aronatkins:
I have successfully been able to extract output metadata from the attributes of the `rmarkdown::render` result.

Given code in an Rmd like:
````
```{r echo = FALSE}
rmarkdown::output_metadata$set(subject = "hello, world")
```
````

we can:
```
result <- rmarkdown::render(.....)
atts <- attributes(result)
jsonlite::toJSON(att$rmd_output_metadata, auto_unbox = TRUE)
## => {"subject":"hello, world"}
```

Note: this snippet is missing existence checking, etc.

yihui:
Sounds good. FYI, you can obtain a single attribute using `attr(result, "rmd_output_metadata", exact = TRUE)`.

PR #1222, opened 2017-12-15
Build PDF through tinytex::latexmk()

yihui:
Basically the function `latexmk()` in this package has been moved to the **tinytex** package (and improved there).

The other major change is that with `keep_tex = TRUE`, we only need one pass of Pandoc now (.Rmd -> .tex, and .tex is compiled to PDF through `latexmk()` instead of Pandoc). Previously two passes were needed (one for .tex and one for .pdf), which was inefficient.

Also note that **tinytex** does not require the LaTeX distribution TinyTeX or TeX Live (e.g., MikTeX is fine).


DESCRIPTION
```diff
@@ -74,6 +74,7 @@ Imports:
     methods,
     stringr (>= 1.2.0)
 Suggests:
+    tinytex,
```
jjallaire: What about importing it and making it the default behavior?

yihui: Yeah I was about to ask you. My only concern is that tinytex is a new package on CRAN, and I'm not sure if users with R versions lower than 3.3 are able to install it since CRAN won't provide binaries. Well, they surely can, but they just have to `install.packages('tinytex', type = 'source')` (it is a pure R package and does not require compilation).

I was wondering if we need to wait a little longer before fully routing to tinytex.

jjallaire:
Yeah, that's a good point. Maybe wait a while. Note that when we do installs of rmarkdown in RStudio we can work around this by explicitly doing a source install of tinytex (in fact, we can do this right away even if tinytex is a Suggests).

@jmcphers

PR #1231, opened 2018-01-02
Support Powerpoint Presentation with Pandoc >= 2.0.5

yihui:
Known Pandoc issues (not sure when they will be fixed):

- Links on images don't work `[text](link)`.

- A Markdown table with an empty column name in the first column doesn't work.

Known knitr issues (will be fixed relatively soon):

- Cannot specify size for R plots: https://github.com/yihui/knitr/issues/1478


jjallaire:
👍

PR #1233, opened 2018-01-04
Add `library()` to html_vignette template

colearendt:
It seems (I haven't yet found documentation) that best practices suggest adding `library(MY_PACKAGE_NAME)` to the top of package vignettes (`devtools::check` fails without it).  In concert with discussion at https://github.com/r-lib/usethis/issues/190 , I suggest something like the following:

````
```{r setup, include = FALSE}
knitr::opts_chunk$set(
  collapse = TRUE,
  comment = "#>"
)
#library(YOUR_PACKAGE_HERE)
```
````

I added as a comment to maintain a valid Rmd.  I could not decide if this warranted an issue of its own.

### DIFF
--- a/inst/rmarkdown/templates/html_vignette/skeleton/skeleton.Rmd
+++ b/inst/rmarkdown/templates/html_vignette/skeleton/skeleton.Rmd
@@ -14,6 +14,7 @@ knitr::opts_chunk$set(
   collapse = TRUE,
   comment = "#>"
 )
+#library(YOUR_PACKAGE_HERE)
 ```
 
 Vignettes are long form documentation commonly included in packages. Because they are part of the distribution of the package, they need to be as compact as possible. The `html_vignette` output type provides a custom style sheet (and tweaks some options) to ensure that the resulting html is as small as possible. The `html_vignette` format:

### CODE REVIEW

**Comment by yihui** [comment]
> `devtools::check` fails without it

I don't think this is true. If it is, it must be a bug of `devtools::check`. 

The point is, if you want to use any objects in an add-on package, you must either `library(pkg)`, or `pkg::object`. In other words, `library(pkg)` is not the only way to go, so I don't want to make this suggestion in the default template (otherwise users may gain a false belief that all packages must be _attached_ before any objects can be used).

**Comment by colearendt** [comment]
That makes sense.  I guess I am partly confused by _when_ the package being built is attached to the vignette build.

For instance, `devtools::build_vignettes()` works fine without `library(pkg)` or `pkg::blah` (it seems to attach/`load_all` the package before building the vignettes.  That does not seem to be the case for `devtools::check`, `pkgdown::build_site`, etc.  I.e. you need to have the package installed and then reference via `library(pkg)` or `pkg::`.  Intuitively, I would expect a vignette to already have the package attached, since it is a part of the package development.  I.e. I expect a vignette to be built within the package's namespace, just like other functions within the package operate.

> false belief that all packages must be attached before any objects can be used

I guess I am not thinking so much about "all packages" but about the "package being developed."  Looking at all of the `dplyr` and `ggplot2` vignettes, for instance, `dplyr` always includes `library(dplyr)` and `ggplot2` always includes a `library(ggplot2)` in the vignettes. 

I would also think most who are building a package vignette are likely comfortable with the `library()` and `pkg::` conventions, and probably prefer to have the package attached when designing a vignette, which is specifically focused on teaching about the package in question.

Just trying to help others avoid the confusion I had about whether or not I _should_ include a `library()` call in my vignette.  Perhaps documentation is the easier solution.  Happy to have your suggestions about how best to avoid confusion on vignette development!

**Comment by yihui** [comment]
I rarely use **devtools** myself for package development, and I don't always `library()` the package being developed in the vignettes, e.g., https://github.com/yihui/knitr/blame/master/vignettes/knitr-intro.Rmd I don't have problems with `R CMD check`.

> For instance, `devtools::build_vignettes()` works fine without `library(pkg)` or `pkg::blah` (it seems to attach/`load_all` the package before building the vignettes.

If your guess is true (I don't believe so), I'd think that would be a _very bad_ idea of **devtools**. It is likely that you called `load_all()` by yourself before you build vignettes, and that would be the source of confusion -- you should clearly know that `load_all()` is for your convenience only and not what you would use R or R packages routinely: it exposes all objects (exported or not exported) in your workspace, so you don't need `pkg::` or `pkg:::` to access these objects _while developing your package_. It is indeed very convenient. However, when you are back to the real world (including package vignettes and tests), you should follow the normal way to use objects in the package, which is either `library(pkg)` or `pkg::`.

**Comment by colearendt** [comment]
Ahh yes.  You are correct - I was operating under a (false) assumption that `build_vignettes` was operating in a fresh R environment like `knit` does.  Thanks for your time on this - I am learning a lot and correcting a slew of misconceptions!

One last tidbit is that testing usually (at least with `testthat`) has the package attached already by default.  Now that I think through it a bit more clearly, though, it does make sense that `testthat` would operate in the package's namespace (e.g. testing internal functions, perhaps? although I am unsure of [whether that is advisable](http://r-pkgs.had.co.nz/tests.html)) while a vignette would operate "outside" to show what the external interface looks like.

And actually, `devtools::test()` loads via `devtools::load_all()`
```
...
    message("Loading ", pkg$package)
    ns_env <- load_all(pkg, quiet = TRUE)$env
...
```

Hopefully I am the only one who landed with this confusion after [reading about vignettes](http://r-pkgs.had.co.nz/vignettes.html).  I will keep an eye out for the similarly confused and see if there are other ways to assist if it crops up.  It seems to me that `build & install` is a better general workflow for interactive package development than `load_all`, since it establishes the line between namespaces better (especially for vignette building, knitting, etc.).

yihui:
I can certainly understand that. The interactive development is very convenient, but eventually you need to check everything in a fresh R session, because that is how your users will use your package, and you cannot make assumptions on what is available and what is not.

Yes, for testing, it is a good idea not to have to use the triple colon `:::` to access internal objects. Tests are supposed to be run by developers only. Basically there are an "inner" world and an "outer" world. Tests live in the inner world (for developers). Vignettes should live in the outer world (for end-users).

Thanks for the discussion!

PR #1239, opened 2018-01-09
Update params.R to support passwordInput

skadauke:
It's useful to be able to enter a password as a knitr parameter. This allows passing login credentials, e.g. to a database, that may be required to generate a reproducible report.


skadauke:
Thanks Yihui!
Stephan

Issue #1235, opened 2018-01-04
TOC does not indent Level 3 Headings


carolynwclayton:
The table of contents in an HTML Rmarkdown document indents Level 2 headings correctly, but does not indent Level 3 headings further than Level 2 headings.

Here is the code used to generate the issue:
```
---
title: "Example"
output:
  html_document:
    toc: true
    toc_float:
      collapsed: false
    toc_depth: 3
---

# Title
## Section 1
### Section 1.1
## Section 2
```

Here is the resulting table of contents:
![screen shot 2018-01-04 at 3 17 48 pm](https://user-images.githubusercontent.com/10274351/34586826-e0823ade-f162-11e7-9526-9ee8d5432e4a.png)

Here is my session info:
![screen shot 2018-01-04 at 3 20 36 pm](https://user-images.githubusercontent.com/10274351/34586835-e690cec2-f162-11e7-8cc0-8d712f2ff4a2.png)

chauhan-nitin:
Indent with each heading followed by the heading size required.

yihui:
Should be fixed now. Thanks for the report!

PR #1243, opened 2018-01-14
Indent Level Headings in TOC (#1235)

RLesur:
Here is a proposition to issue #1235.

image

Romain Lesur


yihui:
Thanks a lot! I was wonder if there was a more parsimonious way to write the CSS that makes use of the natural parent-child relationships of different levels of headers, instead of relying on the attribute `[data-tag="X"]`.

RLesur:
I modified the PR. It now uses parent-child relationships, but the code is not very parsimonious :roll_eyes: 
Sorry, I'm not a CSS expert!

mattaws:
is this still open?  if it is, is there a workaround to indenting 3rd/4th level?

yihui:
@mattaws Please try the development version here. Thanks!

PR #1284, opened 2018-03-10
Add `compact-title` variable to default LaTeX templates.

crsh:
Hi,

I have recently experienced problems with the default LaTeX template. When I tried to use the `apa6` document class the LaTeX package `titling`, which is included to change title format to be more compact, caused the title page to be rendered incorrectly. I was unable to resolve this issue by further customizing the LaTeX preamble.

This pull request adds a `pandoc` variable named `compact-title` to the default LaTeX templates that controls whether the `titling` package is used. The variable defaults to `TRUE` when the variable is not specified in the YAML front matter.

Additionally, I have slightly amended the regex used to search the input file for parameter set in the YAML front matter to work even if users add white space in between the parameter names and the colon. This prevents the preprocessor from overwriting user settings when white space is used to improve the readability of the YAML front matter like so:

~~~YAML
---
title     : "Test document"
author    : "Author"
date      : "10 3 2018"
geometry  : margin=2in
output    : pdf_document
---
~~~


crsh:
Hi @yihui, thanks for the feedback and sorry about conflating two changes here; I'll keep things separate with my next pull request.

I might just add that the reason I'm trying to amend the default LaTeX template rather than creating a complete custom template is that it would greatly improve maintainability of my package. You have kindly developed the machinery to make the default template work with different versions of `pandoc` and to support newly added features with every version. I have found it hard to keep up with these developments (I'm not a full-time developer) and it seems rather inefficient to parallely reimplement these features in my package.

crsh:
Hi, is there anything I can do to facilitate a decision on this pull request? My further package development depends on whether the change to the template is accepted.

PR #1287, opened 2018-03-15
Merge roxygen @details sections into top 3+ paragraphs

rich-iannone:
Roxygen documentation text from the `@details` section is moved toward the top (becoming the 3rd paragraph). Some line breaks were also added to improve readability within the functions themselves.

### DIFF
--- a/R/html_resources.R
+++ b/R/html_resources.R
@@ -3,20 +3,17 @@
 #' Given an R Markdown document or HTML file, attempt to determine the set of
 #' additional files needed in order to render and display the document.
 #'
-#' @param input_file path to the R Markdown document or HTML file to process
-#' @param encoding the encoding of the document
+#' This routine applies heuristics in order to scan a document for
+#' possible resource references.
 #'
-#' @details This routine applies heuristics in order to scan a document for
-#'   possible resource references.
+#' In R Markdown documents, it looks for references to files implicitly
+#' referenced in Markdown (e.g. \code{![alt](img.png)}), in the document's
+#' YAML header, in raw HTML chunks, and as quoted strings in R code chunks
+#' (e.g. \code{read.csv("data.csv")}).
 #'
-#'   In R Markdown documents, it looks for references to files implicitly
-#'   referenced in Markdown (e.g. \code{![alt](img.png)}), in the document's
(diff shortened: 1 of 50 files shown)


R/html_resources.R
```diff
@@ -70,17 +69,23 @@ find_external_resources <- function(input_file,
     path = character(0),
     explicit = logical(0),
     web = logical(0))
+
   input_dir <- dirname(normalize_path(input_file))
 
   # discover a single resource--tests a string to see if it corresponds to a
   # resource on disk; if so, adds it to the list of known resources and returns
   # TRUE
-  discover_single_resource <- function(path, explicit, web) {
```
yihui: Personally I'm fine with the original style, since the list of arguments were short enough. The new style leaves a large amount of white spaces on the left. Usually I don't prefer wasting spaces like that if possible. My coding style is often very dense, so I can read most code without moving my eyes over a long distance.

Again, no strong opinion. I never really argue about coding styles -- use whatever style that makes you feel most comfortable if you are the main person to maintain the code.

R/html_resources.R
```diff
@@ -189,8 +197,8 @@ discover_rmd_resources <- function(rmd_file, encoding,
   rmd_content <- read_lines_utf8(rmd_file, encoding)
   writeLines(rmd_content, md_file, useBytes = TRUE)
 
-  # create a vector of temporary files; anything in here will be cleaned up on
-  # exit
+  # create a vector of temporary files; anything in here
+  # will be cleaned up on exit
```
yihui: Looks like you use 55 as the width? Mine is 80 (the default in RStudio).

R/html_resources.R
rich-iannone: I might actually revert these types of line breaks for functions with low amounts of args in internal functions. I agree it's probably counter productive.

R/html_resources.R
rich-iannone: I mostly try to make the comments and examples a certain (shorter) width and try to keep everything else within 80.

rich-iannone:
@yihui after code reviews are approved, do I squash and merge?

yihui:
Yes. When I approve a PR, it means I'm good with the changes. You may merge any time you like.

Issue #1289, opened 2018-03-20
Problems with absolute paths in plots under LaTeX (Windows)


NMarkgraf:
Hi Yihui,

well, me again, I got into trouble with the handling of the rmardown generated plots in LaTeX.

As you might now, we are using Rmarkdown for a little Project with many writes on different operating systems. So our scripts have to run under Windows and *nix systems. But there is a little problem with the handling of plots.

I've made a short example, with is include in this bug report, but I will explain it  first:

Our directory structure is like this:

```
+-content/
!   +-child.Rmd
+-images/
!  +-rmarkdowrn.png
+-build/
+-MainTestDocument.Rmd
+-Makefile.R
```

We use the Makefile.R to build MainTestDocument.Rmd as MTDocument.pdf (as a beamer presentation) in the build/ directory.

We got no problem with including images, like rmarkdown in the images/ directory, because they will include with a relative path. 

But all plots will have a absolute path, with causes Windows system to get in trouble because of the absolute path part to the main document directory. 

We got something like

```tex
\includegraphics{C:\path\to\main\document\dirctory/path/to/generated/images.png}
```

On *nix systems it'll be:
```tex
\includegraphics{/path/to/main/document/dirctory/path/to/generated/images.png}
```

which is also not nice, but will do it. But it would be much nice to have only relative paths or correct paths on windows systems.

Thank in advance,
Norman

BTW: If I use the kntir-button in RStudio it works fine.

[yihui-test.zip](https://github.com/rstudio/rmarkdown/files/1828445/yihui-test.zip)

yihui:
I'm sure this issue has been reported at least once before, although I don't remember which issue it was. Usually I don't recommend using `output_dir` or `output_file` or `intermediate_dir`. They are known to be problematic, e.g. https://github.com/rstudio/rmarkdown/issues/587#issuecomment-168437646 #238 #348 #574 #1033 #861 etc. I believe the only reliable strategy is to render everything in the current directory, and move the output files later.

In short, yes, I admit this is a bug but I'm afraid it will take very long to fix.

NMarkgraf:
Sadly, @yihui, this is not an option for us. There must be a place where the absolute path is added, and the only thing needed there is to replace all "\" by "/". - That will take to long? If you tell me where, I will try to fix it. ... :-(

NMarkgraf:
Okay @yihui,

after looking at the code I will ask you to give the following a try:

in render.R starting at line 363 you have got the following:

```
[...]

PR #1293, opened 2018-03-21
Bug fixing #1289

NMarkgraf:
This will fix the absolute path bug I reported in #1289.

### DIFF
--- a/R/render.R
+++ b/R/render.R
@@ -363,7 +363,8 @@ render <- function(input,
     # use filename based figure and cache directories
     base_pandoc_to <- gsub('[-+].*', '', pandoc_to)
     if (base_pandoc_to == 'html4') base_pandoc_to <- 'html'
-    figures_dir <- paste(files_dir, "/figure-", base_pandoc_to, "/", sep = "")
+    loc_files_dir <- file.path(output_dir, knitr_files_dir(basename(output_file)))
+    figures_dir <- paste0(loc_files_dir, "/figure--", base_pandoc_to, "/")
     knitr::opts_chunk$set(fig.path = figures_dir)
     cache_dir <- knitr_cache_dir(input, base_pandoc_to)
     knitr::opts_chunk$set(cache.path = cache_dir)


R/render.R
```diff
@@ -363,7 +363,8 @@ render <- function(input,
     # use filename based figure and cache directories
     base_pandoc_to <- gsub('[-+].*', '', pandoc_to)
     if (base_pandoc_to == 'html4') base_pandoc_to <- 'html'
-    figures_dir <- paste(files_dir, "/figure-", base_pandoc_to, "/", sep = "")
+    loc_files_dir <- file.path(output_dir, knitr_files_dir(basename(output_file)))
+    figures_dir <- paste0(loc_files_dir, "/figure--", base_pandoc_to, "/")
```
yihui: There is an extra `-` after `/figure`.

NMarkgraf:
Yes, but working absolute paths on windows systems and *nix systems!

Von meinem iPhone gesendet

Am 21.03.2018 um 20:53 schrieb Yihui Xie <notifications@github.com<mailto:notifications@github.com>>:


@yihui commented on this pull request.

I tested your fix and it didn't make a difference. It still generates absolute paths.

________________________________

In R/render.R<https://github.com/rstudio/rmarkdown/pull/1293#discussion_r176217049>:

> @@ -363,7 +363,8 @@ render <- function(input,
     # use filename based figure and cache directories
     base_pandoc_to <- gsub('[-+].*', '', pandoc_to)
     if (base_pandoc_to == 'html4') base_pandoc_to <- 'html'
-    figures_dir <- paste(files_dir, "/figure-", base_pandoc_to, "/", sep = "")
+    loc_files_dir <- file.path(output_dir, knitr_files_dir(basename(output_file)))
+    figures_dir <- paste0(loc_files_dir, "/figure--", base_pandoc_to, "/")


There is an extra - after /figure.

yihui:
Okay. Perhaps I didn't look carefully enough and didn't see any differences. I'll take a look again tomorrow morning, or you could paste the intermediate .tex files before/after the changes in this PR and show the differences.

NMarkgraf:
Will do. this today, in a few minutes!

NMarkgraf:
Currently I am working on converting it to a relative path. But that is quit difficult, I guess. That explains the Travis Ci build fails above! (so forget #1293)

The point in the pull request #1289 fixes the problem with calling pandoc_path_arg(..) much to early.

This converts the "c:/absolute/path/to/working/directory/of/files" into  "c:\absolute\path\to\working\directory\of\files" witch leads to a mixed slash/backslash absolute paths in the LaTeX calls, on Windows systems only(!!!), like "\includegraphics{c:\absolute\path\to\working\directory\of\files/document_files/figure-beamer/afigure.png}". And LaTeX can't handle this mixture and thinks "\" is an escape sequence, so it doesn't handle "\a", "\p", "..." correct and thous errors.

(Again: on Windows systems with output_dir set a a parameter in `render()` calls only!)

Using the unconverted version of of "files_dir", as I will do it in the first pull request,  will also put in an absolute path, but a working one also on Windows systems. Because I avoid the mixture of slash and backslash notation. Be aware, this only happens when you are using an output_dir parameter in the `render()` call. So the kntir-button in RStudio is not affected in any way, and will not leads to the errors in LaTeX!

If you like, I will try to make a relative path version out of it. But this may take a while. The path above (without the extra "-" at the end of "/figure-" will do it and does not changes to current behavior on all other systems.

yihui:
I don't know why, but on my Windows, the forward slash `/` is always used in paths in RStudio.

NMarkgraf:
"The answer, my friend, is showing in the code...!" -- Moment, I have to turn off Bob in my mind...

Here we are.

Using RStudio and the kntir button will normally do it right. As I told you, we are using the `rmarkdown::render()` with a given `output_dir` parameter. And that will make the difference!

Let us examine the code ("rmarkdown/R/render.R", line 14ff):

```
render <- function(input,
                   output_format = NULL,
                   output_file = NULL,
                   output_dir = NULL,
                   output_options = NULL,
                   intermediates_dir = NULL,
                   knit_root_dir = NULL,
                   runtime =  c("auto", "static", "shiny", "shiny_prerendered"),
                   clean = TRUE,
                   params = NULL,
                   knit_meta = NULL,
                   envir = parent.frame(),
                   run_pandoc = TRUE,
                   quiet = FALSE,
                   encoding = getOption("encoding")) {
```

Normaly `output_dir` is `NULL`, but we use something like `render(...,output_dir="build",...)`in the function call. (Keep that in memory, please!)

The first time this is used follows in line 114:

```
 # resolve output directory before we change the working directory in
  # preparation for rendering the document
  if (!is.null(output_dir)) {
    if (!dir_exists(output_dir))
      dir.create(output_dir, recursive = TRUE)
    output_dir <- normalize_path(output_dir)
  }
```

Nothing strange happens here, because `normalize_path()` don't harm anything. Because you use the following in "rmarkdown/R/includes.R":

```
# simple wrapper over normalizePath that preserves NULLs and applies pandoc-
# friendly defaults
normalize_path <- function(path,
                           winslash = "/",
                           mustWork = NA) {

  if (!is.null(path))
    normalizePath(path, winslash = winslash, mustWork = mustWork)
}
```

So the "/" will not changed to "\" in this function call.

Now line 250 back in "render()" again:

```
  # if an output_dir was specified then concatenate it with the output file
  if (!is.null(output_dir)) {
    output_file <- file.path(output_dir, basename(output_file))
  }
  output_dir <- dirname(output_file)
```

This will also not change any "/" to "\". So perfect until here.

But now look an the next lines, where `files_dir` is created (in line 256ff):

```
  # use output filename based files dir
  files_dir <- file.path(output_dir, knitr_files_dir(basename(output_file)))
  files_dir <- pandoc_path_arg(files_dir)
```
 
You see the function call `pandoc_path_arg(files_dir)`? - Let's examine that call, by take a look at the code (line 291 in "pandoc.R"):

```
pandoc_path_arg <- function(path,
                            backslash = TRUE) {

  path <- path.expand(path)

  # remove redundant ./ prefix if present
  path <- sub('^[.]/', '', path)

  if (is_windows()) {
    i <- grep(' ', path)
    if (length(i))
      path[i] <- utils::shortPathName(path[i])
    if (backslash) path <- gsub('/', '\\\\', path)
  }

  path
}

```

We call this function with `backslash=TRUE`, so the line ` if (backslash) path <- gsub('/', '\\\\', path)` will make a "\" out of every "/" in the path on windows systems (only). - Okay?

You may try this, because `pandoc_path_args()` is exported it'll be easy. On a non windows system everything is unchanged. BUT:

Try this on your windows machine:

```
> rmarkdown::pandoc_path_arg("c://test/")
[1] "c:\\\\test\\"
```
You see, it is converted to backslash. - On any other machine it will stay unchanged:
```
> rmarkdown::pandoc_path_arg("c://test/")
[1] "c://test/"
```

Now take a look back an line 250 in "render.R":

```
  # if an output_dir was specified then concatenate it with the output file
  if (!is.null(output_dir)) {
    output_file <- file.path(output_dir, basename(output_file))
  }
  output_dir <- dirname(output_file)
```

if `output_dir` was not `NULL` we have a `file.path()` call, which will add at an "/". Okay maybe the`dirname()` call  will remove this. But it will not be removed, if `output_dir` is an absolut path, or is a bit more complex relative path like "build/me/here".

Again, this won't happened if you using RStudio kntir-button, because `output_dir` will normally  be `NULL` there.

And also, normally this isn't all a bad idea to do, until ... line  366:
```
    figures_dir <- paste(files_dir, "/figure-", base_pandoc_to, "/", sep = "")
    knitr::opts_chunk$set(fig.path = figures_dir)
```

In most (and in all working) cases files_dir is empty here. So `figures_dir` will become "/figure-beamer/" for example. But this is untrue if you use `output_dir` with a not NULL parameter.
Even on all non windows systems, it would be fine! Because all the subdirectory will be separated by slashes "/" in the same direction. 

BUT: On windows systems it will leads to the situation as described.

You may want to test it? - Okay use this:

```
knitr_files_dir <- function(file) {
    paste(tools::file_path_sans_ext(file), "_files", sep = "")
}
output_dir <- "build"
output_file <- "document.pdf"
# if an output_dir was specified then concatenate it with the output file
if (!is.null(output_dir)) {
    output_file <- file.path(output_dir, basename(output_file))
}
output_dir <- dirname(output_file)
# use output filename based files dir
files_dir <- file.path(output_dir, knitr_files_dir(basename(output_file)))
files_dir <- rmarkdown::pandoc_path_arg(files_dir)
files_dir 
``` 

On a non windows machine you will get "build/document_files", as excepted, BUT on a windows machine it'll be: "build\\document_files".

Again, this will cause no trouble unless line 366 (recall it?):

```
    figures_dir <- paste(files_dir, "/figure-", base_pandoc_to, "/", sep = "")
```

If you add this (and a value for "base_pandoc_to") to the example:

```
knitr_files_dir <- function(file) {
    paste(tools::file_path_sans_ext(file), "_files", sep = "")
}
output_dir <- "build"
output_file <- "document.pdf"
# if an output_dir was specified then concatenate it with the output file
if (!is.null(output_dir)) {
    output_file <- file.path(output_dir, basename(output_file))
}
output_dir <- dirname(output_file)
output_dir
# use output filename based files dir
files_dir <- file.path(output_dir, knitr_files_dir(basename(output_file)))
files_dir <- rmarkdown::pandoc_path_arg(files_dir)
files_dir 
base_pandoc_to <- "test"
figures_dir <- paste(files_dir, "/figure-", base_pandoc_to, "/", sep = "")
knitr::opts_chunk$set(fig.path = figures_dir)
figures_dir
```

You can see the different results:

non-windows:
```
> figures_dir
[1] "build/document_files/figure-test/"
```

windows:
```
> figures_dir
[1] "build\\document_files/figure-test/"
```

And that is exactly the point, where LaTeX get's its problems with the "\includegraphics{}"-paths from.

So, please @yihui, take one of my pull requests, or fix it by yourself. I hope I could prove to you or at least convinced you, that there is an error in the code, and where there is this error. The rest is up to you.

Yours
Norman

yihui:
What is your `rmarkdown::pandoc_version()`? I tested both 1.19.2.1 (the version that RStudio currently ships) and 2.x, and it seems both versions will normalize the backslashes to forward slashes for me on Windows. Although I understand your analysis of the source code (which was absolutely correct), I just cannot reproduce the problem.

yihui:
Anyway, please test #1302:

```r
devtools::install_github('rstudio/rmarkdown#1302')
```

I said in #1289 that it might take long to fix it because I'm not the original author of this part of the code in `render()`, and I feel nervous about manipulating file paths in this package (`output_file`, `output_dir`, and `intermediate_dir` look complicated to me). I can surely figure out a fix in a few minutes to your particular issue, but I have no idea if it will have unexpected consequences. The other reason it may take long is that I really hope we could use relative paths in this case if possible, but again, I have no idea if it will break other things. #1302 is the most conservative fix I could think of at the moment.

Issue #1183, opened 2017-10-31
Error when Knitting: --latex-engine has been removed.


dylanjm:
Hello, I'm getting an error when trying to knit a document. 

```
processing file: wsj_07.Rmd
output file: wsj_07.knit.md

/usr/local/bin/pandoc +RTS -K512m -RTS wsj_07.utf8.md --to latex --from markdown+autolink_bare_uris+ascii_identifiers+tex_math_single_backslash --output wsj_07.pdf --template /usr/local/lib/R/3.4/site-library/rmarkdown/rmd/latex/default-1.17.0.2.tex --highlight-style tango --latex-engine /Library/TeX/texbin/pdflatex --variable graphics=yes --variable 'geometry:margin=1in' 
--latex-engine has been removed.  Use --pdf-engine instead.
Try pandoc --help for more information.
Error: pandoc document conversion failed with error 2
Execution halted
```

At first I thought it might have been an issue with the Template Package I've been developing but this error is now also showing up on a vanilla pdf .RMD file. This error did not start happening until I updated & upgraded homebrew this evening. 

# Edit:

I see you guys are aware of this issue as of the most recent Pandoc Update. Will leave this here in case anything has changed since those posts.

yihui:
Yes, two other users have reported this issue. I'll investigate it tomorrow and see if the Pandoc bug https://github.com/jgm/pandoc/issues/2618 is back again. Thanks!

yihui:
Oops just read your error message, and this issue actually has been solved. Please do

```r
devtools::install_github("rstudio/rmarkdown")
```

dylanjm:
@yihui installed the github version of markdown and am now getting this error:

```
output file: homework_18.knit.md

/usr/local/bin/pandoc +RTS -K512m -RTS homework_18.utf8.md --to latex --from markdown+autolink_bare_uris+ascii_identifiers+tex_math_single_backslash --output homework_18.pdf --template /usr/local/lib/R/3.4/site-library/HomeworkBYUI/rmarkdown/templates/homework/resources/homework_template.tex --highlight-style tango --pdf-engine /Library/TeX/texbin/pdflatex 
pdf-engine /Library/TeX/texbin/pdflatex is not compatible with output format latex
Error: pandoc document conversion failed with error 1
Execution halted
```

yihui:
That was reported at https://github.com/rstudio/rmarkdown/issues/1120#issuecomment-340574769 and I'll investigate it soon.

[...]

PR #1301, opened 2018-03-23
latex_engine -> pdf_engine (ugly)

gloaec:
Had to deal with pandoc 2 for the `latex_engine` -> `pdf_engine` major change. Had a presentation on monday. So had to do this with emergency  : 

    grep latex_engine . -Rl | xargs sed -i "s@latex_engine@pdf_engine@g"

Ugly, but it worked. So for anyone having the same problem, you can **TEMPORARILY** get the package from

    devtools::install_github("gloaec/rmarkdown")

cf. #1183


yihui:
Thanks, but this is not necessary because the `latex_engine` argument is compatible with Pandoc 2.x:
https://github.com/rstudio/rmarkdown/blob/37044614f6762a5237344be55bac6d58d894eb06/R/pandoc.R#L246

The reason you had problems in https://github.com/rstudio/rmarkdown/issues/1183#issuecomment-375747250 was because you used a wrong argument name: it should be `latex_engine` instead of `pdf_engine`. The `beamer_presentation` format does not have an option named `pdf_engine`.

Yes, Pandoc 2.x uses the command-line argument `--pdf-engine` instead of `--latex-engine`, but that does not mean we have to rename the R function argument `latex_engine` to `pdf_engine`. In fact, we should not do this because we will lose backward compatibility: the argument `latex_engine` has been working for years, and we cannot just get rid of it. There could be a lot of Rmd documents in the wild that use the `latex_engine` option, and they will be broken after this PR.

gloaec:
@yihui Thanks for your answer and for the clear explanation.

Indeed, I didn't dig that far into the code. For some reason, I couldn't make it work yesterday (even with yaml label `latex_engine`, I guess i was just to tired...)

Anyway, it works now. Thanks for your great work and for the care you give to the community.

PR #1302, opened 2018-03-23
Fix #1289: do not use backslashes in files dir

### LINKED ISSUE #1289: Problems with absolute paths in plots under LaTeX (Windows)

yihui:
Because at least they don't work for LaTeX output.

As I said in #1293 in response to @NMarkgraf, I was not able to reproduce the backslashes (I always got forward slashes in .tex output), but I agree we should always use forward slashes in figure paths.


yihui:
Your remark makes sense. I was aware of the potential risk, but I guess it should be fine (I had checked all instances of `files_dir` in this function).

Regarding relative paths: yes, you can send a PR, but I hope you could help us test it for a couple of weeks with your own Rmd documents before we start to review the PR. Using relative paths is a good idea in my eyes, but much riskier than changing backslashes to forward slashes. Again, I'm not the one who introduced the `output_dir` argument, so I cannot guarantee this PR will be merged soon or ever. Thanks!

yihui:
Eventually I have taken your advice: I only changed `\` to `/` in `figures_dir` instead of `files_dir`. It will be great if you can help test the latest development version. Thanks!

NMarkgraf:
I will check it after Eastern!

Issue #1328, opened 2018-04-27
R Markdown Websites no longer work with non `html_document` formats


jjallaire:
R Markdown Websites were originally designed to work with any HTML based format (so long as it supports the site_libs and self_contained options). However, a recent change intended to address this issue (https://github.com/rstudio/rmarkdown/issues/793) has broken this:

https://github.com/rstudio/rmarkdown/pull/1177

So now the feature which originally supported any HTML based format can only work with html_document.

The original issue #793 is certainly also a legitimate use case (rendering PDFs alongside HTML documents) but this particular fix is too drastic, so I think should be reverted in preference to another fix (PR to follow with the revert, I'm not at all sure what the alternate fix should be, perhaps instantiating the output format and checking it's formals before forwarding on those parameters?).

PR #1329, opened 2018-04-27
don't force the use of html_document with render_site

jjallaire:
fixes https://github.com/rstudio/rmarkdown/issues/1328

### DIFF
--- a/R/output_format.R
+++ b/R/output_format.R
@@ -538,6 +538,17 @@ output_format_from_yaml_front_matter <- function(input_lines,
 
   # merge any output_options passed in the call to render
   if (!is.null(output_options)) {
+
+    # create the output format function so we can check it's formals. strip options
+    # not within the formals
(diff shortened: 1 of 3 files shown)


jjallaire:
The PR now includes a fix to the original issue #793. 

The fix is to strip any options not supported by the target format from `output_options`. Note that this is a much more conservative fix than another possible approach which would have been to always strip unsupported arguments whenever we create a format. This would have also solved the problem but then left YAML options completely unchecked (so e.g. using the wrong option name would just silently fail rather than throwing an error).

This fix also creates a situation where explicitly passed `output_options` are also unchecked, but this parameter is very infrequently used and when used it is often by orchestration/build tools which (as in the case of Rmd websites) will generally find it convenient to pass options that may or may not be supported by the underlying format.

@yihui If you could review when you have the chance it would be much appreciated.

PR #1333, opened 2018-05-01
Add custom-made spinner replacement .gif

### LINKED ISSUE #1263: [Feature Request] New Loading GIF
Pedantic request here but would it be possible to update the 'Loading...' gif that appears prior to rendering a flexdashboard? It feels a bit outdated and pixelated compared to the 'Please wait...' version that appears on regular shiny apps.

Initially filed this in the flexdashboard repo but was advised to file here instead.

Thanks!

### LINKED ISSUE #1328: R Markdown Websites no longer work with non `html_document` formats

rich-iannone:
Fix #1263 I created a custom spinner gif (using the RStudio blue from the new orb). It's not a pure CSS solution but it does look decent as a drop-in replacement and has a relatively small file size.


yihui:
Here is the spinner used on ShinyApps.io:

spinner

jjallaire:
I agree we should use the same one as ShinyApps.io if possible....

rich-iannone:
New loading .gif is in place. I tested it with an example flexdashboard .Rmd and it looks great.

rich-iannone:
@yihui That would involve remaking the .gif. This is something I could totally do, can't guarantee it will look exactly the same but it should be very close. I'll also try to get it to be low in byte size.

yihui:
I meant `tags$p("Please wait...")` (instead of the text in the image).

rich-iannone:
Oh yes, of course! I will do that.

R/shiny.R
```diff
@@ -373,7 +373,7 @@ rmarkdown_shiny_ui <- function(dir, file) {
           id = "rmd_loader_wrapper",
           tags$div(id = "rmd_loader", style = "display: none",
                    tags$img(src = "rmd_resources/rmd_loader.gif"),
-                   tags$p("Loading")))),
+                   tags$p("Please wait...")))),
```
yihui: Please move the text above the gif (that is what ShinyApps.io shows on the loading page).

rich-iannone:
You're welcome. BTW, I'm learning a lot!

PR #1337, opened 2018-05-10
fix typo

maxheld83:


maxheld83:
I thought just an `à` would be fine, but apparently, it is not. Fixed it.

Sorry, this small typo is causing more hassle than it's worth.

Blame the french.

PR #1340, opened 2018-05-11
Add Font awesome 5.0.13 support

rich-iannone:
This replaces the FA 4.5.0 components with new CSS and font files. There are 3 .ttf files now (brands, regular, and solid). The shim JS file is included here to help with the transition from v4 to v5.

One can minimally test this is working with:

````
---
title: a title
output: html_document
---

```{r, include=FALSE}
htmltools::tagList(rmarkdown::html_dependency_font_awesome())
```

<i class="fab fa-r-project"></I>
````

The R project icon does appear.


yihui:
I'm slightly in favor of no version numbers in the dir names of JS libraries (only specify the versions in `htmlDependency()`), which will make it easier to see the diffs when we upgrade these libraries in the future (otherwise we can only see that one library is deleted and one is added), although we probably don't really care about the diffs.

jjallaire:
Aren't the regular icons pro-only?

rich-iannone:
I've removed to version numbers in the dir names. I do think the regular icons are pro-only.

rich-iannone:
@jjallaire I will check that loading works in those instances you mentioned.

Issue #1343, opened 2018-05-14
Vertical spacing after htmlwidgets


aoles:
Hi,

Similar to paragraphs or code chunks, I believe html widgets should have the small 10px spacer between the bottom of the widget and the succeeding content. Otherwise they "stick" on top of the blocks following them, as it is currently the case (see the example bellow). This could be easily fixed by adding the CSS rule to the default HTML template:

```css
.html-widget {
  margin: 0 0 10px;
}
```

I'm looking forward to your thoughts on this.

Best,
Andrzej

````
---
output: html_document
---

```{r}
library(DT)
datatable(cars)
```

```{r}
library(leaflet)
leaflet() %>% addTiles()
```

```{r}
## code
```
````

<img width="883" alt="widgets" src="https://user-images.githubusercontent.com/6545356/39987753-ceb17fe0-5765-11e8-8424-a4166544ae72.png">

jjallaire:
I agree. Here's a PR to address this: https://github.com/rstudio/rmarkdown/pull/1344

PR #1344, opened 2018-05-14
Add 20px bottom margin to htmlwidgets

jjallaire:
This matches the amount of padding/margin after `<pre>` tags. R plots normally have some internal margin and in addition are enclosed in `<p>` which has a bottom margin of 1em.

### DIFF
--- a/inst/rmd/h/default.html
+++ b/inst/rmd/h/default.html
@@ -139,6 +139,9 @@ img {
 .tabbed-pane {
   padding-top: 12px;
 }
+.html-widget {
+  margin-bottom: 20px;
+}
 button.code-folding-btn:focus {
   outline: none;
 }


aoles:
Many thanks @jjallaire for addressing https://github.com/rstudio/rmarkdown/issues/1343. Could you maybe briefly clarify why did you choose the bottom margin of htmlwidgets to be 20px rather than 10px? By inspecting the output of `html_document` it seems to me that the default spacing after `<p>` and `<pre>` tags is set to 10px. Not sure where exactly this is specified, though.

jjallaire:
I did that b/c most R plots have some internal padding (whereas widgets often don't).

aoles:
Thanks for clarifying!

PR #1298, opened 2018-03-21
Use metadata variable `pagetitle` for Pandoc 2.0+

rich-iannone:
When using Pandoc 2.0+, this changes addresses a requirement for a page title in the input document. Closes #1290

Issue #1290, opened 2018-03-20
find_external_resources needs pandoc 2.0 fix


hadley:
```R
rmarkdown::find_external_resources("~/Documents/tidyverse/purrr/README.Rmd")
#> [WARNING] This document format requires a nonempty <title> element.
#>   Please specify either 'title' or 'pagetitle' in the metadata.
#>   Falling back to 'file35ae52634c27.utf8'
#>                   path explicit  web
#> 1 man/figures/logo.png    FALSE TRUE
```

I think the documentation should also mention that pandoc is run, as I found that somewhat surprising.

yihui:
@rich-iannone I think this will be a relatively easy fix (see 6558c3ce3d785133182b12a161fcc4928829e537). Could you do it?

rich-iannone:
Yes! I will do it.

rich-iannone:
@yihui when I was reproducing this output, I got regular output without a warning:

```
rmarkdown::find_external_resources("~/Documents/tidyverse/purrr/README.Rmd")
#>                   path explicit  web
#> 1 man/figures/logo.png    FALSE TRUE
```

To reproduce I downloaded a zip of the purrr repo and placed in my file system in the same location so that the same path to file works. Could it be that this is already fixed, or am I missing something?

yihui:
@rich-iannone I believe @hadley was using Pandoc 2.x.

rich-iannone:
@yihui thanks for that. Will install v2+ Pandoc and then get back into the issue.

PR #1355, opened 2018-05-18
Avoid pagetitle warning from pandoc2.0 when title is missing.

jdblischak:
I noticed the warning produced by pandoc 2+ when the R Markdown file does not have a title in the YAML header. This is the same warning as demonstrated for `find_external_resources` in #1290. The fix in #1298 removes the warning only for `find_external_resources`.

This PR makes the change to the pre_processor function of `html_document_base`, and thus it removes the warning for anything that inherits from this function (including `find_external_resources`).

I also made two other minor improvements:

1. Don't specify `pagetitle` if there is a `title` specified in the YAML header. This keeps the text displayed in the browser tab the same as the document title (which is the same behavior as pandoc 1.0)
1. Instead of using `"PREVIEW"`, use the basename of the input file. This will make it easier to identify the browser tab. Also, this is similar to pandoc's fallback option.

cc @rich-iannone

I tested this with R 3.3.3, pandoc 2.2.1, and macOS 10.10.5. I'm happy to test this on other operating systems if there is interest in merging this PR.

I've previously signed the individual contributor agreement.

### DIFF
--- a/R/html_document_base.R
+++ b/R/html_document_base.R
@@ -108,6 +108,10 @@ html_document_base <- function(smart = TRUE,
 
     preserved_chunks <<- extract_preserve_chunks(input_file)
 
+    # Avoid pagetitle warning from pandoc2.0 when title is missing
+    if (pandoc2.0() && is.null(metadata$title))
+      args <- c(args, "--metadata", paste0("pagetitle=", input_file))
+
     args
   }
 
--- a/R/html_resources.R
+++ b/R/html_resources.R
@@ -377,8 +377,7 @@ discover_rmd_resources <- function(rmd_file,
   html_file <- render(input = md_file, output_file = html_file,
                       output_format = override_output_format,
                       output_options = list(
-                        self_contained = FALSE,
-                        pandoc_args = c("--metadata", "pagetitle=PREVIEW")),
+                        self_contained = FALSE),
                       quiet = TRUE,
                       encoding = "UTF-8")
 


yihui:
@jdblischak Sorry I decided to revert this PR, because it has brought a couple of new issues that would be too hard to fix. The first one was #1434, which was not too bad. The second one was #1607, which is not trivial to fix. I think the case #1607 is bad enough that it is worth reverting this PR.

I don't know what your original motivation for this PR was (i.e. why the title could be missing in **workflowr**). If I understand the problem better, I might be able to come up with a direct fix to that problem.

jdblischak:
@yihui No worries. I apologize that my PR caused you and others so much difficulty.

My original motivation was too remove the pandoc2 warning about a missing title:

```
f <- tempfile()
writeLines(c("```{r}", "1 + 1", "```"), con = f)
rmarkdown::render(f, quiet = TRUE)
## [WARNING] This document format requires a nonempty <title> element.
##   Please specify either 'title' or 'pagetitle' in the metadata.
##   Falling back to 'file63fd452f6059.utf8'
sessioninfo::package_info("rmarkdown")["rmarkdown", ]
##  package   * version date       lib source                            
##  rmarkdown   1.18.4  2019-12-04 [1] Github (rstudio/rmarkdown@3d648d3)
## 
## [1] /home/jdb-work/R/x86_64-pc-linux-gnu-library/3.6
## [2] /usr/local/lib/R/site-library
## [3] /usr/lib/R/site-library
## [4] /usr/lib/R/library
```

I don't bother adding a title to most of the Rmd files I use to test workflowr, so all of those pandoc2 warnings were obscuring the true error messages whenever a test failed. I'll continue to handle this downstream in workflowr.

yihui:
Sounds good. Thanks @jdblischak!

Issue #848, opened 2016-10-31
redundant usage of metavariable `lang` and `babel-lang`


steffenWagner:
Dear `rmarkdown`-team!

I am not sure, if I oversee something regarding the `"default"` pandoc templates for `rmarkdown::pdf_document`, but I think, there could be a simplified approach steering the language options via the YAML metavariables.

Currently there are four pandoc tex templates stored in `inst/rmd/latex`. The two older template versions

  - default.tex
  - default-1.14.tex

only use the meta variable `lang` via

```tex
\documentclass[... $if(lang)$$lang$,$endif$ ... ]{$documentclass$}
...
$if(lang)$
\ifxetex
  \usepackage{polyglossia}
  \setmainlanguage{$mainlang$}
\else
  \usepackage[shorthands=off,$lang$]{babel}
\fi
...
```

The two newer templates 

  - default-1.15.2.tex
  - default-1.17.0.2.tex

use
```tex
\documentclass[... $if(lang)$$babel-lang$,$endif$ ... ]{$documentclass$}
...
$if(lang)$
\ifnum 0\ifxetex 1\fi\ifluatex 1\fi=0 % if pdftex
  \usepackage[shorthands=off,$for(babel-otherlangs)$$babel-otherlangs$,$endfor$main=$babel-lang$]{babel}
$if(babel-newcommands)$
  $babel-newcommands$
$endif$
...
```
Therefore it is necessary in the latter case, to hand over both metavariables `lang` and `babel-lang` via the YAML header to steer the language settings.

By changing the two newer tex templates to 

```tex
\documentclass[... $if(lang)$$lang$,$endif$ ... ]{$documentclass$}
...
$if(lang)$
\ifnum 0\ifxetex 1\fi\ifluatex 1\fi=0 % if pdftex
  \usepackage[shorthands=off,$for(babel-otherlangs)$$babel-otherlangs$,$endfor$main=$lang$]{babel}
$if(babel-newcommands)$
  $babel-newcommands$
$endif$
...
```
only one metavariable would be required to steer the language for the babel package and in addition downwards compatability of the YAML-header and therefore the final Rmd document could be realised.

yihui:
Since you know what to change, could you submit a pull request? Thanks!

rich-iannone:
@steffenWagner we had a discussion about this change and found it was a bit risky to implement without knowing about any downstream effect. If you're still interested, could you file an issue related to this at https://github.com/jgm/pandoc/issues ?

PR #1374, opened 2018-06-18
Addresses redundant usage of metavariable

rich-iannone:
This is a PR related to #848 . The OP was invited to create a pull request but this never occurred. I carefully made the two edits that were recommended to avoid a redundant usage of a metavariable (`lang` -- `babel-lang`) in the two most recent `default...` tex files.


rich-iannone:
That's a good idea. I will remove this PR and carry on with the other actions in the issue.

PR #1389, opened 2018-06-29
add natbiboptions pandoc variable

npjc:
There is no current way to achieve:

```latex
\usepackage[numbers,sort&compress]{natbib}
```

I changed the template to expose the [`natbiboptions`](https://pandoc.org/MANUAL.html#variables-for-latex) pandoc latex variable. Now it optionally allows 1 or more `natbiboptions` to be passed via `pandoc_args`.

in `_output.yml`:
```yaml
rmarkdown::pdf_document:
  citation_package: natbib
  pandoc_args: [
      "--variable", "natbiboptions=numbers",
      "--variable", "natbiboptions=sort&compress"
    ]
```

### DIFF
--- a/inst/rmd/latex/default-1.17.0.2.tex
+++ b/inst/rmd/latex/default-1.17.0.2.tex
@@ -93,7 +93,7 @@ $endfor$
 \fi
 $endif$
 $if(natbib)$
-\usepackage{natbib}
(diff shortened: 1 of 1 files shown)


npjc:
@jjallaire I believe you rstudio has a signed individual contributor agreement from me but it was long ago and so please let me know if I'm mistaken.

PR #1390, opened 2018-06-29
Depend on htmldeps dev package (rstudio/htmldeps) for provision of some html_dependency() fcns

rich-iannone:
This adds `rstudio/htmldeps` to Remotes in DESCRIPTION. The functions in that package are named equivalently to what was in `rmarkdown`.

Inside `htmldependencies.R` in `rmarkdown`, the equivalently-named functions in `htmldeps` are re-exported and exist in the NAMESPACE. Vendor files for the re-exported functions have been removed from `rmarkdown` as they exist in `htmldeps`.

Minimal changes to documentation or attributions have been done. I suppose the full attribution should be equivalently done in `rmarkdown` and in `htmldeps`. A future PR could address consistency in attributions across both packages.


jjallaire:
It doesn't look like the `html_dependency` functions are in fact re-exported (all of the `export` entries have been removed from the NAMESPACE, which will break any package that calls these functions). Here is how we do re-exports in keras: https://github.com/rstudio/keras/blob/master/R/reexports.R

R/html_dependencies.R
```diff
@@ -87,9 +50,9 @@ navbar_icon_dependencies <- function(navbar) {
 html_dependencies_fonts <- function(font_awesome, ionicons) {
   deps <- list()
   if (font_awesome)
-    deps <- append(deps, list(html_dependency_font_awesome()))
+    deps <- append(deps, list(htmldeps::html_dependency_font_awesome()))
```
jjallaire: You won't need any of these explicit qualifiers once you've done the re-export (since they'll all be available as functions within the rmarkdown NAMESPACE)

rich-iannone:
Sorry about the confusion here. Had R CMD Check provide a warning on a somewhat related thing (Remotes in DESCRIPTION) and throught the way I exported was incorrect. Thanks for linking to the keras repo for an example.

R/html_dependencies.R
```diff
@@ -1,4 +1,4 @@
-#' @import htmltools
+#' @import htmltools htmldeps
```
jjallaire: I believe that since you are doing the explicit imports you don't need the `htmldeps` here as well

rich-iannone: Will remove, thanks!

yihui:
@rich-iannone @jjallaire We should release **htmldeps** to CRAN as early as possible so it won't block the next CRAN release of **rmarkdown** (there may be at least one release before rstudio::conf) or further increase the R version requirement of **rmarkdown**: https://github.com/rstudio/htmldeps/issues/4

rich-iannone:
@yihui @jjallaire I think we ought to make the decision whether to continue with **htmldeps** as a package (since its value may have diminished since **shiny** can't take advantage) or put back the dependencies into **rmarkdown**. I'm somewhat in favor of discontinuing the package unless there are other benefits I'm missing.

yihui:
If **shiny** can't take advantage of it, I tend to put everything back in **rmarkdown**.

rich-iannone:
Okay, I will follow up with a PR shortly that will do just that.

yihui:
Sounds good.

@jjallaire What do you think?

jjallaire:
Agreed, let's put everything back in rmarkdown.

Issue #1292, opened 2018-03-20
[DOC] unclear behaviour when both output_dir and output_file are provided


NelleV:
One can both provide output_dir and output_file at the same time. How `rmarkdown::render` behaves is right now unclear in the documentation.

rich-iannone:
@NelleV Thanks for opening this issue awhile back. It's now being addressed in a pull request.

NelleV:
Thanks for letting me know!

rich-iannone:
You're welcome. It's just been merged into the dev version of R Markdown.

PR #1392, opened 2018-07-02
Modify documentation for clarity (#1292)

rich-iannone:
This addresses #1292 and brings some clarity to how the options `output_dir` and `output_file` may interact if used together. These changes also produce a `render.Rd` file from roxygen (it's no longer manually edited).


R/render.R
```diff
@@ -10,6 +10,140 @@
...
+#' @param output_file The name of the output file. If using \code{NULL} then the
+#' output filename will be based on filename for the input file. If a filename
+#' is provided, a path to the output file can also be provided. Note that the
+#' \code{output_dir} option allows for specifying the output file path as well.
+#' Please also note that any directory path provided will create any necessary
+#' directories if they do not exist.
+#' @param output_dir The output directory for the rendered \code{output_file}.
+#' This allows for a choice of an alternate directory to which the output file
+#' should be written (the default output directory of that of the input file).
+#' If a path is provided with a filename in \code{output_file} the directory
```
NelleV: :+1:

man/render.Rd
```diff
@@ -107,8 +114,8 @@ Additional format options can also be specified in metadata. For example:
 \preformatted{
 output:
   pdf_document:
-    toc: true
-    highlight: zenburn
+  toc: true
+highlight: zenburn
```
yihui: It is great to move Rd to roxygen, but you need to be a little cautious about the `\preformatted` tags, in which white spaces are actually meaningful. For example, here the indentation is wrong.

man/render.Rd
```diff
@@ -117,49 +124,46 @@ is passed to \code{render} then the first one defined will be used:
...
-    toc: true
-    highlight: zenburn
-  html_document:
-    toc: true
-    theme: united
+  toc: true
+highlight: zenburn
+html_document:
+  toc: true
+theme: united
```
yihui: Same problem here.

rich-iannone:
@yihui I should refer to that document more often because I guessed (wrongly) at the keyword :( I swear I’ll get this right the next time though!

man/render.Rd
rich-iannone: I can fix these. Thought the formatting survived the copy/paste but no.

yihui: Please feel free to push to master for these two changes. Thanks!

rich-iannone: Just did this. Thanks for reviewing!

PR #1401, opened 2018-07-13
Stop `render()` early when duplicate chunk names found

### LINKED ISSUE #1276: More descriptive error message when two snippets have the same name
Whenever two snippets have the same name (identifier), it is not possible to "knit" any output, which technically is the way it should be.

The problem is the resulting error message, which is difficult to interpret/understand. 
A more descriptive one would help a lot in identifying duplicate snippet names, which can easily arise from copying and pasting code.

rich-iannone:
These changes will stop the render process early when two or more chunks have the same name. It adds a utility function to parse the lines of the input file, yielding `TRUE` if any duplicate chunk names are detected. This function is called in `render()` just after reading the input file. This will save this user time when rendering any long-running input files where chunk names are not carefully checked for duplicates (e.g., ad-hoc analyses with DBs, etc.).

If duplicate chunk names are detected, `stop()` is called with an informative error message.

Fixes #1276

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -69,6 +69,25 @@ read_lines_utf8 <- function(file, encoding) {
   to_utf8(lines, encoding)
 }
 
+# Determine if there are duplicate chunk names in the document's
+# `input_lines`; this allows the render process to terminate early
+# for any long-running rendering tasks
+any_duplicate_chunk_names <- function(input_lines) {
+
+  # Get only the lines related to chunk headers
+  chunk_headers <- input_lines[which(grepl("^```\\{\\w+?", input_lines))]
+
+  # Remove any components not related to the chunk names
+  chunk_names <-
+    gsub("^```\\{[a-zA-Z]+?\\s*\\}$|```\\{[a-zA-Z]+?\\s+?|\\}$|,.*|[a-zA-Z0-9 ]*?\\s*=\\s*.*|\\s*\\}", "",
+         chunk_headers, perl = TRUE)
+
+  # Remove remaining whitespace
+  chunk_names <- gsub("^\\s*", "\\s*$", chunk_names)
+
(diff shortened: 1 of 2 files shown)


R/util.R
```diff
@@ -69,6 +69,25 @@ read_lines_utf8 <- function(file, encoding) {
   to_utf8(lines, encoding)
 }
 
+# Determine if there are duplicate chunk names in the document's
+# `input_lines`; this allows the render process to terminate early
+# for any long-running rendering tasks
+any_duplicate_chunk_names <- function(input_lines) {
```
jjallaire: @yihui Are there any functions or regexs exported from knitr that would facilitate this being done in the most reliable/compatible fashion possible?

R/util.R
```diff
@@ -69,6 +69,25 @@ read_lines_utf8 <- function(file, encoding) {
...
+  # Get only the lines related to chunk headers
+  chunk_headers <- input_lines[which(grepl("^```\\{\\w+?", input_lines))]
+
+  # Remove any components not related to the chunk names
+  chunk_names <-
+    gsub("^```\\{[a-zA-Z]+?\\s*\\}$|```\\{[a-zA-Z]+?\\s+?|\\}$|,.*|[a-zA-Z0-9 ]*?\\s*=\\s*.*|\\s*\\}", "",
+         chunk_headers, perl = TRUE)
+
+  # Remove remaining whitespace
+  chunk_names <- gsub("^\\s*", "\\s*$", chunk_names)
```
yihui: I think you meant `gsub("^\\s+|\\s+$", "", chunk_names)`.

rich-iannone: Ugh... yeah.

rich-iannone:
I'll recommend to the OP that this issue belongs in knitr. I'll close the PR after that's done.

PR #1405, opened 2018-07-16
Implement dropdown in tabsets

### LINKED ISSUE #1116: Feature Request: Dropdown Options in Tabsets
For tabsets created in an HTML document by rmarkdown, tabs are by default displayed horizontally, like so:

image
However, with a little bit of tweaking of the HTML, you can turn the tabset into a dropdown menu:

image

Instead of having to tweak the HTML after Pandoc produces the HTML file, would it be possible to use rmarkdown to specify a dropdown layout? Perhaps the result could look something like this?
```markdown
# Choose a Tab {.tabset .tabset-fade .tabset-dropdown}
## First Tab
Here is the first tab's content.
## Second Tab
Here is the second tab's content
```

For reference, the rmarkdown for this example is as follows:

```markdown
---
title: "Tabset Example"
output: html_document
---

# The Tabset Section {.tabset .tabset-fade}

## First Tab
Here is the first tab's content.

## Second Tab
Here is the second tab's content
```

And the HTML produced for the tabset is this:

```html
<ul class="nav nav-tabs" role="tablist">
	<li role="presentation" class="active">
		<a role="tab" data-toggle="tab" href="#first-tab" aria-controls="first-tab">First Tab</a>
	</li>
	<li role="presentation">
		<a role="tab" data-toggle="tab" href="#second-tab" aria-controls="second-tab">Second Tab</a>
	</li>
</ul>
```
which can be replaced by the following HTML to produce a dropdown menu:
```html
 <ul class="nav nav-tabs" role="tablist">
	 <li class="dropdown">
		<a class="dropdown-toggle" data-toggle="dropdown" href="#" aria-expanded="false">
		  Choose a Tab <span class="caret"></span>
		</a>
		<ul class="dropdown-menu">
		  <li class=""><a href="#first-tab" data-toggle="tab" aria-expanded="false" aria-controls="first-tab">First Tab</a></li>
		  <li class=""><a href="#second-tab" data-toggle="tab" aria-expanded="false" aria-controls="second-tab">Second Tab</a></li>
		</ul>
	 </li>
</ul>
```

rich-iannone:
This allows for a dropdown menu to take the place of a standard tabset. It is used by adding the `.tabset-dropdown` class to a header (`# Heading {.tabset .tabset-dropdown}`). 

This fixes #1116 and incorporates the code provided by @stefanfritsch into `default.html`. This implementation, however, doesn't require an HTML include. An example .Rmd is provided here:

````
---
title: "tabset dropdown"
output:
  html_document
---

```{r setup, include=FALSE}
knitr::opts_chunk$set(echo = TRUE)
```

# R Markdown {.tabset .tabset-dropdown}

## Tab A

This is an R Markdown document. 

## Tab B

When you click the **Knit** button a document will be generated.

## Tab C

Here is a summary table.

```{r cars}
summary(cars)
```

## Tab D

Here is a plot:

```{r pressure, echo=FALSE}
plot(pressure)
```

# Here is another section {.tabset}

## Tab A

This is content in a standard tabset.

## Tab B

This is more content in a standard tabset.

````

### CODE REVIEW

**Comment by yihui** [comment]
And we should add @stefanfritsch as a contributor (`ctb`) in DESCRIPTION if the code was borrowed from #1116. Ideally, we need him to sign the contributor agreement since the code is not trivial, otherwise we'd better implement this independently by ourselves.

**Comment by stefanfritsch** [comment]
I'd be happy to sign the contributor agreement

**Comment by jjallaire** [comment]
@yihui I agree that the default html template is bloated, but this doesn't add a huge amount of new content (maybe 1KB?) so I don't think it warrants the effort to make it conditional (font awesome was much more heavyweight). i.e. Users who care about payload size won't be convinced they should use the format b/c documents are 1KB leaner.

**Comment by yihui** [comment]
@jjallaire I'm suggesting a separate dependency because 1) this PR adds 60 lines of code to our HTML default template that currently has 500 lines of code (i.e. increasing the number of lines by 10% for a corner feature); 2) I really hope we could factor out some of the CSS/JS code into separate HTML dependencies, too, and this could be a start. My point of view was completely from the package maintenance side (managing 10 orthogonal modules, each of 50 lines, is easier than managing a 500-line library in my eyes).

That said, I'm not totally against adding this directly and unconditionally to the template. We could factor out CSS/JS in the future when we have a chance (we probably never will, given that this is not really a critical issue).

**Comment by yihui** [comment]
@stefanfritsch Thanks! You can find the agreement at https://github.com/rstudio/rmarkdown/blob/master/CONTRIBUTING.md

**Comment by jjallaire** [comment]
@yihui I agree that this could be a good start, but I also think that the "real" solution might be as you said to refactor everything to make it more lean by default. I'd be in favor of creating an issue for this and then pursuing it soon rather than externalizing just this code ad-hoc.

**Comment by yihui** [comment]
Sounds good. Let's merge this PR when we get the signed agreement from @stefanfritsch.

**Comment by jjallaire** [comment]
Agreement received from @stefanfritsch

**Comment by rich-iannone** [comment]
I can definitely reformat the CSS so that it fits in with the rest. One question: should I deactivate the RStudio setting `Insert spaces for tab` to get true space characters?

**Comment by yihui** [comment]
No. You should leave it checked. But this setting is only effective when you type and hit the Tab key in the editor. It does not automatically convert tabs to spaces when you copy and paste.

**Comment by rich-iannone** [comment]
Okay, turns out I have spaces but RStudio, when moving over the spaces with arrow keys, gives the impression of tabs. Will make the changes now.

**Comment by yihui** [comment]
I wouldn't notice it, either, had I not looked at the diff page here on Github (tabs look wider than four spaces). In RStudio, a tab does look like four spaces visually (two double spaces).

**Comment by rich-iannone** [comment]
@yihui could you have a quick look at the new NEWS item and see that it fits stylistically?

**Comment by yihui** [comment]
Looks good. I tested it and noticed a vertical scrollbar:

image

I wonder if that is expected. I hope to get rid of it if it is not too difficult.

**Comment by rich-iannone** [comment]
Let me try to get rid of that scrollbar.

**Comment by yihui** [comment]
Yes, it is easy to suppress a scrollbar, but I hope we can figure out the root reason why the scrollbar is there in the first place. If you just forcibly hide it, you may end up (partially) hiding content, too.

**Comment by rich-iannone** [comment]
I agree. Let me do some further research/experimentation on this.

**Comment by yihui** [comment]
This article might be helpful: https://css-tricks.com/findingfixing-unintended-body-overflow/ I have used its JS code before to fix a horizontal scrollbar, and I'm not sure if the samething can be done for vertical overflow.

**Comment by rich-iannone** [comment]
Thanks! This helps.

**Comment by yihui** [comment]
Using `inline-table` does make the scrollbar disappear, although I don't fully understand why (this is beyond my expertise now). There will be a minor issue only in Firefox, though:

image

Note that the bottom border of the tab title is missing. I'm fine with this. It is quite trivial compared to the scrollbar issue.

**Comment by rich-iannone** [comment]
I've tried to address the Firefox issue but it's not as simple as removing the white background (that element also inherits another `#fff` background). It appears that the background in Firefox is drawn one pixel lower.

Given we could accept this, should I this merge now?

File: inst/rmd/h/default.html [hunk 1]
```diff
@@ -232,10 +232,71 @@
...
 <script>
 $$(document).ready(function () {
   window.buildTabsets("$idprefix$TOC");
 });
+
+$$(document).ready(function () {
+  $$('.tabset-dropdown > .nav-tabs > li').click(function () {
+    $$(this).parent().toggleClass('nav-tabs-open')
+  });
+});
```
jcheng5: This (295-299) can be improved using jQuery event delegation, it's a good habit to get into:

https://learn.jquery.com/events/event-delegation/

PR #1406, opened 2018-07-17
Allow `run()` to work if `file = NULL` and multiple .Rmd files are present

### LINKED ISSUE #906: error when attempting to rmarkdown::run a directory with multiple Rmd
When trying to narrow another issue, I hit this:

```
mkdir rmd_dir
touch rmd_dir/rmd_dir.Rmd
touch rmd_dir/rmd_dir.html
touch rmd_dir/foo.html
```

```
> setwd("~/dev/rstudio/play/rmd_dir")
> rmarkdown::run(file = NULL, dir = "~/dev/rstudio/play/rmd_dir", auto_reload = FALSE)
Error in ifelse(!is.null(file), file, default_file) : 
  replacement has length zero
In addition: Warning message:
In rep(no, length.out = length(ans)) :
  'x' is NULL so the result will be NULL
```

rich-iannone:
This allows `run()` to work if `file = NULL` and multiple .Rmd files are present. This selects the first .Rmd file based on alphabetical order. The function documentation has been amended to make this clear. This fixes #906

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -5,6 +5,9 @@ rmarkdown 1.11 (unreleased)
 
 * The icons in `flexdashboard::valueBox()` are not of the full sizes due to the upgrade of FontAwesome in #1340 in the previous version (#1388, rstudio/flexdashboard#189).
 
+* Fix for `run()` to work if `file = NULL` and multiple .Rmd files are present in the search
+path (#1406).
+
 rmarkdown 1.10
 --------------------------------------------------------------------------------
 
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -108,6 +108,13 @@ run <- function(file = "index.Rmd",
     }
   }
 
+  # If no `default_file` yet chosen but multiple .Rmd
+  # files exist in the directory, choose the first based
+  # on an alpha sort
+  if (is.null(default_file) && length(allRmds) > 1) {
+    default_file <- sort(allRmds)[1]
+  }
+
   # form and test locations
   dir <- normalize_path(dir)
   if (!dir_exists(dir))
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -25,8 +25,10 @@
 #' order of preference):
 #' \itemize{
 #'   \item{If \code{dir} contains only one \code{Rmd}, that \code{Rmd}.}
-#'   \item{The file \code{index.Rmd}, if it exists in \code{dir}}
-#'   \item{The file \code{index.html}, if it exists in \code{dir}}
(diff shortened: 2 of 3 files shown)


yihui:
Could you also add a NEWS item?

rich-iannone:
Definitely will do!

NEWS
```diff
@@ -5,6 +5,9 @@ rmarkdown 1.11 (unreleased)
 
 * The icons in `flexdashboard::valueBox()` are not of the full sizes due to the upgrade of FontAwesome in #1340 in the previous version (#1388, rstudio/flexdashboard#189).
 
+* Fix for `run()` to work if `file = NULL` and multiple .Rmd files are present in the search
+path (#1406).
```
yihui: It will be good to mention #906, too.

nit: we have never hard-wrapped lines in NEWS.md.

rich-iannone: Just changed this. Thanks! Do we cite PR and issue numbers interchangeably?

R/shiny.R
```diff
@@ -108,6 +110,13 @@ run <- function(file = "index.Rmd",
     }
   }
 
+  # If no `default_file` yet chosen but multiple .Rmd
+  # files exist in the directory, choose the first based
+  # on an alpha sort
+  if (is.null(default_file) && length(allRmds) > 1) {
```
yihui: Sorry for not testing it earlier. I should have done it myself, but I just tried to reproduce #906, and it does not seem to be reproducible any more _without_ this PR (e.g. with the CRAN version), i.e., the issue seems to have been fixed somehow (note that it was filed two years ago). So I came back and looked at the first `if` statement above. I believe the bug in #906 should not exist any more because there was only one Rmd in that case:

```r
     if (length(allRmds) == 1) {
       # just one R Markdown document
       default_file <- allRmds
```

Therefore this PR was not fixing the bug, but introducing a new behavior instead.

I tend to revert this PR, but improve the documentation -- currently the last `else` clause is undocumented (i.e. the logic under the comment `# look for first one that has runtime: shiny`).

NEWS
yihui: It is not strictly necessary to cite both, but redundancy doesn't hurt here (show both the bug number and PR number).

rich-iannone: Okay, great, will do. Just noticed a ton of line breaks in NEWS at earlier version numbers. Could I take a pass through it and remove them all? (A consistency play.)

yihui: Yes, please (if you have time -- apparently this is not very important).

Issue #1291, opened 2018-03-20
Fail early when output_file's directory do not exist


NelleV:
Right now, Rmardown compiles Rmd files, and then moves the file to the wished filename. It thus fails after compiling the whole document if the output_file's directory do not exist.

It should be fairly straightforward to do an early check, to avoid the compilation to fail at the very last moment.

rich-iannone:
@NelleV Output directories will be created so long as there are write permissions in parent dirs. Are we asking for a check on whether new dirs can be created due to the permissions issues (and if they can't be created, fail fast)?

BTW I've had this exact issue in the past, where I would set up a `render()` though **cron** but the output folder (or a parent, if it was new) didn't have write permissions. The workaround is that you would pre-plan this and use **chmod** as necessary.

yihui:
@rich-iannone I just created a minimal repro:

```r
test_output_file = function() {
  d = tempfile()
  dir.create(d)
  owd = setwd(d)
  on.exit({
    setwd(owd)
    unlink(d, recursive = TRUE)
  }, add = TRUE)
  
  writeLines('hello world', 'foo.Rmd')
  rmarkdown::render('foo.Rmd', output_file = 'does/not/exist/bar.html')
}

test_output_file()
```

Output:

```
processing file: foo.Rmd
  |.................................................................| 100%
  ordinary text without R code


output file: foo.knit.md

/Applications/RStudio.app/Contents/MacOS/pandoc/pandoc +RTS -K512m -RTS foo.utf8.md --to html4 --from markdown+autolink_bare_uris+ascii_identifiers+tex_math_single_backslash+smart --output does/not/exist/bar.html --email-obfuscation none --self-contained --standalone --section-divs --template /Users/yihui/R/rmarkdown/rmd/h/default.html --no-highlight --variable highlightjs=1 --variable 'theme:bootstrap' --include-in-header /var/folders/8r/zh8x49md6vsgskh23jrmy7p80000gn/T//RtmpPkWF5w/rmarkdown-str11fff3a56ed55.html --mathjax --variable 'mathjax-url:https://mathjax.rstudio.com/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML' --metadata pagetitle=foo.utf8.md 
pandoc: does/not/exist/bar.html: openFile: does not exist (No such file or directory)
 Error: pandoc document conversion failed with error 1 
```

rich-iannone:
[...]

PR #1410, opened 2018-07-18
Make corrections for issue #1291

rich-iannone:
This is in a separate PR because I re-forked the repo. This addresses the recommended changes in PR #1408 .

### DIFF
--- a/R/render.R
+++ b/R/render.R
@@ -80,14 +80,14 @@ metadata <- list()
 #' @param output_file The name of the output file. If using \code{NULL} then the
 #' output filename will be based on filename for the input file. If a filename
 #' is provided, a path to the output file can also be provided. Note that the
-#' \code{output_dir} option allows for specifying the output file path as well.
-#' Please also note that any directory path provided will create any necessary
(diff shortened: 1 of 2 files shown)


R/render.R
```diff
@@ -387,6 +387,12 @@ render <- function(input,
   }
   output_dir <- dirname(output_file)
 
+  # Stop the render process early if the output directory does not exist
+  if (!dir_exists(output_dir)) {
+    stop("The directory `", output_dir, "`) does not not exist.",
```
yihui: nit: usually I used quotes, either double or single, instead of backticks in messages, since they feel more natural in text for humans. Backticks feel like Markdown-ish.

rich-iannone:
@yihui I think I got my local/remote versions of my fork in such a state that it was hard to get specific commits to appear in this PR.

R/render.R
rich-iannone: Could I do a simple replacement of backticks to single quotes in a push to master?

yihui: Yes, please go ahead. Thanks!

PR #1416, opened 2018-07-24
Modify pagedtable even background color

### LINKED ISSUE #1413: Darkly Theme Table Text Color
I'm so excited that Darkly is making its way into RMarkdown! However, after applying it to a document, I've noticed that the way tables are formatted makes it impossible to read every other row. Please see the image below. 

image

If you highlight the table, you can see that there is actually information in each of the white rows, they are just white text on a white background. 

image

This document was created with the following yaml header: 

```
---
title: "Title"
author: "Name"
date: "Last Update: `r Sys.Date()`"
output: 
  html_notebook:
    highlight: pygments
    theme: darkly
    toc: yes
    toc_float: yes
---
```

and the table was constructed with the following code chunk:
```{r}
read.csv("../Data/ExampleCSV.csv")
```

I've installed both from the 282ece4 commit and from the master and both seem to produce the same table style. 

Please let me know if I can provide any more supporting documentation

rich-iannone:
This fixes #1413. The issue is that the even-row cell background is a solid white, which doesn't work with themes that have a dark background and light fonts (only dark theme currently is the Bootswatch `darkly` theme).

This change modifies the CSS rule for the paged table `even` rows:

```
.pagedtable .even {
  background-color: #8b8b8b;
  background-color: rgba(140, 140, 140, 0.1);
}
```

By specifying a medium gray (RGB 140, 140, 140) and a very low value for opacity (0.1) we can achieve a tint that is still subtle and works with dark and light background and font colors (i.e., high contrast). I specified a static color in the line before for IE 8 and before (they don't support rgba).

Here are screenshots with the change applied for both the darkly and flatly themes:

darkly_change

flatly_change

### DIFF
--- a/inst/rmd/h/pagedtable-1.1/css/pagedtable.css
+++ b/inst/rmd/h/pagedtable-1.1/css/pagedtable.css
@@ -33,7 +33,8 @@
 }
 
 .pagedtable .even {
-  background-color: #fcfcfc;
+  background-color: #8b8b8b;
+  background-color: rgba(140, 140, 140, 0.1);
 }
 
 .pagedtable-padding-col {


rich-iannone:
Thanks for pointing me toward that page. I hadn't seen it before!

PR #1420, opened 2018-08-03
Force a render when user pkgs don't match the pkgs used in a rendered, shiny-prerendered document

schloerke:
Address: https://github.com/rstudio/learnr/issues/169

By keeping track of the package version, we can force a new render if the user does not have the same package versions as the pre-rendered document.  

This solves the error where a pre-rendered resource can not be mounted (because it doesn't exist!).

- [x] NEWS item
- [ ] test?
- [x] when in Shiny Server and RSC env set in learnr

### DIFF
--- a/R/shiny_prerendered.R
+++ b/R/shiny_prerendered.R
@@ -183,7 +183,6 @@ shiny_prerendered_prerender <- function(
     return(TRUE)
   }
 
-  message("look at packages used in dependencies")
   html_lines <- readLines(rendered_html, encoding = "UTF-8", warn = FALSE)
   dependencies_json <- shiny_prerendered_extract_context(html_lines, "dependencies")
   dependencies <- jsonlite::unserializeJSON(dependencies_json)
@@ -193,6 +192,8 @@ shiny_prerendered_prerender <- function(
     if (is.null(dep$package)) {
       # if the file doesn't exist at all, render again
       if (!file.exists(dep$src$file)) {
+        # might create a missing file compile-time error,
+        #   but that's better than a missing file prerendered error
         return(TRUE)
       }
     } else {
@@ -202,6 +203,7 @@ shiny_prerendered_prerender <- function(
         # has not seen pkg
 
         # depVer could be NULL, producing a logical(0)
+        #   means old prerender version, render again
         if (!isTRUE(packageVersion(depPkg) == depVer)) {
           # was not rendered with the same R package. must render again
           return (TRUE)
--- a/R/shiny_prerendered.R
+++ b/R/shiny_prerendered.R
@@ -216,20 +216,18 @@ shiny_prerendered_prerender <- function(
   }
   # all html dependencies are accounted for
 
+  # check for execution package version differences
   execution_json <- shiny_prerendered_extract_context(html_lines, "execution_dependencies")
   execution_info <- jsonlite::unserializeJSON(execution_json)
-
-  # check for execution package version differences
-  execution_pkgs <- execution_info$packages
-  versions_dont_match <- unlist(Map(
-    execution_pkgs$package,
-    execution_pkgs$version,
-    f = function(package, version) {
-      !identical(get_package_version_string(package), version)
+  execution_pkg_names <- execution_info$packages$package
+  execution_pkg_versions <- execution_info$packages$version
+  for (i in seq_len(nrow(execution_pkgs))) {
+    if (!identical(
+      get_package_version_string(execution_pkg_names[i]),
+      execution_pkg_versions[i]
+    )) {
+      return(TRUE)
     }
-  ))
-  if (any(versions_dont_match)) {
-    return(TRUE)
   }
   # all execution packages match
 
--- a/R/shiny_prerendered.R
+++ b/R/shiny_prerendered.R
@@ -219,11 +219,6 @@ shiny_prerendered_prerender <- function(
   execution_json <- shiny_prerendered_extract_context(html_lines, "execution_dependencies")
   execution_info <- jsonlite::unserializeJSON(execution_json)
 
-  # check that the major R versions match
-  if (!identical(R.version$major, execution_info$rMajorVersion)) {
-    return(TRUE)
-  }
-
   # check for execution package version differences
   execution_pkgs <- execution_info$packages
   versions_dont_match <- unlist(Map(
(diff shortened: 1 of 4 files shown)


jjallaire:
This LGTM and will be a huge improvement on what we have currently (which is obviously much too fragile).

rich-iannone:
Based on a code-pairing between me and Barret, there are some changes that incorporate R package dependencies. One thing that Barret noted was that the package dependencies (in the current form below) will incorporate packages like `graphics` and `methods` where their versions mirror the R version. It could be argued that these packages don't change much between different R versions so we shouldn't invalidate based on these specific packages. 

Here are the changes per file:

*util.R*

add:
```r
get_loaded_packages <- function() {

  packages <- setdiff(loadedNamespaces(), "base")
  loadedversion <- vapply(packages, getNamespaceVersion, "")
  attached <- paste0("package:", packages) %in% search()

  data.frame(packages = packages[attached], loadedversion = loadedversion[attached],
             row.names = NULL, stringsAsFactors = FALSE)
}
```

*render.R*

change (ln 362):
```r
output_options$dependency_resolver <- function(deps) {
  shiny_prerendered_dependencies <<- list(deps = deps, packages = get_loaded_packages())
  list()
}
```

*shiny_prerendered.R*

change (ln 184):
```r
dependencies <- lapply(shiny_prerendered_dependencies$deps, function(dependency) {
```

add (after ln 228):
```r
dependencies_pkgs_json <- jsonlite::serializeJSON(shiny_prerendered_dependencies$packages, pretty = FALSE)
shiny_prerendered_append_context(con, "package_dependencies", dependencies_pkgs_json)
```

yihui:
I think it is fine to ignore base R packages. The original problem was caused by packages that ship HTML dependencies. Base R packages don't have such things.

schloerke:
Documenting decision towards using loaded packages (after execution) vs detecting library and `::` calls in the code (without execution)

There were two main approaches:
1. It was already solved in `packrat` and to use something similar to `packrat::findDependencies`.  This function boils down to regular expressions and the name of the function calls.  This can easily be tricked, but can not be poisoned by the user workspace.
2. We could detect the loaded packages after executing all the R code in the document.  Looking at attached packages would only give results for library or require calls.  Looking at the search path would give all packages that have been loaded or used.  This can not be fooled like packrat, but can be poisoned by external user packages that were not a part of the knitting of the document.  

We decided to go with the loaded packages, as it can not be easily fooled and most renderings of a document are done within RStudio's clean environment and on hosted sites, which will compile in a clean environment.  The only caveat are with console users, where rendering is known to not be in a clean environment by default.

The good part about being aggressive with the _execution package_ version matching is that users will get a cohesive experience.  The worst interaction of not matching the _execution packages_ is having to re-render the shiny prerendered document.  This will be the first interaction of most any pre-render documents, but after re-rendering, it wont happen until the user (or service) updates their local packages.

yihui:
That sounds reasonable to me.

R/util.R
```diff
@@ -513,3 +513,34 @@ package_root <- function(path) {
...
+  )
+}
+# find all loaded packages.
+# May contain extra packages, but contains all packages used while knitting
+get_loaded_packages <- function() {
+  base_r_packages <- c(
+    "base", "compiler", "datasets", "graphics", "grDevices",
+    "grid", "methods", "parallel", "splines",
+    "stats", "stats4", "tcltk", "tools", "utils"
+  )
```
yihui: You can use `rownames(installed.packages(priority = 'base'))` instead of hardcoding the package names.

rich-iannone: Aha! I thought there was a better way (but I couldn't find it). Thanks!

R/shiny_prerendered.R
```diff
@@ -171,17 +135,121 @@ shiny_prerendered_html <- function(input_rmd, encoding, render_args) {
...
+  }
+  # all html dependencies are accounted for
+
+  execution_json <- shiny_prerendered_extract_context(html_lines, "execution_dependencies")
+  execution_info <- jsonlite::unserializeJSON(execution_json)
+
+  # check that the major R versions match
+  if (!identical(R.version$major, execution_info$rMajorVersion)) {
+    return(TRUE)
+  }
```
wch: Remove -- we can use exact version matching for base packages.

R/util.R
```diff
@@ -513,3 +513,30 @@ package_root <- function(path) {
...
+}
+# find all loaded packages.
+# May contain extra packages, but contains all packages used while knitting
+get_loaded_packages <- function() {
+  base_r_packages <- rownames(installed.packages(priority = 'base'))
+
+  packages <- sort(setdiff(loadedNamespaces(), base_r_packages))
+  version <- vapply(packages, get_package_version_string, character(1))
+  attached <- paste0("package:", packages) %in% search()
```
wch: We need to look at all loaded packages, not just the attached ones. For example, if someone does `library(shiny)`, then this will only record the version of shiny, but not dependencies like htmltools.

Another case is if they do `htmltools::div()`, it will not attach htmltools, and this code will also not record the version.

R/util.R
```diff
@@ -513,3 +513,30 @@ package_root <- function(path) {
...
+    as.character(getNamespaceVersion(package)),
+    error = function(e) {
+      NULL
+    }
+  )
+}
+# find all loaded packages.
+# May contain extra packages, but contains all packages used while knitting
+get_loaded_packages <- function() {
+  base_r_packages <- rownames(installed.packages(priority = 'base'))
```
wch: This can be removed -- we might as well look at the base R packages as well, since it's (marginally) safer to use them, and the code will be slightly simpler without excluding them. It also lets us not record the R major version because that comes along with these packages.

R/shiny_prerendered.R
```diff
@@ -171,17 +135,121 @@ shiny_prerendered_html <- function(input_rmd, encoding, render_args) {
...
+
+  # check for execution package version differences
+  execution_pkgs <- execution_info$packages
+  versions_dont_match <- unlist(Map(
+    execution_pkgs$package,
+    execution_pkgs$version,
+    f = function(package, version) {
+      !identical(get_package_version_string(package), version)
+    }
+  ))
```
wch: Change to `for` loop with a short-circuit return.

jjallaire:
We just need to make sure that Shiny Server and RSC run these in a way that ensures that no render occurs (on the assumption that they are in read-only directories). They can I believe do this via an environment variable (which they may or may not already be setting, worth checking on).

schloerke:
@yihui removed [WIP], ready for merge when you're ready.  Thank you!

PR #1433, opened 2018-08-21
Add intermediates generator function for the Powerpoint format

rich-iannone:
This allows for supporting input files to be collected and moved to the intermediates directory when `intermediates_dir` is specified. Fixes#1431

### DIFF
--- a/R/powerpoint_presentation.R
+++ b/R/powerpoint_presentation.R
@@ -45,6 +45,14 @@ powerpoint_presentation <- function(
   # pandoc args
   args <- c(args, pandoc_args)
 
+  saved_files_dir <- NULL
+
+  intermediates_generator <- function(original_input, encoding,
+                                      intermediates_dir) {
+    return(powerpoint_intermediates_generator(saved_files_dir, original_input,
+                                              encoding, intermediates_dir))
+  }
+
   # return output format
   output_format(
     knitr = knitr,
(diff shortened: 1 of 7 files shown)


R/powerpoint_presentation.R
```diff
@@ -54,6 +70,27 @@ powerpoint_presentation <- function(
...
+                                               original_input,
+                                               encoding, intermediates_dir) {
+
+  # copy all intermediates (pandoc will need to bundle them in the Powerpoint)
+  intermediates <- copy_render_intermediates(original_input, encoding,
+                                             intermediates_dir, FALSE)
+
+  # we need figures from the supporting files dir to be available during
+  # render as well; if we have a files directory, copy its contents
+  if (!is.null(saved_files_dir) && dir_exists(saved_files_dir)) {
```
yihui: It seems `saved_files_dir` is always `NULL`? I don't see where it can be assigned a different value.

R/powerpoint_presentation.R
```diff
@@ -54,6 +70,27 @@ powerpoint_presentation <- function(
...
   )
 }
+
+powerpoint_intermediates_generator <- function(saved_files_dir,
+                                               original_input,
+                                               encoding, intermediates_dir) {
+
+  # copy all intermediates (pandoc will need to bundle them in the Powerpoint)
+  intermediates <- copy_render_intermediates(original_input, encoding,
+                                             intermediates_dir, FALSE)
```
yihui: So basically this fix relies on `resource_files` in the YAML frontmatter. I hope that we don't have to use this field. Instead, automatically copy the file specified in the `reference_doc` argument to the intermediate dir. @slopp's example could be:

```yaml
output:
  powerpoint_presentation:
    reference_doc: rstudio.pptx
```

instead of

```yaml
output:
  powerpoint_presentation:
    pandoc_args: --reference-doc=./rstudio.pptx
resource_files:
- rstudio.pptx
```

The latest CRAN version of **rmarkdown** has the `reference_doc` argument in `powerpoint_presentation`, so @slopp I recommend that you use this argument instead of `pandoc_args` (which was a workaround I told Nathan quite a few months ago when Pandoc didn't support PPT templates).

R/powerpoint_presentation.R
```diff
@@ -8,10 +8,18 @@
...
+                                    fig_width = 5,
+                                    fig_height = 4,
+                                    fig_caption = TRUE,
+                                    df_print = 'default',
+                                    smart = TRUE,
+                                    keep_md = FALSE,
+                                    md_extensions = NULL,
+                                    slide_level = NULL,
+                                    reference_doc = 'default',
+                                    pandoc_args = NULL
```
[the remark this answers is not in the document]
rich-iannone: This has been reverted. I agree it is not really relevant to the PR at hand. I think I will do some (safe) reformatting in a separate PR.

Issue #1326, opened 2018-04-20
R Markdown websites share same environment when using `render_site()`?


andrewheiss:
This may be a bug, but it also might be the intended behavior for R Markdown. This comes from [an issue I posted on StackOverflow](https://stackoverflow.com/questions/49930372/use-new-environments-for-each-knitted-file-in-an-r-markdown-website). The reason I think it might be a bug is that when `blogdown` and `bookdown` compile a bunch of Rmd files, each one gets its own new environment, while `rmarkdown::render_site()` seems(?) to use one shared environment for all the Rmd files when knitting, which causes unexpected namespace issues.

For example, using this minimum working example with the following 5 files:

**testing.Rproj**

    Version: 1.0

    RestoreWorkspace: Default
    SaveWorkspace: Default
    AlwaysSaveHistory: Default

    EnableCodeIndexing: Yes
    UseSpacesForTab: Yes
    NumSpacesForTab: 2
    Encoding: UTF-8

    RnwWeave: Sweave
    LaTeX: pdfLaTeX

    AutoAppendNewline: Yes

    BuildType: Website

**\_site.yml**

    name: "testing"
    navbar:
      title: "Testing"
      left:
        - text: "Home"
          href: index.html
        - text: "Step 1"
          href: 01_something.html
        - text: "Step 2"
          href: 02_something-else.html

**index.Rmd**

    ---
    title: "Testing"
    ---

    Hello, Website!

**01_something.Rmd**

    ---
    title: "Step 1"
    ---

    Write some data just for fun.

    ```{r}
    library(tidyverse)
    library(here)

    write_csv(mtcars, file.path(here(), "cars.csv"))
    ```

**02_something-else.Rmd**

    ---
    title: "Step 2"
    ---

    This breaks because `lubridate::here()` and `here::here()` conflict, but *only* when rendering the whole site.

    ```{r}
    library(tidyverse)
    library(lubridate)
    library(here)

    # Do something with lubridate
    my_date <- ymd("2018-04-19")

    # Try to use here() and it breaks when rendering the whole site
    # It works just fine when knitting this file on its own, though, since here is loaded after lubridate
    cars <- read_csv(file.path(here(), "cars.csv"))
    ```

[...]

PR #1443, opened 2018-09-06
Render each file in site in a fresh R process

jennybc:
I have a (legacy) R Markdown website: <https://jennybc.github.io/purrr-tutorial/>

I'm having difficulty building the site with `rmarkdown::render_site()` because the search path is leaked from file to file and, therefore, undesirable and hard-to-control function masking is happening, even though each individual *file* has the correct `library()` calls. Each individual file can be rendered with no problems.

In another context, I would use explicit name spacing to solve this, but that's not so great in a pedagogical setting. Also, it just doesn't seem right that, by the time you render file 50, your search path reflects every `library()` call in files 1 - 49.

Is there any way this would be accepted? I'd be happy to refine.


jennybc:
I'm surprised tests pass with my first commit. They fail for me locally like so:

```
✖ |  6 1     | site [1.1 s]
─────────────────────────────────────────────────────────
test-site.R:54: failure: render_site does not leak search path across files
any(grepl("stringr", b, fixed = TRUE)) isn't false.
─────────────────────────────────────────────────────────
```

Is there any chance `skip_on_cran()` is somehow active on travis?

gaborcsardi:
Some extra motivation: in the near future we'll be able to render the files in parallel, in a bunch of worker processes.

jennybc:
OK let's move forward.

Any suggestions re: the name or key in `_site.yml` to request rendering each file in a separate, fresh process, i.e. `SOMETHING` below? Is there an existing term we should honour? Sounds like we should also keep the prospect of parallel execution in mind.

```
name: "my-website"
output_dir: "_site"
include: ["import.R"]
exclude: ["docs.txt", "*.csv"]
SOMETHING: true
```

Also, my current PR suppresses the usual output from the rendering. We just get the spinner provided by callr, which is probably too minimal. What do we want re: progress reporting?

yihui:
Oh naming. That's the hardest part. In bookdown, I used `new_session`: https://bookdown.org/yihui/bookdown/new-session.html Not sure if that is a good name.

`new_session` can be `true`, `false`, or `parallel` (again, not sure if this is a good idea--blending logical values with string values in the same option).

I don't have an opinion on progress reporting. Perhaps we can do that after someone files a feature request.

gaborcsardi:
I can help with progress reporting, it is actually easy, we can just get the output of the remote process, and print it to the screen.

gaborcsardi:
So the output goes like this:
```r
callr::r(
  function() { for  (i in 1:5) { print(i); Sys.sleep(1) } }, 
  block_callback = function(x) cat(x)
)
```

jennybc:
OK the new behaviour is triggered if and only if `new_session: true` is found in `_site.yml`. callr is moved to Suggests.

How is this looking @yihui?

Not yet handled:

  * Documentation
  * Progress reporting

jennybc:
This doesn't seem right but ... is the `.Rd` file for this hand-edited? Once that's clarified, I will find a place to mention the `new_session` configuration approach in the docs.

@yihui I think the implementation and test is ready for another look.

jdblischak:
I super support this PR. Please see this old issue for another motivating use case: #1326

jennybc:
I have it on good authority that yes this `.Rd` file is maintained by hand. I will make a separate PR to convert it to roxygen and then, once that is merged, merge master into this PR, to keep things separate.

jennybc:
Thanks @yihui!

The more I think about this, the more I think maybe `rmarkdown::render() itself should offer a `new_session` argument. So that it's easy for mere mortals to do with `rmarkdown::render()` what they do when they click the "Knit" button in RStudio. But that is a bigger change and something I would leave to the maintainers. But obviously, if that ever happened, then this PR could be undone and the `new_session: true` config would simply get wired up to that fictional argument.

PR #1466, opened 2018-10-24
Add pandoc_exec function to get pandoc binary path

### LINKED ISSUE #1465: Export pandoc()
I was wondering if you could export `pandoc()`.  There are a few pandoc commands that aren't covered by `pandoc_convert()` or `pandoc_version()` and would be nice to be able to run against the same version of pandoc **rmarkdown** is using. 

In my case I am calling `pandoc --print-default-data-file reference.pptx` to get the pandoc's internal template, which could be different places depending on the system and if the user has placed a template in their pandoc data directory. I'm currently doing so with a `getFromNamespace()` hack:

```
#' @importFrom utils getFromNamespace
get_default_pandoc_pptx <- function(outfile = tempfile(fileext = ".pptx")) {
  sys::exec_wait(getFromNamespace("pandoc", "rmarkdown")(),
    args = c("--print-default-data-file", "reference.pptx"),
    std_out = outfile)
  return(outfile)
}
```

noamross:
Closes #1465


R/pandoc.R
```diff
@@ -716,3 +716,20 @@ find_pandoc_theme_variable <- function(args) {
 pandoc2.0 <- function() {
   pandoc_available("2.0")
 }
+
+#' Get the path of the pandoc executable
+#'
+#' Returns the path of the pandoc executable used by functions in the the
+#' \bold{rmarkdown} package. This is the most recent version of pandoc found in
```
yihui: `\bold` -> `\pkg`

R/pandoc.R
```diff
@@ -716,3 +716,20 @@ find_pandoc_theme_variable <- function(args) {
...
+#' either the system path or shipped with RStudio.
+#'
+#' See the
+#' \href{http://pandoc.org/MANUAL.html}{pandoc manual}
+#' for pandoc commands.
+#' @examples
+#' \dontrun{
```
yihui: BDR used to ask authors to explain why they use `\dontrun{}`, so personally I have given it up. Since this example is trivial, I'd just remove it.

noamross:
Updated and squashed

PR #1475, opened 2018-11-04
Use internal HTML dependency functions (and no longer depend on `htmldeps`)

rich-iannone:
This puts `html_dependency_*()` functions back into **rmarkdown**. There is no longer a requirement to depend on the **htmldeps** package, so references to that package are removed.

This change is based on internal discussions and also at https://github.com/rstudio/htmldeps/issues/4.


yihui:
From the commit history, it seems you reverted things step by step. That was quite a bit of work and would also make us less confident to review/merge the PR since it is hard to tell if all changes in #1390 have been properly reverted. I'd have done `git revert 99c2016` since #1390 was squashed and merged. Or you can actually directly do this on Github. There is a "Revert" button in every PR, e.g. (bottom right)

image

Sorry for mentioning this too late. I'll use `git revert` myself to check if anything was missed.

yihui:
Just verified it and things looked okay. The major net change after #1390 and #1475 is that the version numbers are removed in the dependency folder names, e.g., `jquery-1.11.3/` was moved to `jquery/`. This will be a breaking change for users who have cached code chunks that contain HTML dependencies (paths to dependencies in the rmarkdown package will no longer be found). Personally I don't like version numbers in folder names, either, and I'm fine with this potential breakage.

jjallaire:
@rich-iannone We just made some changes to learnr related to these cached directories so you should check with @schloerke to make sure this interacts okay with those changes.

rich-iannone:
@yihui thanks for doing the checks afterward, I thought about reverting but I wanted to have the clean folder names. 

@jjallaire just spoke with @schloerke about the potential issue with **learnr** with this change and he thinks it won't be a problem.

Issue #654, opened 2016-03-23
scrollbar shown when highligthing with non-default theme (e.g. pygments) and code_folding


jjallaire:
With the following metadata:

``` yaml
title: "Untitled"
output:
  html_document:
    highlight: pygments
    code_folding: show
```

Scrollbars are shown under the code chunks:

<img width="745" alt="screen shot 2016-03-23 at 2 52 33 pm" src="https://cloud.githubusercontent.com/assets/104391/13997133/f222b574-f106-11e5-95b4-e0c448697e08.png">

This does not occur for `highlight: default`

p0bs:
Thanks again for this. Just to say that the scrollbar appears with 'highlight: pygments' on Windows, but not on mac.

jjallaire:
Fixed via https://github.com/rstudio/rmarkdown/commit/b343c980d5e42fba5663c2cf3034bd4b00064d5e

stefanavey:
I have this same issue when I upgraded pandoc from version 1.19.2.1 to 2.1.3. Is this an issue with Pandoc now?

rich-iannone:
@yihui I can definitely prioritize this one.

rich-iannone:
This is now reproducible to me on macOS (recent RStudio, pandoc v2.2.1) when the option `Show scroll bars` is set to `Always`. The code chunk does scroll horizontally regardless of this setting. Will look into a fix now.

PR #1489, opened 2018-12-04
WIP: fix rmarkdown regarding new theme behaviour in pandoc

### LINKED ISSUE #1471: Background color (using zenburn highlight) in R-chunks not showing correctly
When I create an R-notebook and then apply highlight (zenburn) and theme (journal), the background color of the R-chunk shows in grey color and it should be black.  Already tested with both `rmarkdown` versions, the devel and latest in production through CRAN.  I'm using the latest version of Rstudio (Version 1.1.463).

image
**System info:**

```
> sessionInfo()
R version 3.5.0 (2018-04-23)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: macOS High Sierra 10.13.6

Matrix products: default
BLAS: /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/3.5/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
 [1] yaml_2.2.0      xfun_0.4        tinytex_0.9     stringr_1.3.1   stringi_1.2.4  
 [6] mime_0.6        markdown_0.8    magrittr_1.5    knitr_1.20      jsonlite_1.5   
[11] htmltools_0.3.6 htmldeps_0.1.1  highr_0.7       glue_1.3.0      evaluate_0.12  
[16] digest_0.6.18   base64enc_0.1-3 Rcpp_0.12.19   

[1] packrat_0.4.9-3   rmarkdown_1.10.14 tools_3.5.0       rsconnect_0.8.8   compiler_3.5.0   
```

cderv:
This is a proposition to fix #1471. The issue resumes the problem introduced by recent changes in pandoc 2. The way pandoc > 2 insert css styling for highligting when converting to html conflicts with boostrap and theme. 

This proposition tries to adapt rmarkdown default html template to insert css rules that deals with that. 

This may be temporary if we try to report this impact to pandoc and find a clever solution that does not break this.

**This is work in progress** to get feedback and iterate. Also because I think the impact is rather broad among different format that rmarkdown convert to thanks to pandoc.

---
This change is [Reviewable](https://reviewable.io/reviews/rstudio/rmarkdown/1489)

### DIFF
--- a/R/html_document.R
+++ b/R/html_document.R
@@ -266,6 +266,10 @@ html_document <- function(toc = FALSE,
   else if (!is.null(template))
     args <- c(args, "--template", pandoc_path_arg(template))
 
+  # activate pandoc > 2 features in template
+  if (isTRUE(pandoc2.0()))
+    args <- c(args, pandoc_variable_arg("ispandoc2", "1"))
+
   # validate code_folding
   code_folding <- match.arg(code_folding)
 
@@ -297,10 +301,6 @@ html_document <- function(toc = FALSE,
   # additional css
   for (css_file in css)
     args <- c(args, "--css", pandoc_path_arg(css_file))
-
-  # deal with pandoc > 2 features in template
-  if (isTRUE(pandoc2.0()))
-    args <- c(args, pandoc_variable_arg("ispandoc2", "1"))
 
   # manage list of exit_actions (backing out changes to knitr options)
   exit_actions <- list()
--- a/R/html_document.R
+++ b/R/html_document.R
@@ -298,6 +298,10 @@ html_document <- function(toc = FALSE,
   for (css_file in css)
     args <- c(args, "--css", pandoc_path_arg(css_file))
 
+  # deal with pandoc > 2 features in template
+  if (isTRUE(pandoc2.0()))
+    args <- c(args, pandoc_variable_arg("ispandoc2", "1"))
+
   # manage list of exit_actions (backing out changes to knitr options)
   exit_actions <- list()
   on_exit <- function() {
--- a/R/html_document.R
+++ b/R/html_document.R
@@ -266,10 +266,6 @@ html_document <- function(toc = FALSE,
   else if (!is.null(template))
     args <- c(args, "--template", pandoc_path_arg(template))
 
-  # activate pandoc > 2 features in template
-  if (pandoc2.0())
-    args <- c(args, pandoc_variable_arg("ispandoc2", "1"))
-
   # validate code_folding
   code_folding <- match.arg(code_folding)
 
--- a/R/html_document.R
+++ b/R/html_document.R
@@ -267,7 +267,7 @@ html_document <- function(toc = FALSE,
     args <- c(args, "--template", pandoc_path_arg(template))
 
   # activate pandoc > 2 features in template
-  if (isTRUE(pandoc2.0()))
+  if (pandoc2.0())
     args <- c(args, pandoc_variable_arg("ispandoc2", "1"))
 
   # validate code_folding
--- a/inst/rmd/h/default.html
+++ b/inst/rmd/h/default.html
@@ -49,7 +49,7 @@ $endif$
 
 $if(highlighting-css)$
 <style type="text/css">code{white-space: pre;}</style>
-<style type="text/css">
+<style type="text/css" data-origin="pandoc">
 $highlighting-css$
 
 $if(code_folding)$
@@ -63,8 +63,28 @@ $if(code_folding)$
   margin-right: -30px;
 }
 $endif$
-
 </style>
+<!-- apply pandoc div.sourceCode style to pre.sourceCode instead -->
+<script>
+  (function() {
+    var sheets = document.styleSheets;
+    for (var i = 0; i < sheets.length; i++) {
+      if (sheets[i].ownerNode.dataset["origin"] !== "pandoc") continue;
+      try { var rules = sheets[i].cssRules; } catch (e) { continue; }
+      for (var j = 0; j < rules.length; j++) {
+        var rule = rules[j];
+        // check if there is a div.sourceCode rule
+        if (rule.type !== rule.STYLE_RULE || rule.selectorText !== "div.sourceCode") continue;
+        var style = rule.style.cssText;
+        // check if color or background-color is set
+        if (rule.style.color === '' || rule.style.backgroundColor === '') continue;
+        // replace div.sourceCode by a pre.sourceCode rule
+        sheets[i].deleteRule(j);
+        sheets[i].insertRule('pre.sourceCode{' + style + '}', j);
+      }
+    }
+  })();
+</script>
 $if(theme)$
 <style type="text/css">
   pre:not([class]) {
@@ -612,29 +632,6 @@ $if(mathjax-url)$
   })();
 </script>
 $endif$
-
-<!-- apply pandoc div.sourceCode style to pre.sourceCode instead -->
-<script>
-  (function() {
-    var stylesheets = document.styleSheets;
-    for(var i=0; i<stylesheets.length; i++){
-      var rules = stylesheets[i].cssRules;
-      for(var j=0; j<rules.length; j++){
-        var rule = rules[j];
-        // check if there is a div.sourceCode rule
-        if(rule.STYLE_RULE == 1 && rule.selectorText == "div.sourceCode"){
-          var style = rule.style.cssText;
-          // check if color or background-color is set
-          if(style.match('color|background-color')){
-            // replace div.sourceCode by a pre.sourceCode rule
-            stylesheets[i].deleteRule(j);
-            stylesheets[i].insertRule('pre.sourceCode{'+style+'}', j);
-          }
-        }
-      }
-    }
-  })();
-</script>
 
 </body>
 </html>
--- a/inst/rmd/h/default.html
+++ b/inst/rmd/h/default.html
@@ -52,17 +52,6 @@ $if(highlighting-css)$
 <style type="text/css">
 $highlighting-css$
 
-$if(ispandoc2)$
-pre.sourceCode  {
-  $if(highlight-color)$
-  color: inherit;
-  $endif$
-  $if(highlight-backgroundcol)$
-  background-color: transparent;
-  $endif$
-}
-$endif$
-
 $if(code_folding)$
 .sourceCode .row {
   width: 100%;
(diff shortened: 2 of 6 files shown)


cderv:
This simple insertion of to css rules fix the issue with `html_document`
image

but now, as pandoc 2 applies the highlighting rule to `div`, with `notebook` we have a the background color applied even when code is hidden. 

image

and when opening the code
image

Not sure we want that... I'll see if I can add something to change that.

yihui:
Perhaps we should check if the parent node is `div.sourceCode` before deciding where to insert the `div` containing the Show/Hide button:
https://github.com/rstudio/rmarkdown/blob/34d2fd0ebceb91705a54692ee5cfb46f0f09e28e/inst/rmd/h/navigation-1.1/codefolding.js#L48

If we are inside `div.sourceCode`, we need to insert this `div` to the parent node of `div.sourceCode`.

cderv:
Thanks for pointing this out. 
It is exactly what I am trying to do ! it took me some time to discover how rmarkdown inserted the collapsible code ! I do not know the 📦  inside / out ... yet 😉 
I am currently modifying this js file. Will update the PR soon.

cderv:
Ok for notebook I guess
code-folding

Small customisation pandoc css introduce 
```css
div.sourceCode {
    margin-top: 1em;
    margin-right: 0px;
    margin-bottom: 1em;
    margin-left: 0px;
}
```
As we insert `<div class='row'>` just before, I found the margin a little big. So I modified to reduce, using `em` also and not `px`
```css
div.row {
  /* to reduce the space introduce in pandoc div.sourceCode */
  margin-bottom: -0.8em;
}
```
🤔  not sure of that change...
I think I'll test several other format and option to see if something is wrong or if all is fixed.

inst/rmd/h/navigation-1.1/codefolding.js
```diff
@@ -45,7 +45,9 @@ window.initializeCodeFolding = function(show) {
     buttonCol.append(showCodeButton);
     buttonRow.append(buttonCol);
 
-    div.before(buttonRow);
+    var sourceCodeBlock = $('div.sourceCode');
```
yihui: Will this also work with Pandoc 1.9.x? i.e. Does Pandoc 1.9 also generate `div.sourceCode`?

cderv: This is part of the test I need to make. Yes pandoc 1.9.x have a `div.sourceCode` but the css applied to those part have changed. 

So yes this line work and the button is added above without issue, but the margin trick does not work. Will change that. 

Also the main fix 
```css
pre.sourceCode  {
  color: inherit;
  background-color: transparent;
}
```
does not work with pandoc 1.9 because in this pandoc version the background is applied to `pre.sourceCode` so it erase it... - 
if we want something that work with old and new, I need to find another one...  :confused:

cderv:
@yihui here is a proposition to make the fix conditional to pandoc > 2. As the issue is due to change from pandoc 2, we can activate a variable for pandoc template based on the version. This way the changes are only applied if used with pandoc 2. 
This also could be useful to add new features - at some point two template could exist.

This was for me the more obvious fix right now, rather than going with css mastery. (I did not find one that could work both with pandoc < 2 and pandoc > 2.

What do you think ?

inst/rmd/h/navigation-1.1/codefolding.js
cderv: ok - there is not `div.sourceCode` inserted if we use `highlight = 'default'`, whatever the version of pandoc. I need to deal with that.

cderv:
this should now work. I will try to test thoroughly. You can give it a try also. 

* `default` and `textmate` (not pandoc theme) are working. 
* 1.9 and > 2 works also in both case.
* new code is activated only when pandoc theme are added and when pandoc 2 is used

inst/rmd/h/default.html
```diff
@@ -60,10 +67,15 @@
   overflow-x: auto;
 }
 .code-folding-btn {
-  margin-right: -30px;
+  margin-right: 0px;
+}
```
yihui: This was introduced by #1399, and can be safely removed now.

inst/rmd/h/default.html
```diff
@@ -52,6 +52,13 @@
 <style type="text/css">
 $highlighting-css$
 
+$if(ispandoc2)$
+pre.sourceCode  {
+  color: inherit;
+  background-color: transparent;
```
yihui: Actually this will break themes that don't have a background color, such as `pygments` https://github.com/rstudio/rmarkdown/issues/654#issuecomment-380234085 Such themes should inherit the background color from Bootstrap's `pre`.

My head is starting to hurt...

R/html_document.R
```diff
@@ -266,6 +266,10 @@ html_document <- function(toc = FALSE,
   else if (!is.null(template))
     args <- c(args, "--template", pandoc_path_arg(template))
 
+  # activate pandoc > 2 features in template
+  if (isTRUE(pandoc2.0()))
```
yihui: I won't use `isTRUE()` here. If there is anything wrong with the function `pandoc2.0()` (e.g. if it returns `NA`), I'd rather see an error.

cderv: there is no more `isTRUE`

inst/rmd/h/default.html
cderv: I got rid of this rule.

inst/rmd/h/default.html
cderv: Yes it was not clear to me and I wanted to test it further. Thanks for the confirmation. This [blog post](https://www.garrickadenbuie.com/blog/2018/07/06/pandoc-syntax-highlighting-examples/) shows the result of pandoc styling, and indeed pygment as a white background, but with bootstrap it is not the case. As we need to 
* override the `pre` style of bootstrap because pandoc apply the background color and color to the parent
* inherits for the `pre` style of bootstrap to keep the `pre` css rule from bootstrap for theme that don't have background and color
I think we should also do something conditional. 

Of pandoc highlighting style, we have frome the [src code](https://github.com/jgm/skylighting/blob/master/skylighting-core/src/Skylighting/Styles.hs)
*  `pygments` with nothing ➡️ should inherit boostrap _color_ and _background-color_
* `tango` with no default color ➡️  should inherit boostrap color only
*  `haddock` is like `pygments`
*  `monochrome` is like `pygments`

I'll come up with something... 🤔 and yes 🤯

cderv: I made a proposition by manually setting which style has background and color or not. Then with two more template variable, the correct rule is activated. 

In the same idea, with pandoc2 we can do something more dynamic. I hesitate to introduce this in rmarkdown but it could be interesting. We can get highlight information with 
```r
  if (pandoc2.0()) {
    arg <- c("--print-highlight-style", pandoc_highlight)
    json_res <- system2(pandoc(), arg, stdout = TRUE)
    properties <- jsonlite::fromJSON(paste0(json_res, collapse = "\n"))
  }
```

we then get for _pygments_
```r
> properties$`background-color`
NULL
> properties$`text-color`
NULL
```

and for _tango_
```r
> properties$`text-color`
NULL
> properties$`background-color`
[1] "#f8f8f8"
```

We could introduce a function `get_higligthing_styles` in rmarkdown (that we could export) and use this here to test if `is.null` or not. This could even mean that we could fill in the value in
`pre.sourceCode` css rule, instead of _inherits_ or _transparent_

I run this by you before pushing anything. I have a working code locally though - I can share then revert.

cderv:
Initially, I did not want to touch what pandoc is inserting.  
I'll give it a go then. 

Rather than modified, could'nt we add `pre.sourceCode` to the rule if existing with background or color ?
```css
div.sourceCode, pre.sourceCode
  { color: #cccccc; background-color: #303030; }
```
kind of the same but maybe more compatible with pandoc ?

yihui:
Yes you certainly can. I guess the risk is minimal if we only modify this single rule from `div.sourceCode` to `pre.sourceCode`, though. If we define rules for both `div` and `pre`, there is a subtle disadvantage:

## pre.sourceCode only

image

## both div.sourceCode and pre.sourceCode

image

Did you notice the differences in the four corners? That's because Bootstrap defined rounded corners for `pre`. If the `div` has a background color, the corners will be filled.

Implementation-wise, you probably have to loop through `document.styleSheets` to find this particular stylesheet.

cderv:
oh ! I did not notice before reading what the difference was. I read the doc you linked to and some others, I agree that this require iteration. I'll work on that ! thanks for pointing the direction.

cderv:
I reverted back everything and deal with this using javascript code now. So just one file changed 😄 

Do you find this better and easier to maintain ?

I put the script at the end of the template but I wonder if it should be elsewhere... 🤔  Normally, if no pandoc highlighting is chosen, no need to insert this script. I hesitated to put it in this part, that i believe is activated only by using pandoc highlighting
https://github.com/rstudio/rmarkdown/blob/e91e4c27bb223b6b0380a185c6f2fe5fb2ee2404/inst/rmd/h/default.html#L50-65

Also, I think there is another small change. New pandoc 2.0 added the `div.sourceCode` part and applied also a css rule for margin (see in _skylighting_ [25](https://github.com/jgm/skylighting/pull/25/files) and [26](https://github.com/dbaynard/skylighting/commit/4183a820ae9b4d9932efd1009bdd6ae4da7ab524) that modified the [style](https://github.com/dbaynard/skylighting/blob/master/skylighting-core/src/Skylighting/Format/HTML.hs) applied by pandoc)

```css
div.sourceCode { margin: 1em 0; }
```
this create a difference in rendered rmarkdown. I think it is not really impacting but I let you decide if it should be dealt with.

inst/rmd/h/default.html
```diff
@@ -613,5 +613,28 @@ <h4 class="date"><em>$date$</em></h4>
 </script>
 $endif$
 
+<!-- apply pandoc div.sourceCode style to pre.sourceCode instead -->
+<script>
+  (function() {
+    var stylesheets = document.styleSheets;
+    for(var i=0; i<stylesheets.length; i++){
+      var rules = stylesheets[i].cssRules;
```
yihui: Need `try/catch` here since the `cssRules` property is not necessarily accessible: https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet#Notes

cderv:
Perfect ! So, my intuition was correct to move the hack. For rest, I am always learning - nice JS improvement.

Thanks @yihui for your review!

yihui:
You did 99% of the work, and I only did the rest 1% :)

PR #1502, opened 2018-12-18
extend latex_dependency to add extra lines

malcolmbarrett:
This PR adds the ability to add extra lines of LaTeX to the preamble when using `latex_dependency()`. What motivated this was needing package related options to the preamble that weren't in the style of `\pkg[option]`. For instance, using tikz, you often want other tikz libraries.

You can do this in the YAML file using `header-includes`:

```
header-includes: 
  - \usepackage{tikz} 
  - \usetikzlibrary{arrows}
```

But not with `latex_dependency()`. I've simply added `...` to that function and a way for `latex_dependencies_as_string()` to process these extra lines.

Since tikz is so popular and already well supported by R markdown, an alternative (or additional) approach is something similar to html dependency functions, e.g. `latex_dependency_tikz(tikzlibraries = c("arrows", "positioning"))`.


R/latex_dependencies.R
```diff
@@ -1,9 +1,10 @@
 #' Define a LaTeX package dependency
 #' @param name The LaTeX package name
 #' @param options The LaTeX options for the package
+#' @param ... Extra LaTeX code related to the package added to the preamble
 #' @export
-latex_dependency <- function(name, options = NULL) {
-  output <- list(name = name, options = options)
+latex_dependency <- function(name, options = NULL, ...) {
```
yihui: Could you just use `extra_lines` as the argument instead of `...`? Using a general `...` argument will make it slightly difficult for us to add other possible arguments to this function in the future.

R/latex_dependencies.R
```diff
@@ -30,10 +31,12 @@ latex_dependencies <- function(x = list()) {
 # in the head of a document
 latex_dependencies_as_string <- function(dependencies) {
   lines <- sapply(dependencies, function(dep) {
+    extra_lines <- paste(dep$extra_lines, collapse = ",")
```
yihui: Why collapsing by a comma? I feel it should be `\n` instead.

R/latex_dependencies.R
malcolmbarrett: Makes sense!

R/latex_dependencies.R
malcolmbarrett: Oops, typo :persevere: 

Fixed!

malcolmbarrett:
Any opinion on adding `latex_dependency_tikz()`?

malcolmbarrett:
Great, thanks! I added `latex_dependency_tikz()`. I made the documentation very similar to the HTML functions, even though there is only one right now, to be consistent and in case more get added. I can document the single function instead, if you prefer.

malcolmbarrett:
Thanks, @yihui!

PR #1509, opened 2019-01-10
Prevent clash with KOMA-Script classes

adunning:
This was added in #1284. Setting it by default, however, means that the document cannot be compiled if the document class is set to use KOMA-Script (e.g. `documentclass: scrbook`), or presumably any other class that provides `\subtitle`.

### DIFF
--- a/R/pdf_document.R
+++ b/R/pdf_document.R
@@ -176,7 +176,7 @@ pdf_document <- function(toc = FALSE,
 
     # use titling package to change title format to be more compact by default
     if (!has_yaml_parameter(input_test, "compact-title"))
-      args <- c(args, "--variable", "compact-title:no")
+      args <- c(args, "--variable", "compact-title:yes")
 
     if (length(extra_dependencies) || has_latex_dependencies(knit_meta)) {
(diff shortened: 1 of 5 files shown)


adunning:
Ah, sorry, I had assumed that the additions were made to the LaTeX template when the variable was added.

If you're only concerned about preserving the appearance of generated files between versions, what about injecting the extra titling material using header-includes? You could then simply use the default template from Pandoc, which would reduce maintenance (and remove problems right now with using Pandoc 2.x, since it's still trying to apply the older template).

yihui:
That is exactly what I have been hoping to do for years! I really hate the fact that we have been stuck in our own LaTeX template, which is both confusing to some users and difficult to maintain by ourselves. I just need to find some time to figure out the exact changes we made to Pandoc's default template, and see if it is possible to inject these changes via `header-includes` instead.

adunning:
I've modified this so that it does not change any settings, but uses `\providecommand` instead of `\newcommand` to prevent the clash with KOMA-Script (or any packages that provide their own `\subtitle` command).

adunning:
Thank *you*!

PR #1517, opened 2019-01-23
Adds vitae package to Formats page

garrettgman:
vitae is a community developed  R markdown extension for building Resumes and CVs, by Mitchell O'Hara-Wild.


garrettgman:
Thanks @yihui. You are right; they are rebuilt now.

PR #1518, opened 2019-01-23
Changes download links to RStudio Cloud links in the tutorial

garrettgman:
Replaces all file download links in the tutorial with links to RStudio Cloud projects that contain the files (and provide a way for readers to immediately follow along with the tutorial by knitting the files, etc.)


formats.html
```diff
@@ -105,81 +105,17 @@
...
 <script>
 $(document).ready(function () {
   window.buildTabsets("TOC");
 });
-
-$(document).ready(function () {
-  $('.tabset-dropdown > .nav-tabs > li').click(function () {
-    $(this).parent().toggleClass('nav-tabs-open')
-  });
-});
```
yihui: It seems you need to either exclude the changes in `formats.(Rmd|html)`, or rebase your branch against the latest `gh-pages` branch of the upstream repo.

PR #1525, opened 2019-02-01
Support image attributes in ioslides output

### LINKED ISSUE #1524: ioslides does not support Markdown image attribute (e.g. width)
In an RMarkdown presentation, I have a local image embedded via pandoc’s syntax:

```markdown
![Caption](path/to/image.png){width=90%}
```

This is often necessary when the raw image size overflows the slide. However, in ioslides output, the image attributes are ignored.

frederik-elwert:
This fix captures image attributes in the `Image()` and `CaptionedImage()` functions and includes them in the `` tag. Fixes #1524.

### DIFF
--- a/inst/rmd/ioslides/ioslides_presentation.lua
+++ b/inst/rmd/ioslides/ioslides_presentation.lua
@@ -173,17 +173,17 @@ function Link(s, src, tit)
          escape(tit,true) .. "'>" .. s .. "</a>"
 end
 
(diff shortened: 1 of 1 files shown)


frederik-elwert:
Image attributes were added in pandoc 1.16. I also checked the sample.lua file from the 1.19 release, and it also contains image attributes. So that should not be a problem.

yihui:
Okay. Thanks a lot!

PR #1526, opened 2019-02-02
Remove unnecessary directory check

bwiernik:
In `reference_intermediates_generator`, this directory check precludes using `system.file` to refer to templates included in packages (as we do in psychmeta https://github.com/psychmeta/psychmeta/blob/2447d1891bf41d6fa3aa34bd88adee91d9cc8ed2/R/output.R#L810) or otherwise to files located somewhere else on the user's system. This is really limiting. This check isn't necessary, I don't believe—`file.copy` works fine with `reference_doc` located in another directory.


bwiernik:
I was wondering whether it might be possible to merge this PR? The current behavior breaks the Word output in psychmeta, and I would prefer to not write in a clunky hack to work around it.

PR #1530, opened 2019-02-08
Add RStudio Connect Reference to  Homepagee

slopp:
Based off of Google Analytics, rmarkdown.rstudio.com is one of our highest visited websites. This PR adds a band at the bottom of the homepage with a short reference to RStudio Connect that highlights some of the opportunities for using R Markdown in production: publishing, scheduling, parameterized reports, and email distribution. These features are not addressed elsewhere in the site, even though they are common use cases. The goal of this change is to help R Markdown users learn about the additional opportunities available to them without coming across as serving an ad.


slopp:
Thanks for the `optipng` tip @yihui !

PR #1550, opened 2019-03-14
Avoid changing margins with custom document class

### LINKED ISSUE #1549: Disable custom margins with document class specified
If one sets the margins without `geometry`, such as when using KOMA-Script, the custom Rmarkdown setting stops this from taking effect:

```markdown
---
documentclass: scrbook
classoption:
- DIV=12
---
```

At the moment, the margins are set to one inch unless there are custom settings for `geometry` in the YAML front matter:

<https://github.com/rstudio/rmarkdown/blob/cdf4ff0598bbae6c231451bad046e931a84c97aa/R/pdf_document.R#L173>

I would think that the user is very unlikely to want the margins set this way if using a document class other than `article`. Could I suggest that the custom margins be removed if either `geometry` or `documentclass` are set in the front matter?

adunning:
Detect `documentclass` alongside `geometry` in determining whether to change the margins: avoids conflicts when not using the `article` class. Closes #1549.

### DIFF
--- a/R/pdf_document.R
+++ b/R/pdf_document.R
@@ -171,7 +171,7 @@ pdf_document <- function(toc = FALSE,
     input_test <- read_utf8(input_file)
 
     # set the margin to 1 inch if no geometry options or document class specified
-    if (!has_yaml_parameter(input_test, "geometry") || !has_yaml_parameter(input_test, "documentclass"))
+    if (!has_yaml_parameter(input_test, "(geometry|documentclass)"))
       args <- c(args, "--variable", "geometry:margin=1in")
 
     # use titling package to change title format to be more compact by default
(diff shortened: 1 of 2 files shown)


R/pdf_document.R
```diff
@@ -170,8 +170,8 @@ pdf_document <- function(toc = FALSE,
 
     input_test <- read_utf8(input_file)
 
-    # set the margin to 1 inch if no other geometry options specified
-    if (!has_yaml_parameter(input_test, "geometry"))
+    # set the margin to 1 inch if no geometry options or document class specified
+    if (!has_yaml_parameter(input_test, "geometry") || !has_yaml_parameter(input_test, "documentclass"))
```
yihui: You can actually use a regular expression here:
```suggestion
    if (!has_yaml_parameter(input_test, "(geometry|documentclass)"))
```

adunning: I did not know that; thank you!

adunning:
Thanks for your guidance!

PR #1552, opened 2019-03-16
Preserve the toc_float text formatting

### LINKED ISSUE #1548: Enabling toc_float strips the text format
Enabling `toc_float` strips the text format such as *italic*, **blod** or superscript. 
```
---
output:
  html_document:
    toc: yes
    toc_float: yes
    toc_depth: 3
---

## test *italic*

## test **bold**

## test 10^6^
```
image

codetrainee:
This fixes #1548 to keep the `toc_float` text formatting.

Before applying the fix.

image

After applying the fix.

image

### DIFF
--- a/inst/rmd/h/tocify/jquery.tocify.js
+++ b/inst/rmd/h/tocify/jquery.tocify.js
@@ -386,13 +386,13 @@
 
           item.append($("<a/>", {
 
-            "text": self.text()
+            "html": self.html()
 
           }));
 
         } else {
 
-          item.text(self.text());
+          item.html(self.html());
 
         }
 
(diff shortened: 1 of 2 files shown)


inst/rmd/h/tocify/jquery.tocify.js
```diff
@@ -385,17 +385,20 @@
...
-
-            "text": self.text()
+            // Uses html to preserve the text formatting
+            "html": self.html()
 
           }));
 
         } else {
 
-          item.text(self.text());
```
yihui: Does `item.html(self.html())` work? Having a block element `<p>` inside `<li>` might introduce extra margin/padding issues.

codetrainee: Thanks for advice. `item.hmtl(self.html))` works. I've replaced the `<p>` tag.

PR #1563, opened 2019-04-04
Patch pandoc LaTeX template

### LINKED ISSUE #1510: Use new LaTeX templates from Pandoc
This is following on the discussion from #1509. The `default-1.17.0.2.tex` template is based is on the version at <https://github.com/jgm/pandoc-templates/blob/fea830fcd27b4df2a337ac774daa7d4bbca00866/default.latex>, which shows that the differences boil down to the following sections:

- <https://github.com/rstudio/rmarkdown/blob/c0b5945458aa1a99718ee174f9633ed65141f939/inst/rmd/latex/default-1.17.0.2.tex#L108>: added in <https://github.com/rstudio/rmarkdown/commit/bc2a49ed9da00fd5ec414bc192b112bbba34513d> for Pandoc 2.x compatibility, already in newer upstream templates
- <https://github.com/rstudio/rmarkdown/blob/c0b5945458aa1a99718ee174f9633ed65141f939/inst/rmd/latex/default-1.17.0.2.tex#L190>: I'm having trouble finding the reason for this, but if memory serves, this was a very old Pandoc bug in which it wasn't using `\thanks` for footnotes for `\maketitle`
- https://github.com/rstudio/rmarkdown/blob/c0b5945458aa1a99718ee174f9633ed65141f939/inst/rmd/latex/default-1.17.0.2.tex#L195

The last bit is the only significant difference in the repository's template. I think the intent of this was to fix the problem that Pandoc doesn't render `\subtitle` by default. This could be done more efficiently by adding the following via `header-includes`:

```latex
% Render \subtitle in \maketitle
\usepackage{titling}
\providecommand{\subtitle}[1]{%
  \posttitle{%
    \par\end{center}
    \begin{center}\large#1\end{center}
  }%
}
```

This should be fixed upstream; I'll take care of it.

adunning:
Patch the pandoc LaTeX template to include the document subtitle (unnecessary with pandoc 2.6 onwards) using `--include-in-header` rather than overwriting its built-in template, avoiding compability problems with newer versions of pandoc. Closes #1510.


R/pdf_document.R
```diff
@@ -149,9 +140,6 @@ pdf_document <- function(toc = FALSE,
   # content includes
   args <- c(args, includes_to_pandoc_args(includes))
 
-  # make sure the graphics package is always loaded
```
adunning: These lines should be unnecessary, as pandoc sets this internally.

adunning: Never mind; realized it's necessary for older pandoc versions.

mirh:
Needs rebasing I guess-ish?

PR #1578, opened 2019-05-03
Replace hand-written Rd files with roxygen-style comments (#1446)

### LINKED ISSUE #1446: Convert all docs to roxygen
As per slack discussion with @rich-iannone.

Finish converting all docs from hand-curated `.Rd` to roxygen.

Known holdovers (probably incomplete): compile_notebook.Rd, rmd_metadata.Rd, ioslides_presentation.Rd, render_site.Rd (but see #1445)

WerthPADOH:
Should close #1446.  All .Rd files not starting with `% Generated` are now generated by roxygen.
Only 1 new file is added: `R/rmarkdown.R`, which has the package's documentation.


R/ioslides_presentation.R
```diff
@@ -1,7 +1,266 @@
...
+#'
+#' @param logo Path to file that includes a logo for use in the presentation
+#'   (should be square and at least 128x128).
+#' @param slide_level Header level to consider as slide separator (Defaults to
+#'   header 2).
+#' @param incremental \code{TRUE} to render slide bullets incrementally.
+#'   Note that if you want to reverse the default incremental behavior for an
+#'   individual bullet you can preceded it with \code{>}.
+#'   For example: \emph{\code{> - Bullet Text}}.
+#' @param fig_width Default width (in inches) for figures.
```
yihui: I think it will be nice to inherit the docs from other output formats like `@inhertParams html_document`, so we don't have to repeat the documentation of these arguments.

WerthPADOH: This definitely showed where the `ioslides_presentation` docs hadn't kept up with `html_document`.

PR #1594, opened 2019-06-30
With an alternate MathJax URL and a custom template, Pandoc argument should be of the form '--mathjax=URL'

### LINKED ISSUE #1593: Trouble with an alternate MathJax URL and a custom template
I think the problem comes from **rmarkdown**.
I found a little bug in **rmarkdown** when a HTML document is rendered with a custom template and `mathjax: <some_url>`.
For now, here is a workaround for this bug. In https://github.com/topepo/MathJax-test/blob/master/_output.yml, you can use these arguments:
```yaml
bookdown::gitbook:
  css: FES.css
  dev: svglite
  split_by: section
  mathjax: null
  pandoc_args: --mathjax=https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/latest.js?config=TeX-MML-AM_HTMLorMML.js
```
I will open a PR in **rmarkdown** to fix this bug.

_Originally posted by @RLesur in https://github.com/rstudio/bookdown/issues/683#issuecomment-506997150_

RLesur:
This PR fixes #1593 and closes rstudio/bookdown#683. 

With HTML documents, when the user uses an alternate MathJax URL and a custom template, the Pandoc CLI argument generated by **rmarkdown** is `--mathjax URL`.
Therefore, the source code of the MathJax library is included in the final document.

In this case, the Pandoc argument should be of the form `--mathjax=URL`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,8 @@ rmarkdown 1.14
 
 - Added a **pkgdown** site for the **rmarkdown** package: https://rmarkdown.rstudio.com/docs/ (thanks, @apreshill, #1574).
 
+- Fixed the bug #1593: in HTML documents, when a MathJax URL is used with `self_contained = FALSE`, the source code of the MathJax library is included in the document. This bug was first declared in **bookdown** (thanks, @topepo for the bug report rstudio/bookdown#683, and @RLesur for the fix #1594).
+
 
 rmarkdown 1.13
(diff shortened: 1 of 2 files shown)


RLesur:
It seems that the Travis fail is related to the **pkgdown** package which uses an argument now deprecated.

PR #1596, opened 2019-07-04
Update stylesheet for code blocks with line-numbers

atusy:
This PR update stylesheet for code blocks with line-numbers.
Examples are at the bottom.

# Changes

## Background of `<div class="sourceCode">` is changed to transpreant.

In some syntax highlighting, background color of `<div class="sourceCode">` are specified (e.g., tango)
This cause ugly appearance especially in `theme: darkly`
As long as line numbers are not used, background color of the `div` tags are not visible.
Thus, I considered it is okay to letting them `transparent`.

## Background of  `pre` for chunk output are changed to white

Previously, specifying `class.output="numberLines"` caused output code block to have gray background where white is expected.
This is fixed by exact matching of class attributes.
Thus, white background are only applied when

- pre has no class (just like before)
- pre has the classes only related to line numbering on code blocks.

The latter is approached by the exact matching of classes.
Exact matching is required for the backward compatibility.
Otherwise, Bootstrap would not work to fill background colors for example by `class.output = "numberLines bg-warning"`.

# Note

If the output `pre` tag has the `lineAnchors` class, but not the `numberLines`, it is filled with gray.
Because line numbering is not working.

# Examples

## Before

image

## After

image

## Source

````
---
output: 
  html_document:
    highlight: tango
    theme: darkly
---

## Code blocks with line numbers

`class.source='numberLines'` and `class.output='numberLines'`

```{r, class.source='numberLines', class.output='numberLines'}
"hello"
```

`class.output='numberLines lineAnchors'`

```{r, class.source='numberLines', class.output='numberLines lineAnchors', echo = FALSE}
"hello"
```

`class.output='lineAnchors numberLines'`

```{r, class.source='numberLines', class.output='lineAnchors numberLines', echo = FALSE}
"hello"
```

`class.output='lineAnchors'`

```{r, class.source='numberLines', class.output='lineAnchors', echo = FALSE}
"hello"
```

## Bootstrap friendly

`class.output='numberLines bg-warning'` properly applies Bootstrap's `bg-warning` class to style background.

```{r class.output='numberLines bg-warning', echo = FALSE}
"hello"
```
````

### CODE REVIEW

**Comment by atusy** [comment]
> As long as line numbers are not used, background color of the `div` tags are not visible.

I just noticed the background color of the `div` is visible even in the line numbers are not used.
It appears at the outside of rounded corners of `pre` and making the corners as if squared.

**Comment by atusy** [comment]
My update CSS in the below looks too fat.

```css
  pre:not([class]),
  pre[class="numberLines"],
  pre[class="numberLines lineAnchors"],
  pre[class="numberLines line-anchors"],
  pre[class="number-lines"],
  pre[class="number-lines lineAnchors"],
  pre[class="number-lines line-anchors"],
  pre[class="lineAnchors numberLines"],
  pre[class="lineAnchors number-Lines"],
  pre[class="line-anchors numberLines"],
  pre[class="line-anchors number-lines"] {
    background-color: white;
  }
```

I feel I can remove the lines mixing up *camelCases* and *kebab-cases*, which is a bad manner and most users won't do it (I hope).

```css
  pre:not([class]),
  pre[class="numberLines"],
  pre[class="numberLines lineAnchors"],
  pre[class="lineAnchors numberLines"],
  pre[class="number-lines"],
  pre[class="number-lines line-anchors"],
  pre[class="line-anchors number-lines"] {
    background-color: white;
  }
```

What do you think?

atusy:
> Background of <div class="sourceCode"> is changed to transpreant.

This part is approached by #1598, and thus is reverted.

PR #1598, opened 2019-07-07
fix: apply pandoc div.sourceCode style to pre.sourceCode instead

atusy:
This PR is a generalization of a0c0c681f03049a8e017d6ca5295160698cc17a1 (#1471)

Currently, JS converts `background-color` and `color` properties of `div.sourceCode` to `pre.sourceCode` and delete the formers iff the both properties are specified for `div.sourceCode`

Thus, `highlight: tango` with `theme: darkly` clearly shows that `div.source {background: f8f8f8}` is active.
(`tango` only specified `background-color` not `color`)

```
div.sourceCode
  {  background-color: #f8f8f8; }
```

image

By this PR, output becomes

image

## Source

````
---
output:
  html_document:
    theme: darkly
    highlight: tango
---

Corners of code block looks like squared due to background colors of `div`.

```r
"Hello, Rmd"
```

Line numbers have background colors because it is within `div.sourceCode`

```{.numberLines}
"Hello, world"
```

Code blocks with no class attributes are unaffected

```
"Hello, world"
```
````

### DIFF
--- a/inst/rmd/h/default.html
+++ b/inst/rmd/h/default.html
@@ -78,7 +78,7 @@ $endif$
       if (rule.type !== rule.STYLE_RULE || rule.selectorText !== "div.sourceCode") continue;
       var style = rule.style.cssText;
       // check if color or background-color is set
-      if (rule.style.color === '' || rule.style.backgroundColor === '') continue;
+      if (rule.style.color === '' && rule.style.backgroundColor === '') continue;
       // replace div.sourceCode by a pre.sourceCode rule
       sheets[i].deleteRule(j);
       sheets[i].insertRule('pre.sourceCode{' + style + '}', j);

### CODE REVIEW

**Comment by atusy** [comment]
Note: `kate` is the only highlighting theme which does not work well with `theme` that has background colors.

image

This is because of following css embedded by pandoc.

```css
pre.numberSource code > span > a:first-child::before
  { content: counter(source-line);
    position: relative; left: -1em; text-align: right; vertical-align: baseline;
    border: none; display: inline-block;
    -webkit-touch-callout: none; -webkit-user-select: none;
    -khtml-user-select: none; -moz-user-select: none;
    -ms-user-select: none; user-select: none;
    padding: 0 4px; width: 4em;
    background-color: #ffffff;
    color: #a0a0a0;
  }
```

PR #1602, opened 2019-07-14
feat: individually select folding when code_folding: hide

### LINKED ISSUE #664: feature_request: code_folding for a specific chunk
I'm using a lot rmarkdown to construct labs for my student. I've seen the new code_folding option and I was wondering if this option could be specified/modified for some specific chunks. I'm considering using this option so that some hints/solutions are invisible until the students request them.
Thanks,
Erwan

atusy:
This PR enables code_folding for a specific chunk and closes #664.

This works by hiding all source codes as default except for sources that have 'fold-show' class.

- specifying `code_folding: hide` in YAML front matter
- specifying `class.source='fold-show'`

````
---
output:
  html_document:
    code_folding: hide
---

```{r class.source = 'fold-show'}
1
```

```{r class.source = 'fold-show'}
2
```

```{r}
# This source is hidden
```
````

If anyone wants to default to show all and hide specific ones, try

````
---
output:
  html_document:
    code_folding: hide
---

```{r}
knitr::opts_chunk$set(class.source='fold-show')
```

```{r}
'Shown by default'
```

```{r class.source = NULL}
'Hidden manually'
```
````

Maybe we can consider more for better class name?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -15,6 +15,7 @@ Authors@R: c(
   person("Richard", "Iannone", role = "aut", email = "rich@rstudio.com", comment = c(ORCID = "0000-0003-3925-190X")),
   #
   # Contributors, ordered alphabetically by first name
+  person("Atsushi", "Yasumoto", role = "ctb", comment = c(ORCID = "0000-0002-8335-495X")),
   person("Barret", "Schloerke", role = "ctb"),
   person("Christophe", "Dervieux", role = "ctb"),
   person("Frederik", "Aust", role = "ctb", email = "frederik.aust@uni-koeln.de", comment = c(ORCID = "0000-0003-4900-788X")),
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by atusy** [comment]
Thank you for approving!

**Comment by StoyoKaramihalev** [comment]
Thanks for making this happen! It would be quite useful to have a way to do the opposite too. That is, if code is shown by default to only hide a specific chunk, perhaps using a "fold-hide"? For data pipeline demo notebooks it's always nice to make the important code and plots visible, but I'd like to hide the code for making the plots.

**Comment by yihui** [comment]
I agree that will be useful, too. Another PR please? :)

**Comment by atusy** [comment]
Here you are!!
https://github.com/rstudio/rmarkdown/pull/1798

**Comment by StoyoKaramihalev** [comment]
That’s great, thanks so much! :)

**Comment by yihui** [comment]
Thanks @atusy! And @StoyoKaramihalev You can install the development version via:

```r
remotes::install_github('rstudio/rmarkdown')
```

PR #1626, opened 2019-08-17
add new feature: pagebreak for every format

### LINKED ISSUE #1610: add a pagebreak feature in pdf, word and html document
The idea came from a discussion in [rstudio.community](https://community.rstudio.com/t/starting-toc-on-a-new-page-after-title-page-in-rmd-to-word-doc-output/36081/9)

I think this is possible to easily add a feature where a pagebreak would be included in pdf, html or work document. It is already possible in pdf document with latex `\\newpage{}` but it also possible in html or word document using a custom output syntax. 

This feature is one available in [lua filters examples](https://github.com/pandoc/lua-filters/tree/master/pagebreak). Adding `\newpage` or `\pagebreak` anywhere in the rmarkdown document will be modified by lua filter to be a pagebreak in the output document.

That is why I try to implement it in rmarkdown by adding lua filter in the package. See this branch
[cderv/rmarkdown@pagebreak](https://github.com/rstudio/rmarkdown/compare/master...cderv:pagebreak)

I don't know if you would consider adding such a lua file in rmarkdown package. Also, the lua filter is a feature available in pandoc > 2.0. Otherwise, implementing this feature in R directly should also be possible I guess and no dependent of pandoc version. 

Please, tell me if you are interested in a PR or if this is out of scope for Rmarkdown. 
Thanks.

cderv:
This PR implements a new feature that will close #1610  and which is based on pandoc lua filter called pagebreak. 

With this PR, it is now possible to use `\newpage` or `\pagebreak` in any Rmarkdown and will create a pagebreak in the correct output syntax for `word_document()`, `html_document()` and `odt_document()`. For `pdf_document`, those are already working latex command. 
Currently, a vignette in this PR explains how it works. 

This specific pagebreak lua filter is not from me and was taken from https://github.com/pandoc/lua-filters/tree/master/pagebreak. The file can be used as is (with a PR I made today there). 
To use this file, the copyright notice is left in the file and I added the two authors for copyright in DESCRIPTION - I think this is how it should be done.

Also, lua filters are already used in pagedown but not in rmarkdown. This PR adds a small helper to include some lua filters with Rmarkdown. I think lua filter could be a really could way to add more features for several formats easily, and maybe also in other framework (I think of bookdown for example). I did not export the function and reuse a similar one from pagedown, adding a test for pandoc 2.0 as it works for 2+. 

This means this pagebreak feature only work with pandoc >= 2.0. With earlier version, nothing change.

As I said above, I documented in a vignette. I don't know if this is the best place. If this feature makes it to rmarkdown, a part in a future book could also be an idea to documented it (and maybe makes the documentation shorter in the package). 

Finally, I added some tests but only for `html` and `latex` output. I am not sure how to test word or odt easily. So basically, this test that the filter is correctly found and activated.

This PR is now functional. However, some things we could change on how it works: 

* Currently, the lua filter is always activated if pandoc >= 2.0, even if not used. This has no impact on the result but may have impact on rendering time for big documents. (just a guess)
I hesitated on making it optional.
* If you think it is better to make it optional, I see two ways:

   1. add an argument to activate the feature manually in each output format (like `toc`). It could be `pagebreak: false` by default and `pagebreak: true` to activate. As a user, I am not sure this is a good idea. If I wrote `\pagebreak` or `\newpage`, I would like it to work without doing nothing.
   2.  do like in `citeproc_required()` and detect if a keyword is used in the `input_lines`. This would be detecting `\newpage` or `\pagebreak` when output is html, word or odt, then activating the filter only if found.

* I added the filter also for `pdf_document()` but as it has no impact, I believe it could be remove from `pdf_document()` in rmarkdown, even if the lua filter can handle tex output.

I hope you'll find this as interesting as I am.


vignettes/pagebreak.Rmd
```diff
@@ -0,0 +1,153 @@
...
+
+rmarkdown will convert those commands in the correct output format syntax using a [lua filter](#lua-filter) during pandoc conversion. 
+
+# Using with PDF/ latex documents {#pdf}
+
+As the commands are the ones already used in latex syntax, this works as expected in a tex output document, and thus with pdf. Adding a pagebreak was already possible with rmarkdown when output is `pdf_document()` or `latex_document()`, without any restriction about the version of pandoc.
+
+# Using with HTML documents {#html}
+
+A `\newpage` or `\pagebreak` command in a rmarkdown document with output as HTML will be converted by default in this html code with inline style using CSS rule [`page-break-after`](https://www.w3schools.com/cssref/pr_print_pageba.asp)
```
RLesur: IMO, it would be better to link to an official documentation like MDN. W3Schools has many outdated pages (on this subject, see [this SO Meta discussion](https://meta.stackoverflow.com/questions/280478/why-not-w3schools-com)).
The MDN page is https://developer.mozilla.org/en-US/docs/Web/CSS/page-break-after and explain that we should use `break-after: page` now (obviously, `page-break-after` still works and I confess that I use it).

cderv: Thanks ! I'll change the link. 

Do you think we should change the lua filter to use `break-after: page` in inline style ?

RLesur: I think `break-after: page;` should be promoted while explaining that it replaces the older CSS property `page-break-after`.

cderv: Is this already well supported ? I am always unsure with this.
On https://caniuse.com/#search=page-break, it seems the new `break-*` alias is not well supported yet... 🤔

cderv:
Thanks ! I just need to see if we promote `break-*` over `page-break-*`. I am unsure but @RLesur is the css expert here... 

> BTW, this will definitely be a good topic for the R Markdown Cookbook.

I agree. I was waiting to know if it would be accepted. I wrote the vignette with the book in mind. I'll add it in the book: Lua Topic (dr-harper/rmarkdown-cookbook#170) + this new features as example.

vignettes/pagebreak.Rmd
RLesur: Okay, maybe the best solution is to mention both and pointing to caniuse (it is always up to date)

cderv:
@RLesur, do you find it better now ? 

I did not change the `page-break-after` rule used as default in the lua filter but I use `break-after` in the custom class example and add link to caniuse. 

If it is ok with you, then @yihui can merge.

RLesur:
@cderv this is perfect! Just another suggestion, maybe the vignette could mention https://github.com/crsh/rmdfiltr

cderv:
😮 I did not know that one ! Nice ! 
It does not have lot of filters right now but it could be the place to add more if not included as default feature inside rmarkdown.

We can definitely add mention to this in the vignette and in the book.

crsh:
I'd be more than happy to collect additional filters in `rmdfiltr`. PRs are always welcome.

cderv:
@yihui I have just added reference to the two packages known relating to lua filters as suggested by @RLesur. If still ok with you, you can now merge. ☺️

PR #1634, opened 2019-08-29
Enable specifying arbitrary yaml files as alternatives of _output.yml

### LINKED ISSUE #1613: Feature request: pass the name of the _output.yml to render
It would be convenient to pass the name of the _output.yml file to use as part of the render function. It would allows to easily switch from one output configuration to another.

Something similar to bookdown render that allows to pass the _bookdown.yml file name to use.

atusy:
This PR enables `render()` and YAML front matter to specify arbitrary yaml files as alternatives of `_output.yml` or `_output.yaml`, and closes #1613 .

`render()` supports it by the `output_yaml` argument and YAML front matter supports it by the `output_yaml` top level parameter.
There, multiple yaml files can be specified and the first existing one is used.
If `output_yaml` is specified both for render and YAML front matter, then `render` has the priority.
If none are found, then `_output.yml` or `_output.yaml` will be used if they exists.

### Prepare bar.yaml

```r
list(
  html_document = list(toc = TRUE)
) %>%
  yaml::as.yaml() %>%
  writeLines("bar.yml")
```

### Prepare test-yaml.Rmd

````
---
output: html_document
output_yaml:
  - foo.yml
  - bar.yml
---

# A
````

### Render Rmd

If simply rendered by

```r
render("test-yaml.Rmd")
```

then, `bar.yml` is used because `foo.yml` is missing.

Even if `output_yaml` is commented out from the YAML frontmatter, `bar.yml` can still be used via `render()`.

```r
render(
  "test-yaml.Rmd",
  output_yaml = "bar.yml"
)
```


NEWS.md
```diff
@@ -3,6 +3,7 @@ rmarkdown 1.16
 
 - It is now possible to add pagebreak in html, word, latex and odt documents using `\newpage` or `\pagebreak` command in a Rmd file. This is possible thanks to [_pandoc's pagebreak lua filter_](https://github.com/pandoc/lua-filters/tree/master/pagebreak). See `vignette("pagebreak", package = "rmarkdown")`
 
+- Output formats can be configured by arbitrary yaml files which used to be restricted to `_output.yml` or `_output.yaml`. They can be specified via the `output_yaml` argument of `render()` or the `output_yaml` top level parameter of YAML front matter, and the first existing one will be used. If `output_yaml` is specified both for `render()` and YAML front matter, then `render()` has the priority. If none are found, then `_output.yml` or `_output.yaml` will be used if they exists (thanks, @atusy, #1634).
```
yihui:
```suggestion
- Output formats can be configured by arbitrary YAML files, which used to be restricted to `_output.yml` or `_output.yaml`. They can be specified via the `output_yaml` argument of `render()` or the `output_yaml` top-level parameter of YAML front matter, and the first existing one will be used. If `output_yaml` is specified both for `render()` and YAML front matter, then `render()` has the priority. If none are found, then `_output.yml` or `_output.yaml` will be used if they exists (thanks, @atusy, #1634).
```

atusy:
Thanks a lot. Commited your suggestions and also fixed a grammar mistake.

atusy:
Oh I need to update docs as well.

atusy:
About the comments, I see lower case "yaml" in those not only edited by me.
Do you want me to fix them as well?

https://github.com/rstudio/rmarkdown/blob/9611ee4f989e456f073412a9e682dcac82348143/R/output_format.R#L380

PR #1645, opened 2019-09-12
Allow the user to remove default/base html dependencies

ARawles:
I've currently got a specific use case where I have a bunch of automated reports that will all be pushed to a website.
They all have a similar structure and therefore have many similar dependencies. As a result, instead of making many seperate HTML files that each have their own local references to the required libraries (that also get produced each time each file is rendered), I added the references that will be required when the files are live on the website within the RMarkdown files themselves.

This led to the issue of duplicated library references, where the head would contain references to the default libraries (jquery, navigation, etc.) that were included when the file was rendered, and the body containing the references that I added myself that pointed to the libraries on the website server.

To rectify this, I added a `include_base_dependencies` parameter to the `html_document()` function. When set to TRUE, the default libraries are added as usual. When set to FALSE, no extra dependencies are added in the head of the HTML file.

This required minimal changes to the `html_document()` and `html_document_base()` functions to only add the dependencies if `include_base_dependencies` is TRUE.

I understand that this use-case is quite specific and the changes that it required may not be conducive to the philisophy and approach of the package. But I thought I'd create the pull request anyway in case the ability to remove all dependencies from an HTML file is useful to anyone else.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: rmarkdown
 Type: Package
 Title: Dynamic Documents for R
-Version: 1.15.1
+Version: 1.15.1.9000
 Authors@R: c(
   person("JJ", "Allaire", role = "aut", email = "jj@rstudio.com"),
   person("Yihui", "Xie", role = c("aut", "cre"), email = "xie@yihui.name", comment = c(ORCID = "0000-0003-0645-5666")),
--- a/R/html_document.R
+++ b/R/html_document.R
@@ -78,6 +78,8 @@
 #'  default definition or R Markdown. See the \code{\link{rmarkdown_format}} for
 #'  additional details.
 #'@param pandoc_args Additional command line options to pass to pandoc
+#'@param include_base_dependencies Include the base dependencies added by
+#'rmarkdown (currently jquery, jqueryui, tocify, and navigation)
 #'@param extra_dependencies,... Additional function arguments to pass to the
 #'  base R Markdown HTML output formatter \code{\link{html_document_base}}
 #'@return R Markdown output format to pass to \code{\link{render}}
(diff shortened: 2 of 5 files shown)


cderv:
Hi, 
Just a comment and out of curiosity: did you try the `html_fragment()` format ? 
This is a very light format with no feature activated and then no dependencies added.
I am interested to know why it would not be suitable for this case. 🤔 

Basically, all the dependencies I think you are mentioning can be deactivated using arguments from `html_document`.

* `toc_float` controls jquery, jqueryui, tocify dependencies
* `theme` controls jquery & navigation dependencies
* `template` and `highlight` control highlightjs dependency
* `df_print` controls pagedtable dependency

`html_fragment` changes all these argument and also use a custom template. You can look at the code. 

I am just curious : Would that not be enough ?

ARawles:
I haven't actually tried with `html_fragment()`. I'll have a look tomorrow morning and report back.

ARawles:
I think using `html_fragment()` does the job. I'd say that it's a bit of a workaround because it means that I'm producing an HTML file with just a body, but I think that it's a much better approach than adding a whole new parameter to the `html_document()` function to acheive basically the same thing that `html_fragment()` produces but with <head> data.
Thanks for your help @cderv !

jjallaire:
If you do as @cderv suggested and set a bunch of `html_document()` parameters to `null` you will get this output:

```html
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml">
<head>
<meta charset="utf-8" />
<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
<meta name="generator" content="pandoc" />
<meta http-equiv="X-UA-Compatible" content="IE=EDGE" />
<meta name="viewport" content="width=device-width, initial-scale=1">
<title>Hello</title>
</head>
<body>
<p>Hello, World</p>
<!-- code folding -->
</body>
</html>
```

So pretty minimal aside from the stray cole folding comment. The code to produce this output was:

```md
---
title: Hello
output: 
  html_document:
    theme: null
    highlight: null
    df_print: "default"
    mathjax: null
    self_contained: false
---

Hello, World
```

PR #1653, opened 2019-09-23
Correct Pandoc citeproc binary on windows

### LINKED ISSUE #1651: pandoc_citeproc_convert does not work on windows
I think this issue is the cause of https://github.com/rstudio/distill/issues/103

``` r
rmarkdown::pandoc_citeproc_convert("biblio.bib")
#> Error in system(command, intern = TRUE): 'pandoc-citeproc' not found
```

I think this is because `pandoc_citeproc` does not handle windows binaries
https://github.com/rstudio/rmarkdown/blob/73d055257ecb32b92d703ede9c1b55615bcb0629/R/pandoc.R#L666-L674

`citeproc_path` will not exist on windows because of a missing `.exe` at the end. 

This issue is currently only affecting `distill` i think because this package use directly `pandoc_citeproc_convert` and `rmardown` only pass pandoc-citeproc path to pandoc itself, and it is working without the exe.

One fix is to handle windows binary by adding `.exe` to the path.

cderv:
This fixes #1651 and also shoud fix https://github.com/rstudio/distill/issues/103

`pandoc_citeproc()` will now look on windows for a `pandoc-citeproc.exe` in pandoc directory, or if not found, default to a local directory binary. 

`pandoc` works a bit differently but was also returning a path with the`.exe`. It was working because `pandoc()` was not checking for existence of the binary, and calling a binary without `.exe` works on windows. However, I also integrated the `.exe` extension. 

Is this clever or should we only change `pandoc_citeproc.exe` as the issue is currently only with `pandoc_citeproc_convert()` ? 

It is easy to revert commit if required.

I wanted to add tests, but rmarkdown as a few test only and not sure what is expected in this package. Locally, tests passes on my computer.

---
This change is [Reviewable](https://reviewable.io/reviews/rstudio/rmarkdown/1653)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 rmarkdown 1.16
 ================================================================================
 
+- Pandoc and Pandoc citeproc binaries are now found correctly on Windows. This fixes an issue with `pandoc_citeproc_convert()` (thanks @cderv, #1651).
+
 - Added `self_contained` argument to `html_vignette` to keep intermediate directory if `self_contained = FALSE` (thanks, @cderv, #1641)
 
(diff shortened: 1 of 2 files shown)


yihui:
Fixed via a262754956c217d8672a97a8b9cb5c71d5a34a91.

PR #1675, opened 2019-10-18
Fix overlapping content while printing on Firefox with a TOC.

emilio:
Since there's only one empty left-float, only one page gets pushed to the right.

This is a deeper issue which as far as I can tell is not defined in the
specification, see https://github.com/w3c/csswg-drafts/issues/4434 for an
example with multicol.

This fixes the print issue on Firefox by floating the content to the right,
which produces the desired layout and prevents overlapping the table of
contents.

### DIFF
--- a/inst/rmd/h/default.html
+++ b/inst/rmd/h/default.html
@@ -383,6 +383,8 @@ $endif$
 .toc-content {
   padding-left: 30px;
   padding-right: 40px;
+  /* For printing, see https://github.com/w3c/csswg-drafts/issues/4434 */
+  float: right;
 }
 
 div.main-container {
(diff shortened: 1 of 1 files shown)


tdsmith:
A description of the printing issue with some images is https://github.com/webcompat/web-bugs/issues/32569.

emilio:
> This fix is probably okay, but I'd like to put the rule inside @media print {}. Thank you!

Any good reason for that? Given it works both in print and screen (because the layout is two-columns, in a screen the direction in which the second column floats doesn't really matter).

But anyhow, sure.

emilio:
Done

emilio:
Np, thank you!

Issue #1691, opened 2019-11-10
Rmarkdown fails to render in pdf_document with last latex version


cderv:
I think this error is the cause of https://github.com/yihui/tinytex/issues/152

There were some bugs recently with `graphicx` and `grffile` in last latex version. This has been reported in pandoc and upstream. See https://github.com/jgm/pandoc/issues/5848#issuecomment-547246756. 

Changes has been made in pandoc templates now by removing package `grffile`. 
https://github.com/jgm/pandoc/commit/4d5fd9e2fe360e47fd5beab724c612ce29aa39ee

I wonder if we need to do the same with rmarkdown included template. 🤔 
There was also other modification in pandoc template since `pandoc 1.17` - it could be an opportunity to update it. What do you think ? 

Here is a reproductible example (if you have the environment correctly set up :) )
```` r
dir.create(tmp_dir <- tempfile())
xfun::write_utf8(c(
  '---',
  'title: "Untitled"',
  'output: pdf_document',
  '---',
  '',
  '## R Markdown',
  '',
  '```{r pressure, echo=FALSE}',
  'plot(pressure)',
  '```'
), file.path(tmp_dir, "test.Rmd"))
````

``` r
# not working
rmarkdown::render(file.path(tmp_dir, "test.Rmd"),
                  output_options = list(latex_engine = "pdflatex"))
#> processing file: test.Rmd
#> output file: test.knit.md
#> "C:/PROGRA~3/CHOCOL~1/bin/pandoc" +RTS -K512m -RTS test.utf8.md --to latex --from markdown+autolink_bare_uris+tex_math_single_backslash --output test.tex --template "C:\Users\chris\Documents\R\win-library\3.6\rmarkdown\rmd\latex\default-1.17.0.2.tex" --highlight-style tango --pdf-engine pdflatex --variable graphics=yes --lua-filter "C:/Users/chris/Documents/R/win-library/3.6/rmarkdown/rmd/lua/pagebreak.lua" --lua-filter "C:/Users/chris/Documents/R/win-library/3.6/rmarkdown/rmd/lua/latex-div.lua" --variable "geometry:margin=1in" --variable "compact-title:yes"
#> tlmgr search --file --global "/grffile.sty"
#> Warning in parse_packages(logfile, quiet = c(TRUE, FALSE, FALSE)): Failed
#> to find a package that contains grffile.sty
#> ! LaTeX Error: File `grffile.sty' not found.
#> 
#> ! Emergency stop.
[...]

PR #1692, opened 2019-11-11
fix for windows 10 xelatex no image render

smmurphy:
After a recent update I have found that png, jpg, etc. images no longer render with \includegraphics under xelatex on Windows 10 with the latest updates and formats to MikTex (pdf do render). Images all render correctly under pdflatex and luatex, but not using xelatex. Apparently the conflict is in the line `\usepackage{graphicx,grffile}` triggered by higher pandoc versions as opposed to `\usepackage{graphicx}` as the default. This pull request invokes `default.tex` ahead of other template considerations when latex_engine == 'xelatex'. This resolves the issues for my purposes and may be of value to other Windows users who experience the same problem. A more thoughtful integration of this change to the alternate pandoc version templates may be worthy of consideration.

Cross posting this here in case this is of value to others...

https://github.com/MiKTeX/miktex-packaging/issues/131#issuecomment-552168956

### DIFF
--- a/R/pdf_document.R
+++ b/R/pdf_document.R
@@ -114,6 +114,9 @@ pdf_document <- function(toc = FALSE,
     pandoc_available(error = TRUE)
     # choose the right template
     version <- pandoc_version()
+    if (latex_engine == "xelatex")
+      latex_template <- "default.tex"
+    else
     if (version >= "1.17.0.2")
       latex_template <- "default-1.17.0.2.tex"
     else if (version >= "1.15.2")


smmurphy:
I believe this pull request addresses issue #1691 as well.

smmurphy:
Thanks! Unfortunately it looks like the presence of,

```
% grffile has become a legacy package: https://ctan.org/pkg/grffile
\IfFileExists{grffile.sty}{%
\usepackage{grffile}
}{}
```
is till causing problems for me. When I render with xelatex png and jpg images do not appear. When I delete the above from the .tex the file renders as expected.

yihui:
I cannot reproduce the issue on my computer. The **grffile** package has been removed in TeX Live, and when I updated my TeX Live (TinyTeX), it was also removed on my computer. If you are using MiKTeX instead of TinyTeX, you may manually remove the **grffile** package.

smmurphy:
OK, thank you! Found it hiding in `oberdiek`. After removing that everything works as expected with your commit!

u-fischer:
grffile has not been removed from ctan. There was a packaging error in texlive which meant that it was missing for a short time, but it will reappear as a stubs which does nothing. But it is not in oberdiek anymore, so  @smmurphy if you have a grffile.sty in oberdiek then your oberdiek version is not up-to-date.

smmurphy:
Thanks for the clarification @u-fischer I did check for updates with MikTex, but did not try updating `oberdiek` explicitly.

davidcarlisle:
@smmurphy  grffile is not in the oberdiek bundle any more it is in its own grffile package on ctan.

Issue #1700, opened 2019-11-20
rmarkdown::run fails if file is NULL and dir is set


jenzopr:
Dear all,

I am trying to run `rmarkdown::run(file = NULL, dir = "/srv/shiny-server")` to start a shiny server to render an interactive document (lets say at `/srv/shiny-server/index.rmd`). 

The call fails with 
```
> rmarkdown::run(file = NULL, dir = '/srv/shiny-server')
Error in readLines(con, warn = FALSE) : cannot open the connection
In addition: Warning message:
In readLines(con, warn = FALSE) :
  cannot open file 'index.rmd': No such file or directory
```

I was able to track this down to [line 128](https://github.com/jenzopr/rmarkdown/blob/master/R/shiny.R#L128) of `R/shiny.R`, where `file` is `NULL` and `default_file` gets a value of `"index.rmd"`, but unfortunately without the correct directory prefix from `dir`. The following [line 129](https://github.com/jenzopr/rmarkdown/blob/master/R/shiny.R#L129) then tries to get the documents runtime from the `"index.rmd"`s YAML header, whereas the *true* file resides in `/srv/shiny-server/index.rmd` (aka `file.path(dir, default_file)`).

The content of `/srv/shiny-server/index.rmd` is:

```
---
title: testfile
runtime: shiny
---

## Testfile
``` 

My proposed fix would be either to change [line 128](https://github.com/jenzopr/rmarkdown/blob/master/R/shiny.R#L128) to

```
target_file <- file %||% file.path(dir, default_file)
```

of prefix `default_file` directly with `dir` in case `default_file` was `NULL` (i.e in lines 76, 81, 88 and 99) of `R/shiny.R`.

I can file a PR if you're okay with either option. Please let me know what you think.

Best,
Jens

---

By filing an issue to this repo, I promise that

- [X] I have fully read the issue guide at https://yihui.org/issue/.
- [X] I have provided the necessary information about my issue.
    - If I'm asking a question, I have already asked it on Stack Overflow or RStudio Community, waited for at least 24 hours, and included a link to my question there.
    - If I'm filing a bug report, I have included a minimal, self-contained, and reproducible example, and have also included `xfun::session_info('rmarkdown')`. I have upgraded all my packages to their latest versions (e.g., R, RStudio, and R packages), and also tried the development version: `remotes::install_github('rstudio/rmarkdown')`.
    - If I have posted the same issue elsewhere, I have also mentioned it in this issue.
[...]

PR #1703, opened 2019-11-21
Prevent rmarkdown::run from failing when file is NULL and dir is set

jenzopr:
Hi Yihui,

I prefixed all occurrences of `default_file` with values from `dir` in case it was `NULL`. This allows the user to give a full path for `default_file` as well. 

Xref: #1700 

Best,
Jens


jenzopr:
Thank you for the fast reply 👍

yihui:
FYI I just reverted this PR because of #1714 and used the alternative fix you mentioned in #1700. Thanks again!

jenzopr:
Oh, my bad then! Thanks for the note.

PR #1706, opened 2019-11-22
Add Bootstrap 4 support

### LINKED ISSUE #1260: Update to Bootstrap v4?
What is the road map to implementation of Bootstrap 4, since it has been fully released now.

I have tried replacing the Bootstrap version in "C:\Program Files\R\R~version\library\rmarkdown\rmd\h\bootstrap-3.3.5", but it has never worked in its entirety. Any suggested hacks for now?

cpsievert:
This PR allows the `theme` parameter of `html_document()` (and `html_document_base()`) to understand `bslib::bs_theme()` objects, meaning that you can opt-in to Bootstrap 4 this way (Closes #1260):

```yaml
html_document:
  theme: !expr bslib::bs_theme(version = 4)
```

Moreover, if `theme` is a list of values, those values are now implicitly supplied to `bs_theme()`, meaning this an equivalent way to opt-into Bootstrap 4. 

```yaml
html_document:
  theme: 
    version: 4
```

Note that, with `bs_theme()`, you can use any Bootswatch 4 (or 3) theme:

```yaml
html_document:
  theme: 
    version: 4
    bootswatch: minty
```

Or, implement your own custom theme (this one inspired by [Material's dark mode suggestions](https://material.io/design/color/dark-theme.html#usage)). 

```yaml
html_document:
  theme: 
      bg: "#121212"
      fg: "#E4E4E4"
      primary: "#EA80FC"
      secondary: "#00DAC6"
      base_font: !expr bslib::font_google("Open Sans")
      heading_font: !expr bslib::font_google("Proza Libre")
      code_font: !expr bslib::font_google("Fira Code")
      code-color: "#00DAC6"
```

This PR also adds integration with `bslib::bs_themer()` allowing `runtime: shiny` documents to be interactively themed.

````rmd
---
runtime: shiny
output:
  html_document:
    theme: 
      version: 4
---

```{r}
bslib::bs_themer()
```
````

Lastly, if you want to supply custom Sass rules that utilize the underlying `theme`'s Sass, you can now supply `.scss`/`.sass` files to the `css` parameter and those Sass files will be compiled with the `theme`. For example, say you want to define a new class that extends a group of BS4 utility classes to add padding and border around tabsets:

```scss
.with-border {
  @extend .p-3; @extend .border; @extend .border-top-0; @extend .rounded-bottom;
}
```

````rmd
---
output:
  html_document:
    theme: 
      version: 4
    css: with-border.scss
---

# Tabs! {.tabset}

## Summary {.with-border}

Here is a summary table.

```{r}
summary(cars)
```

## Plot {.with-border}

Here is a plot:

```{r echo=FALSE}
plot(pressure)
```
````

Screen Shot 2021-01-18 at 3 54 33 PM

### CODE REVIEW

File: DESCRIPTION [hunk 1]
```diff
@@ -98,11 +99,13 @@ Suggests:
...
+    callr (>= 2.0.0),
+    bootsass
 SystemRequirements: pandoc (>= 1.12.3) - http://pandoc.org
 URL: https://github.com/rstudio/rmarkdown
 BugReports: https://github.com/rstudio/rmarkdown/issues
 License: GPL-3
 RoxygenNote: 7.0.0
 Encoding: UTF-8
 VignetteBuilder: knitr
+Remotes: rstudio/bootsass@carson/bootswatch-rmd
```
jcheng5:
```suggestion
Remotes: rstudio/bootsass@carson/bootswatch-rmd
```

R/html_dependencies.R
```diff
@@ -38,24 +38,62 @@ html_dependency_jqueryui <- function() {
...
+          "js/bootstrap.min.js",
+          # These shims are necessary for IE 8 compatibility
+          "shim/html5shiv.min.js",
+          "shim/respond.min.js"
+        ),
+        stylesheet = paste0("css/", theme, ".min.css")))
+    )
+  }
+
+  if (grepl("^4", version)) {
```
jcheng5:
```suggestion
  if (grepl("^4\\b", version)) {
```

jcheng5:
```suggestion
  if (grepl("^4\\b", version)) {
```

R/html_dependencies.R
```diff
@@ -38,24 +38,62 @@ html_dependency_jqueryui <- function() {
...
+    theme <- if (is.character(theme)) bootsass::bs4_theme_bootswatch(theme) else theme
+
+    if (identical(version, "4-3")) {
+      theme <- bootsass::bs4_themes(theme, bootsass::bs4_theme_bs3compat())
+    }
+
+    # Default to a base font size of 14px (instead of 16px)
+    theme_vars <- bootsass::theme_variables()
+    on.exit(bootsass::theme_variables(theme_vars), add = TRUE)
+    theme_vars[["font-size-base"]] <- theme_vars[["font-size-base"]] %||% "0.875rem"
```
jcheng5: Implement this by using a theme with `$font-size-base: 0.875rem !default;`

R/html_document.R
```diff
@@ -509,26 +531,51 @@ mathjax_config <- function() {
...
+           simplex = 66.3,
+           sketchy = 58,
+           slate = 56.5,
+           solar = 56,
+           spacelab = 58,
+           superhero = 48,
+           united = 56,
+           yeti = 54.5,
+           60)
+  }
```
cpsievert: Put this information in the bootsass package

R/html_document.R
```diff
@@ -538,47 +585,77 @@ pandoc_body_padding_variable_args <- function(theme) {
...
 #'
 #' @param navbar Navbar definition
+#' @param bootstrap_version major version of Bootstrap.
+#' See \code{bootstrap_version} argument on [html_document].
 #' @param links List of navbar links
 #' @return Path to temporary file with navbar definition
 #' @keywords internal
 #' @export
-navbar_html <- function(navbar) {
+navbar_html <- function(navbar, bootstrap_version = c("3", "4", "4-3")) {
```
jcheng5: Add `bootstrap_version <- match.args(bootstrap_version)`

R/html_document.R
```diff
@@ -538,47 +585,77 @@ pandoc_body_padding_variable_args <- function(theme) {
...
-  navbar_html <- sprintf(template, navbar$type, navbar$title, left, right)
   as_tmpfile(navbar_html)
 }
 
 #' @keywords internal
 #' @name navbar_html
 #' @export
-navbar_links_html <- function(links) {
-  as.character(navbar_links_tags(links))
+navbar_links_html <- function(links, bootstrap_version = c("3", "4", "4-3")) {
```
jcheng5: Call `match.args`

R/html_document.R
```diff
@@ -597,20 +674,26 @@ navbar_links_tags <- function(links, depth = 0L) {
...
-                tags$a(
-                  href = "#", class = "dropdown-toggle",
-                  `data-toggle` = "dropdown", role = "button",
-                  `aria-expanded` = "false", link_text),
-                tags$ul(class = "dropdown-menu", role = "menu", submenuLinks)
+        submenuLinks <- navbar_links_tags(x$menu, depth = depth + 1L, version)
+
+        tags$li(
+          class = menu_class,
+          class = if (!bs3_nav) "nav-item",
```
jcheng5: Can you double check that dropdown menus work properly when `bootstrap_version: 4`?

R/html_dependencies.R
```diff
@@ -38,24 +38,62 @@ html_dependency_jqueryui <- function() {
...
 #' @export
-html_dependency_bootstrap <- function(theme) {
+html_dependency_bootstrap <- function(theme, version = c("3", "4", "4-3")) {
+
+  version <- bootstrap_version_normalize(version)
+  theme <- bootswatch_theme_normalize(theme, version)
+
+  if (version == 3) {
+    return(
+      list(htmlDependency(
```
jcheng5: Let's not return a `list` here, as doing so will break backward compatibility. However, `4`/`4-3` can return a list. Also consider modifying htmltools to accept nested lists of html dependencies.

inst/rmd/h/_navbar.html
```diff
@@ -1,21 +1,17 @@
-
-<div class="navbar navbar-%s  navbar-fixed-top" role="navigation">
+<div class="navbar navbar-expand-%s navbar-%s bg-%s fixed-top" role="navigation">
```
jcheng5: Double-check that both `4` and `4-3` look right

inst/rmd/h/default.html
```diff
@@ -110,6 +112,7 @@
 
 $if(theme)$
 <style type="text/css">
+$if(bs3)$
 h1 {
   font-size: 34px;
 }
```
jcheng5: Even if bs4, make sure h1.title is bigger than h1

inst/rmd/h/default.html
```diff
@@ -285,27 +300,30 @@
   var menuAnchor = $$('a[href="' + href + '"]');
 
   // mark it active
-  menuAnchor.parent().addClass('active');
+  menuAnchor.tab('show');
 
   // if it's got a parent navbar menu mark it active as well
   menuAnchor.closest('li.dropdown').addClass('active');
```
jcheng5: This line is probably not needed, I think `menuAnchor` takes care of it.

inst/rmd/h/default.html
```diff
@@ -469,15 +529,20 @@
 $endfor$
 
 $if(theme)$
-<div class="fluid-row" id="$idprefix$header">
+$if(bs3)$
+<div id="$idprefix$header">
+$else$
+<!-- bs4 removed margin-top on h1, h2, ..., h6 -->
+<div id="$idprefix$header" style="margin-top: 20px">
```
jcheng5: Should we add `margin-top` for all headings?? Or for `<section>`??

inst/rmd/h/default.html
```diff
@@ -585,7 +650,9 @@ <h4 class="date">$date$</h4>
 
 $$(document).ready(function () {
   $$('.tabset-dropdown > .nav-tabs > li').click(function () {
-    $$(this).parent().toggleClass('nav-tabs-open')
+    $$(this).parent().toggleClass('nav-tabs-open');
+    $$(this).siblings().removeClass("active");
+    $$(this).addClass("active");
```
jcheng5: Consider making the class name `active` something with less overloaded meanings, if we only care about making the arrow appear.

inst/rmd/h/navigation-1.1/tabsets.js
```diff
@@ -121,15 +128,29 @@ window.buildTabsets = function(tocID) {
...
+    // activate tab (note users can add an .active class)
+    var navLinks = window.BS3_NAV ? tabList.children('li') : tabList.find(".nav-link");
+    $(navLinks[activeTab]).tab("show");
+
+    // make sure active content is shown on load
     var active = $(tabContent.children('div.section')[activeTab]);
     active.addClass('active');
     if (fade)
-      active.addClass('in');
+      active.addClass(window.BS3_NAV ? 'in' : 'show');
```
jcheng5: Verify that this works with 4-3. And try `.in { @extend .show; }`

inst/rmd/h/default.html
cpsievert: It's currently needed when `bootstrap_version: 4`. That's because the navbar dropdown markup we generate in R is not exactly what BS4 wants/expects (namely it seems BS4 wants `a.dropdown-item` instead of `li a.dropdown-item` for the active class to be added properly to the `dropdown`), but even if we did generate the 'correct' markup, it seems to still won't work with `.dropdown-submenu`.

inst/rmd/h/navigation-1.1/tabsets.js
cpsievert: Oh, whoops, this isn't currently working with `4-3`, but `.in { @extend .show; }` does appear to fix it

cpsievert: The actual fading doesn't work for some reason though

cpsievert: PR started here https://github.com/rstudio/bootsass/pull/9

inst/rmd/h/default.html
cpsievert: Done in https://github.com/rstudio/rmarkdown/pull/1706/commits/989bc53cea0902a8faa9473eca216dec2a0b9ddf

R/html_document.R
cpsievert: Pretty sure they work (although, see https://github.com/rstudio/rmarkdown/pull/1706#discussion_r350966939), but it looks like `nav-item` should only be applied to non-dropdowns (https://github.com/rstudio/rmarkdown/pull/1706/commits/66cd4b48f4dcff4024474418f4b76e9f0d5f67f9)

inst/rmd/h/_navbar.html
cpsievert: Things look alright for `4`, but `4-3` has a couple small issues:

* Right menu goes off the page at a medium size
* Collapse menu appears to the left of the brand (in `3` and `4` it appears in the right hand corner)

navbar

cpsievert: Interestingly, issue 1 goes away if I get rid of https://github.com/rstudio/bootsass/blob/master/inst/bs3compat/_navbar_compat.scss#L21 on the `.navbar-right`

cpsievert: Proposed fix https://github.com/rstudio/bootsass/pull/10

inst/rmd/h/default.html
cpsievert: Done in https://github.com/rstudio/rmarkdown/pull/1706/commits/87f8b0f1ff294e07fcdbcebc64ba333c9393390e

inst/rmd/h/default.html
cpsievert: Done in https://github.com/rstudio/rmarkdown/pull/1706/commits/69161c178ca7a3f59870e4cbd68ef428a2985239

inst/rmd/h/default.html
```diff
@@ -130,6 +133,26 @@
...
 }
+$else$
+/*
+BS3 -> BS4 saw a removal of margin-top which is a pretty jarring change for Rmd
+https://github.com/twbs/bootstrap/blob/f17f882d/dist/css/bootstrap.css#L1219-L1250
+https://github.com/twbs/bootstrap/blob/4de4874e/scss/_reboot.scss#L82-L89 */
+h1, h2, h3 { margin-top: 20px; }
+h4, h5, h6 { margin-top: 10px; }
+
+/* Pandoc uses align attribute to align content but BS4 squashes that */
```
cpsievert: Maybe this should as a part of `bootstraplib::bs_sass()`? flexdashboard suffers from the same problem

inst/rmd/h/default.html
```diff
@@ -130,6 +133,26 @@
 h6 {
   font-size: 12px;
 }
+$else$
+/*
+BS3 -> BS4 saw a removal of margin-top which is a pretty jarring change for Rmd
+https://github.com/twbs/bootstrap/blob/f17f882d/dist/css/bootstrap.css#L1219-L1250
+https://github.com/twbs/bootstrap/blob/4de4874e/scss/_reboot.scss#L82-L89 */
+h1, h2, h3 { margin-top: 20px; }
+h4, h5, h6 { margin-top: 10px; }
```
cpsievert: Maybe this should come as part of `bootstraplib::bs_sass()` as well (at least when `version = "4-3"`)?

jcheng5:
I get a warning:

> Input to asJSON(keep_vec_names=TRUE) is a named vector. In a future version of jsonlite, this option will not be supported, and named vectors will be translated into arrays instead of objects. If you want JSON object output, please use a named list instead. See ?toJSON.

but only when both `bootstrap_version: 4-3` or `4`, and `runtime: shiny` are used:

```
---
title: "test"
output:
  html_document:
    bootstrap_version: 4-3
runtime: shiny
---
```

R/html_document.R
```diff
@@ -360,8 +378,22 @@ html_document <- function(toc = FALSE,
...
+          args <- c(args, pandoc_body_padding_variable_args(theme))
+        } else {
+          # The $navbar-height SASS var is set by the bootstraplib theme
+          padding <- sass::sass_file(pkg_file("rmd/h/scss/navbar-padding.scss"))
+          padding_css <- bootstraplib::bootstrap_sass(
+            padding, theme = paste0(theme, "@", bootstrap_version)
+          )
+          css_file <- tempfile(fileext = ".css")
+          write_utf8(padding_css, css_file)
+          args <- c(args, "--css", pandoc_path_arg(css_file))
```
cpsievert: This won't work as `css_file` is a full path to a temporary file...

everdark: Even for `self_contained: true` I still got the local full path hard-coded in the final rendered page,
so the published site won't be able to have the css at all.

erieza09za08:
اهلا وسهلا اصدقائي الاوفياءونريد نرجو منكم المساعدة اصدقائي اعذروني على هذا فانتم افضل الاعمال والاصول معكم

R/html_document.R
```diff
@@ -478,21 +511,25 @@ knitr_options_html <- function(fig_width,
...
-    "yeti")
+bs_theme_match <- function(theme, version) {
+  # Explicit NULL has special meaning (no bootstrap)
+  if (is.null(theme)) return(NULL)
+
+  if (isTRUE(theme %in% c("default", "bootstrap"))) {
+    return("bootstrap")
+  }
+
+  # Resolve Bootstrap 3 case without a bootstraplib dependency
```
cpsievert, line 522: Think about maybe avoiding duplication here

R/html_document_base.R
```diff
@@ -52,8 +52,16 @@ html_document_base <- function(theme = NULL,
 
   output_dir <- ""
 
-  # dummy pre_knit and post_knit functions so that merging of outputs works
-  pre_knit <- function(input, ...) {}
+  # If theme represents a bs_theme() object, then we coerce it into one and set
+  # it as the global theme pre-knit (so users can modify)
+  old_theme <- NA
```
cpsievert, line 57:
```suggestion
  old_theme <- NULL
```

inst/rmd/h/default.html
```diff
@@ -47,13 +47,6 @@
 
 $if(highlightjs)$
 <style type="text/css">code{white-space: pre;}</style>
-$if(theme)$
-<style type="text/css">
-  pre:not([class]) {
-    background-color: white;
-  }
-</style>
-$endif$
```
cpsievert, line 56: This is duplicated logic (it also appears on line 107). That logic should probably also be pulled out and sass-ified (so it respects the theme's `bg`)

inst/rmd/h/default.html
```diff
@@ -340,8 +330,42 @@
   padding: 10px;
   border-right: 1px solid #ddd;
 }
+$else$
```
cpsievert, line 333: Is this really needed?

DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: rmarkdown
 Type: Package
 Title: Dynamic Documents for R
-Version: 2.5.3
+Version: 2.6.0
```
cderv, line 4: Just a note that we don't forget to renumber correctly.
Will be 2.7.x I think

NEWS.md
```diff
@@ -1,6 +1,8 @@
 rmarkdown 2.6
 ================================================================================
 
+- `html_document` (and `html_document_base`)'s `theme` arguments gains the ability to accept a list of values, which are passed along to `bslib::bs_theme()`, making it much easier create custom themes with Bootstrap 4 or 3. To learn more, see <https://rstudio.github.io/bslib> (thanks, @cpsievert, #1260).
```
cderv, line 4: Just a note to not forget to move this in the right place.

R/html_dependencies.R
```diff
@@ -55,9 +60,59 @@ html_dependency_bootstrap <- function(theme) {
...
+    stylesheet = paste0("css/", theme, ".min.css"),
+    head = format(tags$style(HTML(
+      "h1 {font-size: 34px;}
+       h1.title {font-size: 38px;}
+       h2 {font-size: 30px;}
+       h3 {font-size: 24px;}
+       h4 {font-size: 18px;}
+       h5 {font-size: 16px;}
+       h6 {font-size: 12px;}"
+    )))
```
cpsievert, line 72: Probably not worth moving this from html template because shinytest records this

R/html_dependencies.R
```diff
@@ -55,9 +60,59 @@ html_dependency_bootstrap <- function(theme) {
...
+    theme$version %||% bslib::version_default()
+  } else {
+    substr(html_dependency_bootstrap("default")$version, 1, 1)
+  }
+}
+
+is_bs3_compatible <- function(theme) {
+  version <- theme_version(theme)
+  identical(version, "3") || grepl("\\+3$", version)
+}
```
cpsievert, line 113: Remove all uses of this

inst/rmd/h/navigation-1.1/tabsets.js
```diff
@@ -121,15 +128,24 @@ window.buildTabsets = function(tocID) {
       tab.detach().appendTo(tabContent);
     });
 
-    // set active tab
-    $(tabList.children('li')[activeTab]).addClass('active');
+    // activate tab (note users can add an .active class)
+    var navLinks = window.BS3_NAV ? tabList.children('li') : tabList.find(".nav-link");
+    $(navLinks[activeTab]).tab("show");
```
cpsievert, line 133: I think all the changes here except for maybe this change?

cpsievert:
@cderv @yihui ok, this is finally ready for review, but just FYI, I may want to make some minor cosmetic tweaks over the next 24 hours as I do testing.

yihui:
@cpsievert Sounds great! Thanks for the heads up! I'll take a look tomorrow.

R/html_dependencies.R
```diff
@@ -55,9 +65,60 @@ html_dependency_bootstrap <- function(theme) {
...
+
+theme_version <- function(theme) {
+  if (is_bs_theme(theme)) {
+    bslib::theme_version(theme)
+  } else if (is.list(theme)) {
+    theme$version %||% bslib::version_default()
+  } else {
+    substr(html_dependency_bootstrap("default")$version, 1, 1)
+  }
+}
```
cderv, line 119: This is a place that will error if **bslib** is not installed. I saw that when trying a very simple example with the default template Rmd. I added this in the header and with no **bslib** installed : 
````yaml
output: 
  html_document:
    theme:
      bootswatch: flatly
````
and got this 
```r
Error in loadNamespace(name) : there is no package called 'bslib'
Calls: <Anonymous> ... loadNamespace -> withRestarts -> withOneRestart -> doWithOneRestart
Exécution arrêtée
```
because `bslib::version_default()` is called without checks on the package existence. 

I think we should check and fail early if **bslib** features are asked for in `theme` but the package is not installed.

cderv (changes requested):
Thanks @cpsievert ! I have seen a few things (have a few questions). I let @yihui do a more thorough review. Here are two more general comments for discussions:

### About **bslib** possibly not installed 

**bslib** is in Suggest, that means it could not be installed when **rmarkdown** is installed. The only check I have seen is so that `is_bs_theme()` returns FALSE if not installed. It may not be enough to be kind to users so that they don't have a clear message about this for two reasons: 

1. This will only result to FALSE if not installed with no message or warning or error explaining that **bslib** needs to be installed for using specific argument in `theme:`
2. It will still error in some places were the check is not done. See one of the comments below. 

I think we should check and fail fast if **bslib** is not installed but `theme` is not a character. In `html_document` and `html_document_base`

### About refactoring of bootstrap dependency

You introduce a `bootstrap_dependencies()` only to handle the result of `html_dependency_bootstrap` (list or not). 

We could have modified `html_dependency_bootstrap()` output directly by changing the output to a one element list, so that we could continue to use it internally. But instead there is a new wrapper. 
I think this is because it could break other packages that uses this exported function. (but maybe a rev dep check would confirm that ?)

However, I think it will not be easy to use the exported `html_dependency_bootstrap()` without the helper `bootstrap_dependencies()` now: The output is either a list of html_dependency elements, or on html_dependency element. 

``` r
str(rmarkdown::html_dependency_bootstrap(bslib::bs_theme(bootswatch = "flatly")), 1)
#> List of 3
#>  $ :List of 9
#>   ..- attr(*, "class")= chr "html_dependency"
#>  $ :List of 10
#>   ..- attr(*, "class")= chr "html_dependency"
#>  $ :List of 9
#>   ..- attr(*, "class")= chr "html_dependency"
str(rmarkdown::html_dependency_bootstrap("flatly"), 1)
#> List of 10
#>  $ name      : chr "bootstrap"
#>  $ version   : chr "3.3.5"
#>  $ src       :List of 1
#>  $ meta      :List of 1
#>  $ script    : chr [1:3] "js/bootstrap.min.js" "shim/html5shiv.min.js" "shim/respond.min.js"
#>  $ stylesheet: chr "css/flatly.min.css"
#>  $ head      : chr "<style>h1 {font-size: 34px;}\n       h1.title {font-size: 38px;}\n       h2 {font-size: 30px;}\n       h3 {font"| __truncated__
#>  $ attachment: NULL
#>  $ package   : NULL
#>  $ all_files : logi TRUE
#>  - attr(*, "class")= chr "html_dependency"
```

Having a function that have a non stable output is always a risk and more complicated  for the user. the wrapper you've done is to insure this output stability.  

Also, as the doc of `?html_dependency_bootstrap` suggest, I was expecting passing a list of `bs_theme()` argument would work but it is not. 
``` r
str(rmarkdown::html_dependency_bootstrap(list(bootswatch = "flatly")), 1)
#> List of 10
#>  $ name      : chr "bootstrap"
#>  $ version   : chr "3.3.5"
#>  $ src       :List of 1
#>  $ meta      :List of 1
#>  $ script    : chr [1:3] "js/bootstrap.min.js" "shim/html5shiv.min.js" "shim/respond.min.js"
#>  $ stylesheet: chr "css/flatly.min.css"
#>  $ head      : chr "<style>h1 {font-size: 34px;}\n       h1.title {font-size: 38px;}\n       h2 {font-size: 30px;}\n       h3 {font"| __truncated__
#>  $ attachment: NULL
#>  $ package   : NULL
#>  $ all_files : logi TRUE
#>  - attr(*, "class")= chr "html_dependency"
```

Should this be supported or the doc be changed ? 

As this is for advanced usage, maybe it is not so important but I prefer to share my thoughts on this one before it is too late.
Also because, this exported function [is used](https://github.com/search?l=&p=1&q=html_dependency_bootstrap+extension%3AR&ref=advsearch&type=Code) to create extensions and making it work easily even with **bslib** support could help. After it will start to be used, it could be too late (or more difficult than now)

R/html_document.R
```diff
@@ -321,9 +326,23 @@ html_document <- function(toc = FALSE,
...
-  # additional css
-  for (css_file in css)
-    args <- c(args, "--css", pandoc_path_arg(css_file, backslash = FALSE))
+  # additional sass/css
+  for (f in css) {
+    if (is_bs_themeish(theme)) {
+      theme <- bslib::bs_add_rules(as_bs_theme(theme), sass::sass_file(f))
+      next
+    }
+    is_sass <- grepl("\\.s[a,c]ss$", f)
```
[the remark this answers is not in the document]
cpsievert: No, thanks -- just an oversight on my end

cpsievert:
Thanks for the great feedback @cderv, 7982cb6 should address the `Suggest`ion concerns. It also centralizes `theme` value checking and adds some new additional error messages if the `theme` value is something we don't support.

As for `html_dependency_bootstrap()`'s type safety -- I agree that ideally we should always return a list of dependencies, but I'm pretty concerned about the implications for existing code -- there are about [10-15 packages on CRAN](https://github.com/search?l=r&q=%22html_dependency_bootstrap%22+user%3Acran+language%3AR&ref=searchresults&type=Code&utf8=%E2%9C%93) that would need to be updated and probably lots of non-CRAN projects that would break. @yihui, do you feel this is a breaking change worth taking on?

cderv (approved):
That is what I thought about the breaking change. This is the type of things we would like to change (fix) but we can't really as it comes with a cost. 😞 

Last thing I feel is missing is that this PR won't help improve code coverage 😅 
Some of the helper could benefit some simple test. **rmarkdown** is not the package with the more unit test but I hope to improve that soon. So no need to rush, I can do it later. I believe you have tested this PR with a lot of examples throughout the year of work on this one. 

Thanks for the work on this, this is a great feature !

cpsievert:
Thanks! Keep in mind that I'll be giving our QA team a handful of examples to do manual testing with, and I'll likely also be adding some automated tests to https://github.com/rstudio/shinycoreci-apps

cderv:
Oh nice! I was not aware of (forgot 🙄 ) that! That is great !

R/html_document_base.R
```diff
@@ -66,26 +78,25 @@ html_document_base <- function(theme = NULL,
...
     # resolve and inject extras, including dependencies specified by the format
     # and dependencies specified by the user (via extra_dependencies)
     format_deps <- list()
     format_deps <- append(format_deps, html_dependency_header_attrs())
     if (!is.null(theme)) {
-      format_deps <- append(format_deps, list(html_dependency_jquery(),
-                                              html_dependency_bootstrap(theme)))
+      format_deps <- append(format_deps, list(html_dependency_jquery()))
+      format_deps <- append(format_deps, bootstrap_dependencies(
+        # If TRUE, a as_bs_theme(theme) has been set globally (so users may customize it)
```
yihui, line 93:
```suggestion
        # If TRUE, an as_bs_theme(theme) has been set globally (so users may customize it)
```

yihui (approved):
Thanks! It looks good to me overall! I just hope **bslib** could get on CRAN soon.

> do you feel this is a breaking change worth taking on?

No, let's just keep the current behavior of `html_dependency_bootstrap()`. We could export `bootstrap_dependencies()` if any other package authors want to use a function that always returns a list of dependencies. If we are going to export it, I'd probably rename it to something like `html_dependency_bstheme()` for consistency with other `html_dependency_*()` functions.

cpsievert:
> I just hope bslib could get on CRAN soon.

Don't we all 😅. 

> We could export bootstrap_dependencies() if any other package authors want to use a function that always returns a list of dependencies. 

It seems a better (near-term) option would be to label `html_dependency_bootstrap()` as superseded and point to `bslib::bs_theme_dependencies()`, but it seems fine to do nothing for now until the sass/bslib stack matures more. In the longer term, we could also consider completely replacing `html_dependency_bootstrap()` with `bslib::bs_theme_dependencies()` (and remove the Bootstrap asset files from rmarkdown)

cpsievert:
Also, are y'all totally against bumping to 2.7.0 to signify this fairly significant feature (I've been using 2.7.0 to signal messages in `{thematic}` and `{bslib}`)?

yihui:
> In the longer term, we could also consider completely replacing `html_dependency_bootstrap()` with `bslib::bs_theme_dependencies()` (and remove the Bootstrap asset files from rmarkdown)

That sounds good to me.

> Also, are y'all totally against bumping to 2.7.0 to signify this fairly significant feature

The next CRAN release will be v2.7. We bump `z` in `x.y.z` when we introduce changes, and bump `y` (or `x` for really significant changes) when making CRAN releases.

cpsievert:
Ah, ok, I changed the version to `2.6.6`, feel free to merge (I can't)

inst/rmd/h/default.html
```diff
@@ -501,16 +439,24 @@
 $if(toc_float)$
 
 <!-- setup 3col/9col grid for toc_float and main content  -->
-<div class="row-fluid">
+<div class="row">
```
cderv, line 442: @cpsievert was this change necessary for **bslib** and bootstrap 4 supports ? 

Asking because it changes some layout for some custom format based on `html_document()`. 
For example: https://github.com/Bioconductor/BiocStyle/issues/86 (see both doc examples paragraph width)

it is a small difference but I am wondering if this could impact other custom formats. So asking for your knowledge on this `row` vs `row-fluid`. 

Thanks.

cpsievert: `.row-fluid` is a Bootstrap 2 class that changed to `.row` in Bootstrap 3 https://getbootstrap.com/docs/3.4/migration/

cderv: Thanks. I understand now. 
This means that about the paragraph width change, this is because `html_document` was using an old class and no bootstrap CSS was used. Switching to BS3 `row-fluid` class makes this CSS rule to apply
```css
.row {
    margin-right: -15px;
    margin-left: -15px;
}
```
this lead to a change in style for all `html_document` output. 

A side effect of this PR. Glad I know about that now. I may document that in NEWS maybe.

PR #1707, opened 2019-11-27
Made `rmarkdown:::list_template_dirs()` a little bit faster

jooyoungseo:
* Replaced `for` loop with `lapply()`, which gives a little bit faster performance.

### DIFF
--- a/R/draft.R
+++ b/R/draft.R
@@ -141,13 +141,13 @@ list_template_dirs <- function() {
 
     # check to see if the package includes a template folder
     template_folder <- system.file("rmarkdown", "templates", package = pkg)
-    if (dir_exists(template_folder)) {
+    if (fs::dir_exists(template_folder)) {
 
       # it does; list each template directory within the template folder
       template_dirs <- list.dirs(path = template_folder, recursive = FALSE)
-      for (dir in template_dirs) {
+      lapply(template_dirs, function(dir) {
         cat(pkg, "|", dir, "\n", sep = "")
-      }
+      })
     }
   }
 }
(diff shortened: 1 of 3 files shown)


R/draft.R
```diff
@@ -141,13 +141,13 @@ list_template_dirs <- function() {
 
     # check to see if the package includes a template folder
     template_folder <- system.file("rmarkdown", "templates", package = pkg)
-    if (dir_exists(template_folder)) {
+    if (fs::dir_exists(template_folder)) {
```
yihui: Currently **fs** is only a `Suggests` dependency, and I don't want to make it a hard dependency. If you don't have a strong reason about why `fs::dir_exists` has to be used here, I'd not change it here and [I also hope that this kind of changes can be left out](https://yihui.org/en/2018/02/bite-sized-pull-requests/) because they are not essential.

jooyoungseo:
Thank you very much for walking me through each detail benchmark, @yihui!

The reason why I replaced the `for loop` with `lapply()` is because I thought it should be more efficient in any ways.

[Hadley Wickham's Advanced R/Looping patterns](http://adv-r.had.co.nz/Functionals.html) explains as follows:

> It’s useful to remember that there are three basic ways to loop over a vector:
> 
> 1. loop over the elements: for (x in xs)
> 2. loop over the numeric indices: for (i in seq_along(xs))
> 3. loop over the names: for (nm in names(xs))
> 
> The first form is usually not a good choice for a for loop because it leads to inefficient ways of saving output. This is slow because each time you extend the vector, R has to copy all of the existing elements.

But, I admit my ignorance! It was a great learning chance again!

yihui:
A `for` loop is not unconditionally worse than `lapply()`. The point made in "Advanced R" was that you should avoid using `for` loops to extend an object in each step of the loop (in the manner of `x <- c(x, new_value)`). The actual inefficient thing is to make a copy of an object in each step, not the `for` loop itself.

Thanks for the contribution anyway!

Issue #1359, opened 2018-05-28
Interplay of header-includes, extra_dependencies, and in_header in pdf_document()


crsh:
Hi,

I've noticed some unexpected interplay between the different methods of customizing the document preamble in PDF documents. `header-includes` and `includes: in_header` both work individually, but when I use both `header-includes` appears to be ignored:

~~~yaml
---
title: "Untitled"
output: 
  pdf_document:
    keep_tex: true
    includes:
      in_header: preamble.tex

header-includes:
  - \usepackage{booktabs}
---
~~~

`preamble.tex` could for example simply contain

~~~tex
\usepackage{tabulary}
~~~

However, once I include the additional `extra_dependencies` everything is included.

~~~yaml
---
title: "Untitled"
output: 
  pdf_document:
    keep_tex: true
    extra_dependencies:
      - tabularx
    includes:
      in_header: preamble.tex

header-includes:
  - \usepackage{booktabs}
---
~~~

Is this intended behavior? To me it would be useful and less confusing if these three approaches could be used independently of one another.

crsh:
Hi, sorry to bother again. In case this is not intended behavior, I'd be willing to contribute a pull request to fix this. Any pointers as to where this could originate would be appreciated.

mnazarov:
This is actually an issue upstream in `pandoc` - see https://github.com/jgm/pandoc/issues/3139 and links from there

dmurdoch:
One other observation:  if you include `extra_dependencies` in the `pdf_document` call, then inline R code in the `header-includes:` field is not executed.  For example, this works:

    ---
    output: 
      pdf_document:
        keep_tex: true
    header-includes: 
      - \usepackage{fancyhdr}
      - \pagestyle{fancy}
      - '`r paste0("\\fancyhead[CO,CE]{", params$figureno, "}")`'
    params:
      figureno: "Fig. 1-1"
    ---

but this doesn't:

    ---
    output: 
      pdf_document:
        keep_tex: true
        extra_dependencies: fancyhdr
    header-includes: 
      - \pagestyle{fancy}
      - '`r paste0("\\fancyhead[CO,CE]{", params$figureno, "}")`'
    params:
      figureno: "Fig. 1-1"
    ---

If you look at the generated `.tex` file in the second one, you can see that the R code was never executed, and the literal string including it was inserted into the header.

    This is some text.

pauljohn32:
[...]

PR #1710, opened 2019-11-30
knit metadata from yaml that could contains R code

### LINKED ISSUE #1709: R code inside header-includes is no longer evaluated
Since 1.18, R code inside `header-includes` is not evaluated, but outside of it is.

Reproducible example (succeeds with 1.17, fails with 1.18):
```
---
title: "Issue"
author: "mcol"
header-includes:
- \externaldocument{`r sprintf("file_%s", format(Sys.Date(), "%Y-%m-%d"))`}
date: "`r format(Sys.Date(), '%-d %B %Y')`"
output: pdf_document
---

Test
```

Error message:
```
! Undefined control sequence.
l.66 \externaldocument
                      {`r sprintf("file_%s", format(Sys.Date(), "%Y-%m-%d"))`}

Error: Failed to compile test.tex.
```

Session info:
```
R version 3.6.0 (2019-04-26)
Running under: Ubuntu 16.04.4 LTS

Locale:
  LC_CTYPE=en_GB.UTF-8       LC_NUMERIC=C
  LC_TIME=en_GB.UTF-8        LC_COLLATE=en_GB.UTF-8
  LC_MONETARY=en_GB.UTF-8    LC_MESSAGES=en_GB.UTF-8
  LC_PAPER=en_GB.UTF-8       LC_NAME=C
  LC_ADDRESS=C               LC_TELEPHONE=C
  LC_MEASUREMENT=en_GB.UTF-8 LC_IDENTIFICATION=C

Package version:
  base64enc_0.1.3 digest_0.6.23   evaluate_0.14   glue_1.3.1
  graphics_3.6.0  grDevices_3.6.0 highr_0.8       htmltools_0.4.0
  jsonlite_1.6    knitr_1.26      magrittr_1.5    markdown_1.1
  methods_3.6.0   mime_0.7        Rcpp_1.0.3      rlang_0.4.2
  rmarkdown_1.18  stats_3.6.0     stringi_1.4.3   stringr_1.4.0
  tinytex_0.17    tools_3.6.0     utils_3.6.0     xfun_0.11
  yaml_2.2.0

Pandoc version: 1.17.2
```
---

- [X] I have fully read the issue guide at https://yihui.org/issue/.
- [X] I have provided the necessary information about my issue.
- [X] I have learned the Github Markdown syntax, and formatted my issue correctly.

cderv:
Previously before #1359 `header-includes` was passed to pandoc through yaml header. If Rmd yaml header contains R code, md yaml header will not as it has been knit already. And `header-includes` is used in pandoc. 

The conflict between `in_header` and `header-include` is a pandoc issue really. The fix in https://github.com/rstudio/rmarkdown/commit/5499ec63dbf4b15eb8769e4a0c59843ac53abfd2 has the effect of using `--include-in-header` no matter what now. This means that if `header-includes` contains R code it will no more be knitted and pass as is to pandoc. That causes issue for people using it. 

@yihui This PR is a draft as it is more to illustrate the issue and a possible fix that closes #1709. It shows metadata could be knitted before being passed to pandoc. 

Here is way to do it by knitting the file content and replacing it. 
Another way would be to knit the header metadata before passing it to new `append_in_header`
```r
    args <- append_in_header(
         knitr::knit(text = metadata[["header-includes"]], quiet = TRUE)
   )
```

I think there is other places that uses information from the yaml _as-is_ and where inline r expression in yaml won't be taken into account. (for `includes` for example). This is also a draft as I am not sure if we want to support this everywhere or fix bugs as they come. 

This is WIP as this needs more thoughts and some special caseS handling. 

I wonder if metadata to pass to pandoc should be retrieve from knitted output rather Rmd file. Could be a way to handle all this.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: rmarkdown
 Type: Package
 Title: Dynamic Documents for R
-Version: 1.18.3
+Version: 1.18.4
 Authors@R: c(
   person("JJ", "Allaire", role = "aut", email = "jj@rstudio.com"),
   person("Yihui", "Xie", role = c("aut", "cre"), email = "xie@yihui.name", comment = c(ORCID = "0000-0003-0645-5666")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@ rmarkdown 1.19
 
 - For the output format `pdf_document()`, we no longer adjust the vertical spacing of the title area. This means there will be a larger spacing above the document title in PDF. If you prefers the old (smaller) spacing, please download https://github.com/rstudio/rmarkdown/blob/f6961af/inst/rmd/latex/compact-title.tex and include it to the preamble via the `includes` option of `pdf_document`. However, please note that this means you won't be able to have multiple authors in the `author` field of the YAML frontmatter, unless you use a custom LaTeX template. With the default LaTeX template, you will run into the error in #1716. Besides, the `compact-title` option in YAML is no longer supported.
 
+- R code in the `header-includes` field in the YAML frontmatter stopped working in the previous version of **rmarkdown**. The code should be evaluated before passing to Pandoc (thanks, @mcol #1709, @cderv #1710).
+
 - The `encoding` argument is no longer passed to the `intermediates_generator` of R Markdown output formats. The `intermediates_generator` function can only accept arguments `input_file` and `intermediates_dir` now (see `?rmarkdown::output_format`). This is a breaking change to developers. If you are an output format developer, you have to remove the `encoding` argument in your `intermediates_generator` if your output format uses this function.
 
 - The `encoding` argument is no longer passed to the `render` element of the site generator (see `?rmarkdown::render_site`).
--- a/R/pdf_document.R
+++ b/R/pdf_document.R
@@ -108,7 +108,11 @@ pdf_document <- function(toc = FALSE,
   # table of contents
   args <- c(args, pandoc_toc_args(toc, toc_depth))
 
-  append_in_header <- function(text, file = as_tmpfile(text)) {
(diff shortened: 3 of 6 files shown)


cderv:
From the different tests made in this draft PR, it proves that Rmd yaml header content can't be passed as-is to pandoc because they could contain inline R code. If Rmarkdown still want to continue supporting that, I see two solutions:

* Explicitly knit content before passing it to pandoc. This correspond to [this changes](https://github.com/rstudio/rmarkdown/pull/1710/files/a7647fe6419c04fd7c67633b434f8351a90eb15f)

* In rmarkdown, after knitting and during preprocessing phases, I think it could be better to use the yaml header content from `*.utf8.md` file which is the result of knitting and the input file from pandoc. In this file the yaml header will contain no inline R code as it already would have been process during the knitting phase. This correspond for this issue to [this changes](https://github.com/rstudio/rmarkdown/pull/1710/files/e99913e7970806da6191b3270c9ec495e3d82163)

This second point could be applied elsewhere I guess and it could be a way to support inline R code in yaml header corresponding to pandoc variable in [YAML metadata block](https://pandoc.org/MANUAL.html#extension-yaml_metadata_block) that Rmarkdown would pass directly at command line instead of letting pandoc deals with them through yaml block at parsing. 

@yihui I leave this PR in pause waiting for further decision. I hope it will be of help to see what is the better thing to do. If needed, I'll add some documentations and tests later based on what will be decided.

yihui:
I'm in favor of the second option. That is, update the object `yaml_front_matter` after `knitr::knit()` in the `render()` function. Thank you!

cderv:
> any R code in YAML should be evaluated before YAML is passed to Pandoc 

That is what I also think was the best thing to do here after my different tries. ☺️ 
Thanks !

PR #1715, opened 2019-12-02
add ext argument to md_document

atusy:
This PR generalizes `md_document` by adding ext argument, so that we can generate a wide variety of formats.
Below is an example YAML frontmatter to generate pdf with ConTeXt.

```yaml
---
output:
  md_document:
    variant: context
    ext: ".pdf"
---
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,7 +9,7 @@ rmarkdown 1.19
 
 - `render_site()` can render R scripts in addition to Rmd files if you set `autospin: true` in `_site.yml` (thanks, @zeehio, #1564).
 
-- Added `ext` argument to `md_document()`. Its default value is ".md". This argument is intended to be used together with `variant` argument (e.g., `variant = "context"` and `ext = ".pdf"`).
+- Added `ext` argument to `md_document()`. Its default value is ".md". This argument is intended to be used together with `variant` argument (e.g., `variant = "context"` and `ext = ".pdf"`) (thanks, @atusy, #1715).
 
 rmarkdown 1.18
 ================================================================================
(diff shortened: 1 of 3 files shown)


cderv:
@atusy does this mean for you `md_document` can be used with any output format ? 

If I may share some thoughts about that: 

It feels strange that this function designed for creating markdown output can be used for creating pdf too. From my understanding, `variant` for `md_document` seems to correspond more to [Pandoc Markdown Variants](https://pandoc.org/MANUAL.html#markdown-variants) and `context` is not one of them. It should be `pdf_engine` or `pandoc_to` to be consistent with pandoc MANUAL I guess. Using `md_document` as a generic output format is not the clearer. Rmarkdown works well with one function for one output type. Mixing things would mean maybe less easy to maintain in the future. 
 
I think I would go with a more specific format with correct pandoc args, pre-processor and post-processor. It can be more suited for supporting ConText. What do you think ?

💭 Just sharing thoughts with you on this - I find it interesting when people share their concerns with me regarding my suggestions, hope it is fine with you. Cheers.

atusy:
I'm happy to listen to your thoughts.
To be honest, I also have the strange feeling LOL.
Despite of it, I made this PR because we have `bookdown::markdown_document2`, which shares the context of "generalization".

Let me explain why I come up with this PR.

- Using `variant` for non-markdown output is sometimes useful.
For example, `variant: native` renders native Haskell, and might be helpful on writing Pandoc filters.
- Generalized `md_document` can be a workaround to output formats that Pandoc supports, but the rmarkdown package does not support yet.

Anyway, this PR is not urging.

PR #1717, opened 2019-12-03
Notify shiny when slidy slides change

schloerke:
Fixes rstudio/learnr#308

When slides not initially displayed are displayed, Shiny still believes they are hidden.  By attaching to slidy's add_observer, we can notify Shiny to re-check what is "shown".


schloerke:
Future reference note...

I'm guessing this fix still does not work if there's a (second or later) build stage that has shiny output as it would still be hidden at time of slide display.  (I do not think this is common.)  BUT it is better than before and hooks into the existing `slidy` API.  I do not know of a `next build occurred` event to listen to without changing `slidy`'s code.

PR #1727, opened 2019-12-08
add keep_native option (closes #1726)

### LINKED ISSUE #1726: Producing native Abstract Syntax Tree from and within an Rmd YAML
Currently, it does not seem like we have a convenient workaround to produce native `Abstract Syntax Tree (AST)` within an Rmd document.

For a markdown file, we use the following command in terminal:

```markup
pandoc -f markdown -s -t native -o output.native file.md [--lua-filter=lua_file.lua] ...
```

Of course, we could use the same command for any Rmd files; however, it would be so much convenient if we could had a relevant YAML option like `keep_tex` and `keep_md`.

What if we add a `keep_native` boolean option within an Rmd file YAML that can generate `[file_name].native` when set to true?

atusy:
This PR enables output formats to keep native haskell files.
Currently, this is a draft, which modifies least functions (`html_document`, `pdf_document`, `output_format`, and `render`).
If @yihui likes the idea, I would like to add the `keep_md` argument to other formats, and then update Rd files and NEWS.


atusy:
Tested with

```
---
title: "Untitled"
output: 
  html_document:
    keep_native: true
  pdf_document:
    keep_native: true
---

a
```

atusy:
I think keep_native is a too advanced option. I close this PR and will be back with some other alternatives some day.

PR #1732, opened 2019-12-10
Fix #1723 more generally as a part of html_document_base()

### LINKED ISSUE #1723: Tabsets don't work in pandoc 2.8
Another compatibility issue with pandoc 2.8+ (cf. https://github.com/rstudio/bookdown/issues/832) happens when using `.tabset`s. The difference is due to the following change: when running
```
pandoc -t html4 --section-divs <<< "# test {.tabset}"
```
gives in pandoc 2.7.3:
```
<div id="test" class="section level1 tabset">
<h1>test</h1>
</div>
```
and in pandoc 2.8:
```
<div id="test" class="section level1">
<h1 class="tabset">test</h1>
</div>
```
so the selector https://github.com/rstudio/rmarkdown/blob/07a0ca4c3b8a11cbaf3c8050fe2064b91de2f0b8/inst/rmd/h/navigation-1.1/tabsets.js#L136 fails. A solution is to change the above to
```js
var tabsets = $("div.section.tabset");
if (tabsets.length == 0)
  tabsets = $("div.section").has(">.tabset");
```
also https://github.com/rstudio/rmarkdown/blob/07a0ca4c3b8a11cbaf3c8050fe2064b91de2f0b8/inst/rmd/h/navigation-1.1/tabsets.js#L61 needs to be changed to 
```
var match = tabset.attr('class').match(/level(\d)+/);
```
but it is a bit more tricky with sticky tabsets, hopefully someone can take it up from here...

---
<details>
  <summary>minimal example:</summary>

```
---
title: test tabsets
output: html_document
---

# Tabs {.tabset}

## Number 1

Text 1

## Numer 2

Text 2
```
</details>

---

<details>
  <summary>session info</summary>

```
R version 3.6.1 (2019-07-05)
Running under: Ubuntu 18.04.3 LTS

Locale:
  LC_CTYPE=en_US.UTF-8          LC_NUMERIC=C                  LC_TIME=en_IE.UTF-8           LC_COLLATE=en_US.UTF-8       
  LC_MONETARY=en_IE.UTF-8       LC_MESSAGES=en_US.UTF-8       LC_PAPER=en_IE.UTF-8          LC_NAME=en_IE.UTF-8          
  LC_ADDRESS=en_IE.UTF-8        LC_TELEPHONE=en_IE.UTF-8      LC_MEASUREMENT=en_IE.UTF-8    LC_IDENTIFICATION=en_IE.UTF-8

Package version:
  base64enc_0.1.3  digest_0.6.21    evaluate_0.14    glue_1.3.1       graphics_3.6.1   grDevices_3.6.1  highr_0.8        htmltools_0.4.0 
  jsonlite_1.6     knitr_1.25       magrittr_1.5     markdown_1.1     methods_3.6.1    mime_0.7         Rcpp_1.0.2       rlang_0.4.0     
  rmarkdown_1.18.5 stats_3.6.1      stringi_1.4.3    stringr_1.4.0    tinytex_0.16     tools_3.6.1      utils_3.6.1      xfun_0.10       
  yaml_2.2.0      

Pandoc version: 2.8
```
</details>

---

cpsievert:
Note that flexdashboard, and potentially other formats that build on `html_document_base()` also currently suffer from #1723. Compared to https://github.com/rstudio/rmarkdown/commit/c6433aae9530d4f000dc21b073432500b539827a, this patch has the benefit of patching downstream dependencies that rely on the old pandoc `--section-div` behavior (not just the tabset navigation logic specific to rmarkdown).


inst/rmd/h/pandoc/elevate-section-attrs.js
```diff
@@ -0,0 +1,14 @@
+$(function() {
+  $("div[class*='section level']").each(function(i, el) {
+    var $el = $(el);
+    var headers = $el.children().filter(":header");
+    for (var h = 0; h < headers.length; h++) {
+      var header = headers[h];
+      var attrs = header.attributes;
+      for (var a = 0; a < attrs.length; a++) {
+        $el.attr(attrs[a].name, attrs[a].value);
+      }
```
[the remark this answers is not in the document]
cpsievert: Ok, yea that sounds right...Is it safe to say, with Pandoc < v2.8, that are sections headers *always* empty?

inst/rmd/h/pandoc/elevate-section-attrs.js
```diff
@@ -0,0 +1,14 @@
+$(function() {
+  $("div[class*='section level']").each(function(i, el) {
+    var $el = $(el);
+    var headers = $el.children().filter(":header");
+    for (var h = 0; h < headers.length; h++) {
+      var header = headers[h];
+      var attrs = header.attributes;
+      for (var a = 0; a < attrs.length; a++) {
+        $el.attr(attrs[a].name, attrs[a].value);
```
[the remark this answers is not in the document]
cpsievert: Ok yea, https://github.com/rstudio/rmarkdown/pull/1732/commits/2607fe1645b4ad4cce0b61ecb5b72af18e2f6263 effectively does this for class (via `addClass()`) but I'll make sure to do this for other attributes as well

inst/rmd/h/pandoc/elevate-section-attrs.js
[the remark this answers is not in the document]
yihui: Yes, I think so.

inst/rmd/h/pandoc/elevate-section-attrs.js
```diff
@@ -0,0 +1,16 @@
+$(function() {
+  $("div[class*='section level']").each(function(i, el) {
```
jcheng5: Is this different/better than `"div.section.level"`?

inst/rmd/h/pandoc/elevate-section-attrs.js
```diff
@@ -0,0 +1,16 @@
...
+    var $el = $(el);
+    var headers = $el.children().filter(":header");
```
jcheng5: All headers, or just the first child?

jcheng5:
Is this our permanent approach to solving this? Or are we going to adopt our position once the next Pandoc release changes behavior again (assuming jgm doesn't end up choosing the pre-2.8 behavior)?

inst/rmd/h/pandoc/elevate-section-attrs.js
cpsievert: Yea, `level` has a suffix reflecting the level of the section (e.g., `level1`, `level2`, etc).

This selector matches any div where `class` contains `'section level'`. I'm not 100% sure the ordering of those classes always appears to be case for the versions of pandoc we're applying this patch to

jcheng5: `div.section[class*='level']`?

inst/rmd/h/navigation-1.1/tabsets.js
```diff
@@ -132,15 +132,6 @@ window.buildTabsets = function(tocID) {
       tabset.rmarkdownStickyTabs();
   }
 
-  // Pandoc 2.8 adds attributes on headers instead of their parent divs, and we
-  // need to move the 'class' attribute from header to div (#1723)
-  $('.tabset').filter(':header').each(function(i, el) {
```
jcheng5: Would it have the same effect if we kept this implementation but changed `$('.tabset').filter(':header').each` to `$(':header').each` (or `$(':header:first-child').each`)?

inst/rmd/h/pandoc/elevate-section-attrs.js
cpsievert: Maybe it could just be the first...can you think of scenario where this matters @yihui?

inst/rmd/h/navigation-1.1/tabsets.js
cpsievert: Essentially, yes, but a few things to note:

* `$(':header')` is not very performant https://api.jquery.com/header-selector/
* We'd still need to add/remove other attributes besides `class` (to support flexdashboards `data-*` attributes)
* We'd still need to move this logic outside of tabsets.js and into a general dependency of `html_document_base()` (if we've found a relevant pandoc version)

inst/rmd/h/pandoc/elevate-section-attrs.js
yihui: Each section has one and only one direct child, so it shouldn't matter. To be safe, we can select the first child.

yihui:
> Is this our permanent approach to solving this? Or are we going to adopt our position once the next Pandoc release changes behavior again (assuming jgm doesn't end up choosing the pre-2.8 behavior)?

The next version of Pandoc will repeat attributes on headers and divs: https://github.com/jgm/pandoc/issues/5965. The approach in this PR should work with both v2.8 and the next release. jgm doesn't want to go back to the pre-2.8 behavior. Even if he does, the fix here is still harmless because there wouldn't be any attributes on headers.

inst/rmd/h/pandoc/elevate-section-attrs.js
cpsievert: Sort of related to this concern...should the selector be `div[class*='level'], section[class*='level']` so that we're also compatible with `html5`?

```shell
pandoc -t html5 --section-divs <<< "# Hello {.foo}" 
```

```html
<section id="hello" class="level1">
<h1 class="foo">Hello</h1>
</section>
```

inst/rmd/h/pandoc/elevate-section-attrs.js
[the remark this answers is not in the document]
RLesur: > Is it safe to say, with Pandoc < v2.8, that are sections header attributes are _always_ empty?

Header attributes are empty for Pandoc < v2.8 when `--section-divs` is used. Otherwise, headers can have attributes. With Pandoc 2.7.3:

```bash
pandoc -t html4 <<< "# test"
```
gives
```html
<h1 id="test">test</h1>
```

inst/rmd/h/pandoc/elevate-section-attrs.js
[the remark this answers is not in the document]
cpsievert: I decided just to overwrite the section attribute (if isn't class). Maybe we should accumulate strings, but I figured we'd keep it simple until we find a need for it.

inst/rmd/h/pandoc/elevate-section-attrs.js
yihui: We probably won't move to `html5` in the foreseeable future. It will break a lot of things (we'll have to bite the bullet one more time someday). Anyway, it is harmless to add the `section` selector here.

inst/rmd/h/pandoc/elevate-section-attrs.js
```diff
@@ -0,0 +1,21 @@
+$(function() {
+  $("div.section[class*='level'], section[class*='level']").each(function(i, el) {
+    var $section = $(el);
+    var headers = $section.children().filter(":header");
```
yihui: You can use the `.first()` method:

```suggestion
    var $header = $section.children().filter(":header").first();
```

so you won't need `$(headers[0])` later.

R/html_dependencies.R
```diff
@@ -314,3 +314,19 @@ html_dependency_rsiframe <- function() {
   )
 }
 
+# Pandoc 2.8 added attributes on headers instead of their parent divs, and we
+# need to move the 'class' attribute from header to div (#1723)
+html_dependency_elevate_section_attrs <- function() {
+  if (!rmarkdown::pandoc_available()) return(NULL)
```
yihui:
```suggestion
  if (!pandoc_available()) return(NULL)
```

inst/rmd/h/pandoc/elevate-section-attrs.js
cpsievert: Thanks, done in https://github.com/rstudio/rmarkdown/pull/1732/commits/63a3717c7c5a673d54ed0ce210863bd0b7ef47f9

cderv:
FWIW, pandoc 2.9 is already released
https://github.com/jgm/pandoc/releases/tag/2.9

RLesur:
I've just observed that adding this fix for any HTML output format is a little bit annoying because this systematically adds jQuery dependency for a rather small JS script.  
For instance, **pagedown** output formats now inherits jQuery and there is no option to avoid that.

yihui:
@RLesur That sounds annoying indeed. We should get rid of the jQuery dependency here. If @cpsievert doesn't have time for it, I'll do it myself. Thanks for the report!

cpsievert:
@yihui that'd be great, I have other priorities at the moment.

yihui:
@RLesur Done. I decided not to support 2.8 but only 2.9+ because 2.9 was released only 20 days after 2.8. Users who were in a hurry to upgrade Pandoc just need to upgrade again.

RLesur:
Thanks @yihui!

PR #1750, opened 2019-12-28
explicitly call print when df_print: default (fixes #1614)

### LINKED ISSUE #1614: rmarkdown::render issue when calling render interacitvely
I found an issue where dataframes are not printing when calling `rmarkdown::render` from a chunk. 

To reproduce, create an RMarkdown file `test.Rmd` with the following contents and open it in RStudio: 

````
---
title: "Render Issue"
author: "Mark Kirzon"
date: "8/2/2019"
output: html_document
---

```{r setup, include=FALSE}
knitr::opts_chunk$set(echo = TRUE, cache=FALSE)
library(tidyverse)   
```

This prints fine when the document is knit with the 'Knit' button or Ctrl+Shift+K
```{r}
a = data.frame(A = c(1:5), B = head(LETTERS, 5), stringsAsFactors = F)
print(a)
```

If this chunk is executed, the above dataframe is not displayed in the output html. I expected the behavior to be the same. 
```{r, eval=FALSE, include=FALSE}
rmarkdown::render("test.Rmd")
```
````

This is what it looks like when knit using the button/keyboard shortcut:
image

This is what it looks like when executing the block: 
image

I think this is a bug as I am unable to print a normal dataframe when executing a chunk with the `rmarkdown::render` call. Note that it works fine if the same `rmarkdown::render` is run from the command line, rather than executed as a chunk. 

### Why I need to be able to execute from a chunk 
I like to include a chunk at the end of my analyses that has the various parametrization configurations that I frequently need to loop over. I've found it's really convenient and organized when all parametrizations are stored with the analysis. Here's the template that I include at the end of my analyses, which loops over N parameter configurations and updates the N html files. 
````
```{r looped_render, include=FALSE, eval=FALSE}
# Run this chunk interactively to render this document with custom parameters, to a custom destination

# Configure Output and parameters
path_out      = "" # If empty, current directory
output_prefix = tools::file_path_sans_ext(basename(rstudioapi::getSourceEditorContext()$path)) # Output filename, needs to be non-empty
list_params = list(list()) # This is a list of lists. Leave as is if the document has no parameters

for (params_tmp in list_params) {
  print(params_tmp)
  
  output_fn = sprintf("%s%s.html", output_prefix, paste(names(params_tmp), unname(params_tmp), sep= "_", collapse = "&"))
  
  rmarkdown::render(
    input         = basename(rstudioapi::getSourceEditorContext()$path),
    output_file   = file.path(path_out, output_fn),
    params        = params_tmp)
}
```
````

### PR DESCRIPTION
**Pull request by atusy**
This PR fixes #1614 .
Though I do not see why, `knit_print.data.frame` must call `print` explicitly when interactively rendering Rmd files and `df_print` is default.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,7 +11,7 @@ rmarkdown 2.1
 
 - Removed the jQuery dependency in `html_document_base()` (#1723). To avoid bugs like #1723, Pandoc 2.8 users have to upgrade to Pandoc 2.9+.
 
-- Fixed data frames not being printend when `render()` is called interactively (thanks, @atusy, #1750)
+- Fixed data frames not being printed when `render()` is called interactively (thanks, @atusy, #1750)
 
 
 rmarkdown 2.0
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,8 @@ rmarkdown 2.1
 
 - Removed the jQuery dependency in `html_document_base()` (#1723). To avoid bugs like #1723, Pandoc 2.8 users have to upgrade to Pandoc 2.9+.
 
+- Fixed data frames not being printend when `render()` is called interactively (thanks, @atusy, #1750)
+
 
 rmarkdown 2.0
 ================================================================================
--- a/R/knit_print.R
+++ b/R/knit_print.R
@@ -21,6 +21,8 @@ knit_print.data.frame <- function(x, ...) {
     if (identical(context$df_print, knitr::kable)) {
       res <- one_string(c('<div class="kable-table">', '', knitr::kable(x), '', '</div>'))
       knitr::asis_output(res)
+    } else if (identical(context$df_print, print)) {
+      print(x)
     } else {
       context$df_print(x)
     }

### CODE REVIEW

File: NEWS.md [hunk 1]
```diff
@@ -11,6 +11,8 @@ rmarkdown 2.1
 
 - Removed the jQuery dependency in `html_document_base()` (#1723). To avoid bugs like #1723, Pandoc 2.8 users have to upgrade to Pandoc 2.9+.
 
+- Fixed data frames not being printend when `render()` is called interactively (thanks, @atusy, #1750)
```
cderv:
```suggestion
- Fixed data frames not being printed when `render()` is called interactively (thanks, @atusy, #1750)
```

R/knit_print.R
```diff
@@ -21,6 +21,8 @@ knit_print.data.frame <- function(x, ...) {
     if (identical(context$df_print, knitr::kable)) {
       res <- one_string(c('<div class="kable-table">', '', knitr::kable(x), '', '</div>'))
       knitr::asis_output(res)
+    } else if (identical(context$df_print, print)) {
```
cderv: @atusy, I understand this as if `context$df_print` is identical to `print` call `print(x)` else call `context$df_print(x)`. but in this case isn't it the same ?  `context$df_print(x)` will be the same as `print(x)`. What am I missing ?

atusy: It should be the same, but happened to be different.
Anyway, I'll check the way to reproduce the issue.
Thanks!!

atusy:
I don't have time these days...
I close this PR for now.

PR #1751, opened 2019-12-28
Using knit field in YAML when rendering (closes #1711)

### LINKED ISSUE #1711: Feature request: using knit field in YAML when rendering
As a non-RStudio user, I was wondering if there would be any way for `rmarkdown::render()` detects a value for `knit` field in YAML, and passes it to the specified render, if any.

For example, I encounter the following line in YAML from time to time:

```yaml
knit: reprex::reprex_render
```

```yaml
knit: bookdown::render_book
```

It would be so nice if `rmarkdown::render()` could automatically decide on which rendering process that needs to be used according to `knit` field.

atusy:
This PR implements a requested feature in #1711 .
In order to let `render` use function defined at `knit` field, the `customize` parameter must be `TRUE`.
The default value is set to `interactive()` so to avoid the RStudio's knit button from evaluating the `knit` field repeatedly when `knit` field calls `render` again (e.g., when `knit: rmarkdown::render` is specified).

### DIFF
--- a/R/render.R
+++ b/R/render.R
@@ -211,7 +211,8 @@ NULL
 #' @param quiet An option to suppress printing of the pandoc command line.
 #' @param encoding Ignored. The encoding is always assumed to be UTF-8.
 #' @param customize Render input by a function defined by the \code{knit}
-#' field at the top level of the YAML front matter of the input.
+#' field at the top level of the YAML front matter of the input. Default to
+#' \code{FALSE}.
 #' @return
 #'   When \code{run_pandoc = TRUE}, the compiled document is written into
 #'   the output file, and the path of the output file is returned. When
@@ -252,7 +253,7 @@ render <- function(input,
                    run_pandoc = TRUE,
                    quiet = FALSE,
                    encoding = "UTF-8",
-                   customize = interactive()) {
+                   customize = FALSE) {
 
   yaml_knit <- yaml_front_matter(input)[["knit"]]
   if (!is.null(yaml_knit)) {
--- a/R/render.R
+++ b/R/render.R
@@ -212,7 +212,8 @@ NULL
 #' @param encoding Ignored. The encoding is always assumed to be UTF-8.
 #' @param customize Render input by a function defined by the \code{knit}
 #' field at the top level of the YAML front matter of the input. Default to
-#' \code{FALSE}.
+#' \code{FALSE}. If \code{TRUE}, then only the \code{input} argument is passed
+#' to the rendering function, and the others are ignored.
 #' @return
 #'   When \code{run_pandoc = TRUE}, the compiled document is written into
 #'   the output file, and the path of the output file is returned. When
@@ -257,16 +258,7 @@ render <- function(input,
 
   if (customize) {
     yaml_knit <- yaml_front_matter(input)[["knit"]]
-    if (!is.null(yaml_knit)) {
-      return(eval(parse(text = yaml_knit))(
-        input = input, output_file = output_file, output_dir = output_dir,
-        output_options = output_options, output_yaml = output_yaml,
-        intermediates_dir = intermediates_dir, knit_root_dir = knit_root_dir,
-        runtime = runtime, clean = clean, knit_meta = knit_meta, envir = envir,
-        run_pandoc = run_pandoc, quiet = quiet, encoding = encoding,
-        customize = FALSE
-      ))
-    }
+    if (!is.null(yaml_knit)) return(eval(parse(text = yaml_knit))(input))
   }
 
   perf_timer_start("render")
--- a/R/render.R
+++ b/R/render.R
@@ -210,6 +210,8 @@ NULL
 #' output.
 #' @param quiet An option to suppress printing of the pandoc command line.
 #' @param encoding Ignored. The encoding is always assumed to be UTF-8.
+#' @param customize Render input by a function defined by the \code{knit}
+#' field at the top level of the YAML front matter of the input.
 #' @return
 #'   When \code{run_pandoc = TRUE}, the compiled document is written into
 #'   the output file, and the path of the output file is returned. When
@@ -249,7 +251,20 @@ render <- function(input,
                    envir = parent.frame(),
                    run_pandoc = TRUE,
                    quiet = FALSE,
-                   encoding = "UTF-8") {
+                   encoding = "UTF-8",
+                   customize = interactive()) {
+
+  yaml_knit <- yaml_front_matter(input)[["knit"]]
+  if (!is.null(yaml_knit)) {
+    return(eval(parse(text = yaml_knit))(
+      input = input, output_file = output_file, output_dir = output_dir,
+      output_options = output_options, output_yaml = output_yaml,
+      intermediates_dir = intermediates_dir, knit_root_dir = knit_root_dir,
+      runtime = runtime, clean = clean, knit_meta = knit_meta, envir = envir,
+      run_pandoc = run_pandoc, quiet = quiet, encoding = encoding,
+      customize = FALSE
+    ))
+  }
 
   perf_timer_start("render")
 
--- a/R/render.R
+++ b/R/render.R
@@ -255,16 +255,18 @@ render <- function(input,
                    encoding = "UTF-8",
                    customize = FALSE) {
 
-  yaml_knit <- yaml_front_matter(input)[["knit"]]
-  if (!is.null(yaml_knit)) {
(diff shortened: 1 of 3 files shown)


atusy:
I guess `FALSE` is safer than `intearctive()` as the default value of `customize` in terms of backward compatibility.

R/render.R
```diff
@@ -249,7 +252,20 @@ render <- function(input,
                    envir = parent.frame(),
                    run_pandoc = TRUE,
                    quiet = FALSE,
-                   encoding = "UTF-8") {
+                   encoding = "UTF-8",
+                   customize = FALSE) {
```
cderv: what does the `customize` argument is used for ? It does not seem to be used... 🤔

R/render.R
```diff
@@ -249,7 +252,20 @@ render <- function(input,
...
+  yaml_knit <- yaml_front_matter(input)[["knit"]]
+  if (!is.null(yaml_knit)) {
+    return(eval(parse(text = yaml_knit))(
+      input = input, output_file = output_file, output_dir = output_dir,
+      output_options = output_options, output_yaml = output_yaml,
+      intermediates_dir = intermediates_dir, knit_root_dir = knit_root_dir,
+      runtime = runtime, clean = clean, knit_meta = knit_meta, envir = envir,
+      run_pandoc = run_pandoc, quiet = quiet, encoding = encoding,
+      customize = FALSE
+    ))
```
cderv: I think this will create some errors. I don't think we can assume that any function passed through the `knit` yaml option is a function that will take all these arguments ! 

First try was with this small example with new reprex feature
````r
---
output: reprex::reprex_document
knit: reprex::reprex_render
---

Describe you issue very briefly here. Then show it with a minimal, self-contained example in the following R chunk.

```{r}
x <- 1
y <- "2"
x + y
```
````
and I got this error 
```
> render("test.Rmd")
Error in eval(parse(text = yaml_knit))(input = input, output_file = output_file,  : 
  unused arguments (output_file = output_file, output_dir = output_dir, output_options = output_options, output_yaml = output_yaml, intermediates_dir = intermediates_dir, knit_root_dir = knit_root_dir, runtime = runtime, clean = clean, knit_meta = knit_meta, envir = envir, run_pandoc = run_pandoc, quiet = quiet, customize = FALSE)
```` 
as expected. 

I think it will break other use of this knit field when rendering without the knit button with `rmarkdown::render`. like some usage in the wild for [slidify](https://github.com/ramnathv/slidify/commit/07794a3c2d228822c173c1ad5eacf5ab54a42062). I tested and got same type of error as expected. 

Another test is trying to render the bookdown book which have this [knit field](https://github.com/rstudio/bookdown/blob/70f9c07b2049f772b377956047421172069193d9/inst/examples/index.Rmd#L5). I am not able to render the book using `render("index.Rmd")`. I get a weird error about a stack pile. `render` could be called recursively maybe 🤔 
But still it does not work.

File: R/render.R (hunk 1 above)
Comment (atusy) [reply R/render.R]: Oops that is my mistake. I'll fix it.

File: R/render.R (hunk 2 above)
Comment (atusy) [reply R/render.R]: Thanks for the review!
At first I thought all arguments should be passed, but now am thinking that we only have to pass input.
If then, the code becomes simple, and the behavior becomes much closer to what RStudio's knit button does.
What do you think?

File: R/render.R (hunk 1 above)
Comment (atusy) [reply R/render.R]: fixed via 0e8d7b0

File: R/render.R (hunk 2 above)
Comment (atusy) [reply R/render.R]: Updated via 6c92b6f so that only the input argument is passed.

**Comment by atusy** [comment]
Note that `customize` parameter must be `FALSE` in the `knit` field.
Otherwise, `render` will customize itself repeatedly.

```
---
knit: function(input) render(input, customize = TRUE)
---
```

```
render("test.Rmd", customize = TRUE)
#> Loading rmarkdown
#> Error: node stack overflow
#> Error during wrapup: node stack overflow
```

File: R/render.R (hunk 2 above)
Comment (cderv) [reply R/render.R]: > At first I thought all arguments should be passed, but now am thinking that we only have to pass input.
There need at least to be two fields. That is what the RStudio IDE expect: first argument is the input file name then a encoding field. Like in the [slidify](https://github.com/ramnathv/slidify/commit/07794a3c2d228822c173c1ad5eacf5ab54a42062) example.

**Comment by cderv** [comment]
After digging through this, I am not sure `rmarkdown::render` is the place to recreate what the RStudio knitr button does. In fact, one clicking on this knit button, if a `knit` field is define with a function or a function name, this function is used, otherwise `rmarkdown::render` is used, unless it is shiny runtime then `rmarkdown::run` is used. 

Why not add this behavior is another function ? This way no special argument like `customize` is required. 
It would be something like this 
```r
custom_render <- function(input_file, encoding = "UTF-8", ...) {
  yaml_header <- rmarkdown::yaml_front_matter(input_file)
  render_fun <- yaml_header[["knit"]]
  if (!is.null(render_fun)) {
    if (is.character(render_fun)) {
      render <- eval(parse(text = render_fun))
    } else if (is.function(render_fun)) {
      render <- render_fun
    } else {
      stop("knit field in yaml should be a function or a function name", call. = FALSE)
    }
  } else {
    render <- rmarkdown::render
  }
  render(input_file, encoding = encoding, ...)
}
```

Seem less impactful for breaking change than modifying `rmarkdown::render`

cderv:
Not sure you can use this with RStudio button 
```yaml
knit: !expr reprex::reprex_render
````
SImpler function : 

```r
custom_render <- function(input_file, encoding = "UTF-8", ...) {
  yaml_header <- rmarkdown::yaml_front_matter(input_file)
  render_fun <- yaml_header[["knit"]]
  if (!is.null(render_fun) && is.character(render_fun)) {
    render <- eval(parse(text = render_fun))
  } else {
    render <- rmarkdown::render
  }
  render(input_file, encoding = encoding, ...)
}
```

jooyoungseo:
@cderv, I don't think `encoding` argument should work every time.

For instance, `xaringan::inf_mr()` does not have `encoding` argument so it would cause an error if user has defined `knit: xaringan::inf_mr` in their YAML. Furthermore, `xaringan::inf_mr()` does have different name for the `input` argument; it uses `moon`.

That's why I have implemented the folowing code:

``` r
render_all <- function(input, output_format = "all", open = TRUE, ...) {
  render_method <- rmarkdown::yaml_front_matter(input = input)$knit
  render_method <- ifelse(!is.null(render_method), lazyeval::lazy_eval(render_method), rmarkdown::render)
  render_method_args <- methods::formalArgs(render_method)
  first_arg <- render_method_args[1]

  render_args <- list()
  render_args[[first_arg]] <- input
  if ("output_format" %in% render_method_args) {
    render_args[["output_format"]] <- output_format
  }
  elipsis <- list(...)

  render_results <- do.call(what = render_method, args = c(render_args, elipsis))

  if (open) {
    if (class(render_results) == "character") {
      lapply(render_results, utils::browseURL)
      return(render_results)
    } else {
      return(render_results)
    }
  } else {
    return(render_results)
  }
}
```

cderv:
> For instance, xaringan::inf_mr() does not have an encoding argument so it would cause an error if user has defined knit: xaringan::inf_mr in their YAML. Furthermore, xaringan::inf_mr() does have different name for the input argument; it uses moon.

Yes you are right. But this means also that

```yaml
knit: xaringan::inf_mr
```
can't be used in a Rmd document inside RStudio because knitting with the Knit Button will throw an error about unused encoding argument. 

I think the `knit` field in yaml, which is currently (almost) undocumented is first designed to be used with RStudio. The RStudio IDE use this field to know what to do.

atusy:
Thanks all.
After I changed the default value of  `customize` to `FALSE`, I also started to feel that customized rendering is better to be done as a separate function.
I close this PR.

@cderv, do you want to make new PR for `custom_render`?

I leave some notes below.

## encoding arguments

@cderv is right.  We need it, though we can ignore it because `render` ignores it.

## RStudio requires knit field be a character

https://github.com/rstudio/rstudio/blob/c4125fa648d830a7e13a320ecc01c6189872c01f/src/cpp/session/modules/SessionRMarkdown.R#L140

## How to use `inf_mr`

It is a bit tricky to use `inf_mr` in knit.
You have to define function and wrap it by parenthesis.

```r
knit: (function(x, ...) xaringan::inf_mr(x))
```

yihui:
I agree with @cderv and @atusy above that if this is implemented at all, it should be done in a separate function instead of the `render()` function (which is complicated enough).

PR #1771, opened 2020-02-12
Enable toc-title in default.html

atusy:
Recent pandoc has an option `toc-title`.
This PR updates `default.html` to enable the option.
It works bot for `toc_float` as well.

For the consistency with other parameters, maybe it is a choice to add `toc_title` as a parameter of `html_document`.
If so, we should also consider adding the parameter to other output formats (e.g., `word_document`)

````
---
output:
  html_document:
    toc: true
    toc_float: true
toc-title: TOC
---

# foo

# bar
````

### DIFF
--- a/inst/rmd/h/default.html
+++ b/inst/rmd/h/default.html
@@ -535,7 +535,7 @@ $if(toc)$
 $if(toc-title)$
 <h2 id="$idprefix$toc-title">$toc-title$</h2>
 $endif$
-$table-of-contents$
+$toc$
 </div>
 $endif$
 $endif$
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: inst/rmd/h/default.html [hunk 1]
```diff
@@ -529,7 +532,10 @@ <h4 class="date">$date$</h4>
 $else$
 $if(toc)$
 <div id="$idprefix$TOC">
-$toc$
+$if(toc-title)$
+<h2 id="$idprefix$toc-title">$toc-title$</h2>
+$endif$
+$table-of-contents$
```
yihui: The variable `table-of-contents` was introduced in Pandoc 2.0: https://github.com/jgm/pandoc/blob/master/changelog.md#pandoc-20-2017-10-29 We need to wait a little longer before changing `toc` to `table-of-contents`.

```suggestion
$toc$
```

Issue #794, opened 2016-09-01
validate requires file but does not check href


timelyportfolio:
While developing `googlefontR`, I discovered that [lines](https://github.com/rstudio/rmarkdown/blob/master/R/html_dependencies.R#L217-L220) only check for `file` and not `href` when validating dependencies.  It seems either `file` or `href` should make a valid dependency.  I would like to not have to download every font available from Google.  Is it possible to validate with just `href` as `src`?

timelyportfolio:
Any inclination to address this, especially in light of https://github.com/rstudio/shiny/issues/1389?  Will close if no interest or inclination, but the asynchronous behavior should not affect non-shiny rmarkdown.  Downloading, storing, and maintaining local dependencies is difficult, but impossible in the case of something like google fonts.

crazycapivara:
I also have the problem with my [deckgl-r-package](https://github.com/crazycapivara/deckgl). In the latest version I added some js libs via the `href` parameter to not blow up the package size. I am creating a gallery site but cannot use features where the libs are included via `href`.

matthewstrasiotto:
same here with my package [mandrake](https://github.com/matthewstrasiotto/mandrake) . 

As of  https://github.com/rstudio/htmltools/pull/188 , crossorigin resource checks can use SRI and are safe, but I was very surprised to see that these two APIs are not in alignment. 

For now, I suppose I'll install @crazycapivara 's fork (thank you for that) to drive my site.

cderv:
Hopefully this should now be fixed in **rmarkdown** dev version. 

```r
remotes::install_github("rstudio/rmarkdown")
```

Thanks for the report - it was long overdue. Thanks @crazycapivara and @matthewstrasiotto for the two PRs that greatly help makes this finally happens.

matthewstrasiotto:
Ah, sad to not get a contributor listing, but i must admit i did forget my promise to give you push access to my fork.
Thank you making it happen!

cderv:
The commit for the PR is still co-authored by you - this was not loose in the process of making another PR. See https://github.com/rstudio/rmarkdown/commit/7107c68047c6952fada50aa8df6cbc2f75ad7cd4

yihui:
[...]

PR #1805, opened 2020-04-22
Support href dependencies

crazycapivara:
If dependencies are provided via the `href` property the document will not render.

This PR checks whether the given url for `href` exists and creates the needed tags for the header, so that the document can be rendered.

### DIFF
--- a/R/html_dependencies.R
+++ b/R/html_dependencies.R
@@ -194,16 +194,26 @@ html_reference_path <- function(path, lib_dir, output_dir) {
 # return the html dependencies as an HTML string suitable for inclusion
 # in the head of a document
 html_dependencies_as_string <- function(dependencies, lib_dir, output_dir) {
+  # Only copy deps with path
+  dependencies_href <- filter_dependencies(dependencies, "href")
+  dependencies <- filter_dependencies(dependencies, "file")
 
   if (!is.null(lib_dir)) {
     dependencies <- lapply(dependencies, copyDependencyToDir, lib_dir)
     dependencies <- lapply(dependencies, makeDependencyRelative, output_dir)
   }
-  return(renderDependencies(dependencies, "file", encodeFunc = identity,
+  #return(
+  tags <- renderDependencies(dependencies, "file", encodeFunc = identity,
                             hrefFilter = function(path) {
                               html_reference_path(path, lib_dir, output_dir)
                             })
-  )
+  #)
+  HTML(paste0(c(tags, renderDependencies(dependencies_href, "href")), collapse = "\n"))
+}
+
+filter_dependencies <- function(dependencies, src_type = "file") {
+  deps <- lapply(dependencies, function(dep) if(!is.null(dep$src[[src_type]])) return(dep))
+  deps[!sapply(deps, is.null)]
 }
 
 # check class of passed list for 'html_dependency'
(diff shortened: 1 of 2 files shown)


crazycapivara:
#794

jcheng5:
For what it’s worth, we don’t validate hrefs anywhere else in htmltools, shiny, or htmlwidgets. Also, it’s not impossible that the href is actually relative (we do this in Shiny for example).

crazycapivara:
You are right, we can skip the test and just make it render with deps provided via `href`. So we can also drop the RCurl dep.

matthewstrasiotto:
Hi @yihui , please consider pulling this, I've just had a PR merged in https://github.com/rstudio/htmltools/pull/188 that allows rendering integrity / crossorigin.

When rendering my document, I was surprised to see that the htmltools API for htmlDependencies is incompatible with your API.

Is there anything holding this PR back? I'd be happy to help resolve it

matthewstrasiotto:
The RCurl dependency is definitely not worth it, there's no reason that it needs to be guaranteed that a resource is available @ build time in order for it to available @ runtime, 

for example, 
- if the building host doesn't have network connectivity/is behind a firewall, etc etc. 
- intermittent failures with the remote CDN at build time do not imply it won't be available @ runtime. 

I'm going to fork your branch @crazycapivara & fully remove the Rcurl dep, and any corresponding checks.

cderv:
I think will go with #1948 for fixing this and not check the href url in **rmarkdown**

R/html_dependencies.R
```diff
@@ -224,19 +234,36 @@ validate_html_dependency <- function(list) {
...
+      file <- system.file(file, package = list$package)
+    if (!file.exists(file)) {
+      utils::str(list)
+      stop("path for html_dependency not found: ", file, call. = FALSE)
+    }
+  }
+
+  if (!is.null(list$src$href)) {
+    href <- list$src$href
+    if (!RCurl::url.exists(href))
```
jcheng5, line 254: FWIW we don't do this check in other code paths (like Shiny); if an href is provided, we just assume it'll work. My only reason for pointing this out is because it's possible that the process that's rendering the Rmd might be running on a different machine, with different network connectivity, than the browser that will ultimately be loading the output. (For example, it's pretty common for RStudio Connect or RStudio Server Pro to be run on an airgapped host.)

R/html_dependencies.R
```diff
@@ -224,19 +234,36 @@ validate_html_dependency <- function(list) {
...
+      stop("href for html_dependency not found: ", href, call. = FALSE)
   }
 
   list
 }
 
+# If we do not want to use RCurl dep
+validate_dependency_url <- function(src) {
```
jcheng5, line 262: See above note--I think it's better not to check

R/html_dependencies.R
```diff
@@ -194,16 +194,26 @@ html_reference_path <- function(path, lib_dir, output_dir) {
 # return the html dependencies as an HTML string suitable for inclusion
 # in the head of a document
 html_dependencies_as_string <- function(dependencies, lib_dir, output_dir) {
+  # Only copy deps with path
+  dependencies_href <- filter_dependencies(dependencies, "href")
+  dependencies <- filter_dependencies(dependencies, "file")
```
jcheng5, line 199: I think it's not necessary to split the dependencies into `file` and `href` if you take advantage of two things:

1. `copyDependencyToDir` and `makeDependencyRelative` can both be passed `mustWork = FALSE` in which case it will just ignore non-file based dependencies
2. `renderDependencies` will take multiple src types, e.g. `srcType = c("file", "href")`, and it will use the first of those that's present in the dependency.

jcheng5 (changes requested):
Thank you for this contribution!

cderv:
Thanks a lot for you comment @jcheng5 ! 
We're currently working on #1948 because it was another PR forked from here without the RCurl check. No worries, your comments apply there too.

R/html_dependencies.R
cderv: > copyDependencyToDir and makeDependencyRelative can both be passed mustWork = FALSE in which case it will just ignore non-file based dependencies

That will simplify indeed ! thanks

> renderDependencies will take multiple src types, e.g. srcType = c("file", "href"), and it will use the first of those that's present in the dependency.

It seems we use special `encodeFunc = identity` and a special `hrefFilter=` function for `srcType = "file"`. 
Currently it prevents us to do one call for all srcType because we don't want `href` srcType to be processed with those functions. 
https://github.com/rstudio/rmarkdown/blob/145046175c721af185aa6ba3ecc9262f12dc7369/R/html_dependencies.R#L306-L309

It is an old code but it could have been a requirement for Pandoc. Processing both differently would lead to this https://github.com/rstudio/rmarkdown/pull/1948#discussion_r608469286

Do you remember maybe ? (long shot from blame finding https://github.com/rstudio/rmarkdown/commit/4ad77fa4b3782c8c3628ddb035474a5a03e18d66#diff-028517fb0fb57172622ddaa0acc9296ae94664b57e4d390760e14c0c835301c3R74)

jcheng5: Ah, sorry, I have no recollection. If you need me to I can try to figure it out.

cderv:
This PR (as the other one) has been ported into #2151 which is now now merged. I'll close this PR then. 
Thanks again @crazycapivara !

PR #1824, opened 2020-05-16
paste image into rmd or blogdown post

xiaoa6435:
a addin convenient for  insert image. when have binding shortcuts (for example, cmd + v),  we can just take a screenshoot or copy a image, then 'cmd + v'  paste to rmd or blogdown post.

This will also display a dialog on paste -- which could always be instantly dismissed with Enter -- to supply a meaningful filename (can be override by options(rmarkdown.show_dialog = FALSE)).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -101,7 +101,8 @@ Suggests:
     tibble,
     fs,
     pkgdown,
-    callr (>= 2.0.0)
+    callr (>= 2.0.0),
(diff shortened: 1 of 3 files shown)


xiaoa6435:
Fixes #rstudio/rstudio#4038.

xiaoa6435:
en, xfun seems to be a more proper place, I have submitted a PR in xfun.

Issue #1828, opened 2020-05-28
Rmd-to-html codeblock highlight delivers unpleasant information for screen readers


jooyoungseo:
When code highlight is applied for `rmarkdown::html_document` output, screen-reader users experience unneccessary difficulties in reading code blocks.

I am not going to duplicate the issue here so please kindly visit [pandoc/issue#6352](https://github.com/jgm/pandoc/issues/6352#issuecomment-626106786) that I have detailed there.

Fortunately, the fix will be implemented in the next `pandoc` version; however, I wanted to create backword solution towards previous versions <= v2.9.2.1.

To this end, I have attempted to write a simple `lua` filter, but that does not seem to be possible unless we consider postprocessing an html output.

Many blind people, including me, have already experienced this challenge caused by redundant codeblock ID information, and I think we need to come up with some universal solutions for Rmd-to-html output process given that many tutorials and online books have been produced by rmarkdown/pandoc under the hood.

Does anyone have a great solution?

jooyoungseo:
I think I have figured out how to address this issue globally using js and `rmarkdown:::html_dependencies()`. I will make a PR soon.

jooyoungseo:
The following code would resolve this issue when embedded in html output:

``` js
/* Hide empty <a> tag within CodeBlock for screen reader accessibility (see https://github.com/jgm/pandoc/issues/6352#issuecomment-626106786) --> */

const codeList = document.getElementsByClassName("sourceCode");

for(var i in codeList) {
  var linkList = codeList[i].getElementsByTagName('a');
  for(var j in linkList) {
    if(linkList[j].innerHTML === "") {
      linkList[j].setAttribute('aria-hidden', 'true');
    }
  }
}
```

@yihui, where would be the best place for me to put the `js` I have  written for this fix? I added a new function (referencing `rmd/h/accessibility/this.js>,                                                                                                         `) to `htmlDependency.R`, and called the function within `html_document` by appending it to the `extra_dependencies` list on top of `html_dependency_jquery`, `html_dependency_jqueryui`, etc. But, that does not call my js when rendered. Should I directly modify `default.html`? But, wouldn't it be a global fix for people using different html themes?

atusy:
@jooyoungseo Would you share your branch of your forked repository so that we can help you?

jooyoungseo:
Thanks very much @atusy! I have made a PR as a draft via [#1828](https://github.com/rstudio/rmarkdown/pull/1833)

[...]

PR #1833, opened 2020-06-02
Screen reader accessibility improvement for highlighted codeblock of html output

jooyoungseo:
This PR addresses #1828, and have been tested and confirmed with many hours of testing with various combinations of different screen readers and browsers.

## Current Issue

When code highlight is used for html output, screen reader users hear unnecessary and annoying information in codeblock, such as "#cb1-1," "#cb3-1," "#cb5-1," and "#cb5-2."

This is because the code anchor (`<a>`) produced by `pandoc skyligting` has empty content, and each `id` attribute value is read out by screen readers.

To avoid this issue, we need to put `aria-hidden = "true"` per each empty `<a>` code anchor within "source code" class.

While the `pandoc` fix will be already made and will be available in next version [see pandoc/issue#6352](https://github.com/jgm/pandoc/issues/6352#issuecomment-626106786), we need a backword solution towards current and previous `pandoc` versions (between v2.7.3 and v2.9.2.1).

I have written a simple js that can address this issue, and this simple fix can dramatically improve the screen reader accessibility for any `html_document`-like output.

## Testing File

````

---
output:
  html_document:
    highlight: "tango"
---

```{r}
summary(airquality$Ozone)
mean(airquality$Ozone)
```


```{r}
data(iris)
tail(iris)
```
````

## Output to Screen Readers (without this fix)

* Notes: this issue can be resolved with this PR.

```
#cb1-1
 summary(airquality$Ozone)
## Min. 1st Qu. Median Mean 3rd Qu. Max. NA’s 
## 1.00 18.00 31.50 42.13 63.25 168.00 37
#cb3-1
 mean(airquality$Ozone)
## [1] NA
#cb5-1
 data(iris)
#cb5-2
 tail(iris)
## Sepal.Length Sepal.Width Petal.Length Petal.Width Species
## 145 6.7 3.3 5.7 2.5 virginica
## 146 6.7 3.0 5.2 2.3 virginica
## 147 6.3 2.5 5.0 1.9 virginica
## 148 6.5 3.0 5.2 2.0 virginica
## 149 6.2 3.4 5.4 2.3 virginica
## 150 5.9 3.0 5.1 1.8 virginica
```

### CODE REVIEW

**Comment by jooyoungseo** [comment]
@atusy, it still does not work. Haha, I do not understand why this simple thing takes so long time. Any advice would be greatly appreciated.

**Comment by jooyoungseo** [comment]
FYI, the js itself works perfectly when I render with `includes(after_body())`.

File: inst/rmd/h/pandoc/accessibleCodeBlock.js [hunk 1]
```diff
@@ -0,0 +1,16 @@
...
+
+  for(var i in codeList) {
+    var linkList = codeList[i].getElementsByTagName('a');
+    for(var j in linkList) {
+      if(linkList[j].innerHTML === "") {
+        linkList[j].setAttribute('aria-hidden', 'true');
+      }
+    }
+  }
+}
```
atusy, line 16: Missing `)`

R/html_dependencies.R
```diff
@@ -71,6 +71,19 @@ html_dependency_tocify <- function() {
...
 
+# Create an HTML dependency for pandoc_codeblock_accessibility
+#' @rdname html-dependencies
+#' @export
+html_dependency_accessibleCodeBlock <- function() {
+
+  htmlDependency(
+    name = "accessibleCodeBlock",
+    version = "0.0.1",
+    src = pkg_file("rmd/h/pandoc"),
```
atusy, line 82: This will cause other js files be copied to `lib_dir` when `self_contained = FALSE`.
I recommend moving this file back to `accessibility`.
I guess you followed my suggestion, which was immediately removed after posting...
Sorry about this.

inst/rmd/h/pandoc/accessibleCodeBlock.js
atusy: and also `;`

So that final line becomes `});`

jooyoungseo:
@atusy, thanks very much! However, here is a very picky issue.

I have tested and confirmed that neither `document.addEventListener('DOMContentLoaded', function() { ... });` nor `$(document).ready(function() { ... });` add the `aria-hidden` attribute to the empty anchors.

This is very weird because it works successfully if I move the script down after body (of course, after removing the dom load dispatch).

Humm... I think I need to dig into how to tackle this tacky problem.

jooyoungseo:
@atusy, I am so excited to tell you that we have made it! Please kindly ignore the previous comment (it turned out the issue caused by a screen reader bug that delayed caught by the previous buffer history). With several testing with various combinations of browsers and screen readers, this fix is now working welll and safely! :) Thanks very much again. I will add a line to `NEWS.md` shortly.

atusy:
Congrats and thank you for crediting me!!

PR #1834, opened 2020-06-02
 fix dependency not included properly 

atusy:


atusy:
Sorry I opened a PR on wrong place.
Its opened here: https://github.com/jooyoungseo/rmarkdown/pull/1

Issue #1603, opened 2019-07-15
Folding occurs on result code blocks that have specific language classes


atusy:
Currently, application of code folding depends on whether code blocks have specific language classes (`r`, `python`, `bash`, `sql`, `cpp`, `stan`, `julia`)

https://github.com/rstudio/rmarkdown/blob/d4b7bf18b95371cb57e5f54aff579d70a6651396/inst/rmd/h/navigation-1.1/codefolding.js#L20

Thus, code folding is applied for result code blocks if they have the above classes.
This can be an unexpected behavior for users.

````
---
output:
  html_document:
    code_folding: show
---

```{r, class.output='r', comment=''}
identity
```
````

![image](https://user-images.githubusercontent.com/30277794/61249779-0d7e7a80-a791-11e9-9878-d78e308aef62.png)

## xfun::session_info('rmarkdown')

```
R version 3.6.1 (2019-07-05)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Debian GNU/Linux 9 (stretch), RStudio 1.2.1335

Locale:
  LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8     LC_MONETARY=en_US.UTF-8    LC_MESSAGES=C             
  LC_PAPER=en_US.UTF-8       LC_NAME=C                  LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

Package version:
  base64enc_0.1.3  digest_0.6.20    evaluate_0.14    glue_1.3.1       graphics_3.6.1   grDevices_3.6.1  highr_0.8        htmltools_0.3.6  jsonlite_1.6    
  knitr_1.23       magrittr_1.5     markdown_1.0     methods_3.6.1    mime_0.7         Rcpp_1.0.1       rmarkdown_1.14.1 stats_3.6.1      stringi_1.4.3   
  stringr_1.4.0    tinytex_0.14     tools_3.6.1      utils_3.6.1      xfun_0.8         yaml_2.2.0      

Pandoc version: 2.7.3
```

---

By filing an issue to this repo, I promise that

- [o] I have fully read the issue guide at https://yihui.name/issue/.
- [o] I have provided the necessary information about my issue.
[...]

PR #1835, opened 2020-06-02
generalize code_folding of html_document

atusy:
This PR generalize `code_folding` by allowing any engines.
A downside is that turning on and off the `code_folding` invalidates chunk cache.
However, I think it is affordable because users do not change this option frequently.


atusy:
Note that this will also fix https://github.com/rstudio/rmarkdown/issues/1603

atusy:
@yihui 

Your suggestion sounds reasonable, and is reflected.
Thank you!!

PR #1856, opened 2020-07-06
accept file_scope parameter to render

jjallaire:
- New `pandoc_run()` function for running pandoc using `system2()` arguments and return value (a separate function is required for this so that various standard command line arguments and environment variables can be provided).

- Made `file_scope` a parameter to `render()` rather than an option for output format definitions. This was done to enable callers to have dynamic behavior around `file_scope` (e.g. based on whether duplicate footnotes exist).

Required by https://github.com/rstudio/bookdown/pull/912


jjallaire:
The related bookdown PR https://github.com/rstudio/bookdown/pull/912 has been closed so am closing this now as well.

Issue #1860, opened 2020-07-11
Square brackets in md_document


DanChaltiel:
Hi dev team,

I'm using `pkgdown` to document my package, and I thus have to generate an `index.md` file.

My documentation has some square brackets, and the knitting process unexpectedly escapes them. However, it seems that this is then interpreted incorrectly in the further HTML conversion, as the brackets disappear and the inside turns to italic (I think it triggers the math formula format). 

This happens whether I use the Rstudio "knit" button or I use `knitr::knit` myself. 

For now (until I'm able to correct it), you can see an example [here](https://danchaltiel.github.io/crosstable/index.html#margins-and-totals): the second row of the table should have square brackets, as median[IQR]. This also happens using the RStudio "preview" button on the `md` file.

This behavior has already been described in 2016 in https://github.com/rstudio/rmarkdown/issues/667 but was closed without being truly solved. The OP accepted a workaround: including all bracket sequences as inline code. However, this won't work in generated HTML tables such as mine.

Here is a reproducible example:

```Rmd
---
output: md_document
---

I want to describe something as median[IQR].
```

Resulting `md` file:
```md
I want to describe something as median\[IQR\].
```

Resulting HTML:
```html
<p>I want to describe something as median
	<span class="math display">
		<span class="MathJax_Preview" style="color: inherit;"/>
		<div class="MathJax_Display" style="text-align: center;">
			<span class="MathJax" id="MathJax-Element-1-Frame" tabindex="0" data-mathml="<math xmlns=&quot;http://www.w3.org/1998/Math/MathML&quot; display=&quot;block&quot;><mi>I</mi><mi>Q</mi><mi>R</mi></math>" role="presentation" style="text-align: center; position: relative;">
				<nobr aria-hidden="true">
					<span class="math" id="MathJax-Span-1" style="width: 2.574em; display: inline-block;">
						<!-- some mathjax things -->
					</span>
				</nobr>
				<span class="MJX_Assistive_MathML MJX_Assistive_MathML_Block" role="presentation">
					<math xmlns="http://www.w3.org/1998/Math/MathML" display="block">
						<mi>I</mi>
						<mi>Q</mi>
						<mi>R</mi>
					</math>
				</span>
			</span>
		</div>
		<script type="math/tex; mode=display" id="MathJax-Element-1">IQR</script>
	</span>.
</p>

```

[...]

PR #1863, opened 2020-07-18
md_document gains knitr as a markdown variant

atusy:
This PR solves #1860 by allowing `md_document` to output `knitr::knit`ted markdown.
It internally runs pandoc, but replaces the result with an input markdown.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,8 @@ rmarkdown 2.4
 
 - When customizing formats with the `output_format` function, `pre_knit`, `opts_hooks`, and `knit_hooks` can now refer to `rmarkdown::metadata`. Previously, `rmarkdown::metadata` returned `list()` in these functions (thanks, @atusy, #1855).
 
+- `md_document` gains "knitr" as a markdown variant so to create `knitr::knit`ted markdown instead of `pandoc`ed markdown (thanks, @atusy, #1863).
+
 rmarkdown 2.3
 ================================================================================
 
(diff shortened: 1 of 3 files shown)


cderv:
Interesting ! Just adding some thoughts for now... 

It seems a bit odd to mix this with pandoc stuff... 🤔 like output format option are not useful for this variant.
Also pandoc is still run for conversion even if the result is ditched at the end. Seems a bit too much.

Is there a real benefit of using directly `knitr::knit` if you want a md file resulting only by knitr ? 
I did not compare the output yet. I know rmarkdown set a bunch of knitr options and stuffs, so maybe there is a benefit of using the knit result from inside render. 

And also, I think this is equivalent to `rmarkdown::render(..., output_options = list(keep_md = TRUE), run_pandoc = FALSE)`

atusy:
> It seems a bit odd to mix this with pandoc stuff
I totally agree with this.

Although, I'm still uncertain if this PR is a good idea, there are two benefits.
First, we have an option to remove a YAML frontmatter.
The YAML frontmatter can be annoying when markdown is rendered on GitHub.
Second, RStudio users can easily create the output by keyboard short cut or GUI.

atusy:
I close this PR because I found an easy workaround with RStudio.

```yaml
---
knit: (knitr::knit)
---
```

atusy:
Oops the above way is not a solution.
The output contains the YAML front matter, which would be annoying on GitHub.

cderv:
> The YAML frontmatter can be annoying when markdown is rendered on GitHub.

But wouldn’t you use the github_document() format for that ? For a document
aimed at being displayed on GitHub ?

atusy:
Oh yes! That is true. Thanks for the comment!!

I agree the conversion of `index.Rmd` -> `index.md` does not matter on GitHub.
I close the PR and mention the workaround on the original issue.

Issue #1864, opened 2020-07-18
`render` with `intermediates_dir` argument fails when applying original template to `output_format`


atusy:
This is related to #457.
When `intermediates_dir` is specified, `pandoc` runs under the `intermediates_dir`.
As a result, pandoc fails to find template when its path is relative from `input_file`'s working directory.

There are two ways to solve the problem.

1. normalize path of the template file
2. copy the template file to `intermediates_dir`

Which would be a better choice? Maybe 1?

# reprex

````r
pandoc_args <- rmarkdown::html_document()$pandoc$args
file.copy(
  pandoc_args[which(pandoc_args == '--template') + 1L],
  'template.html'
)
writeLines(
  paste('---',
        'output:',
        '  html_document:',
        '    template: template.html',
        '---',
        sep = '\n'),
  'example.Rmd'
)
rmarkdown::render('example.Rmd', intermediates_dir = tempdir())
````

Unset `intermediates_dir` will solve the problem.

## return

````
Could not find data file templates/template.html
Error: pandoc document conversion failed with error 97
````
<details><summary> xfun::session_info('rmarkdown')</summary>

R version 4.0.2 (2020-06-22)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 20.04 LTS, RStudio 1.3.1056

Locale:
  LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8       
  LC_COLLATE=en_US.UTF-8     LC_MONETARY=en_US.UTF-8    LC_MESSAGES=C             
  LC_PAPER=en_US.UTF-8       LC_NAME=C                  LC_ADDRESS=C              
  LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

Package version:
  base64enc_0.1.3 digest_0.6.25   evaluate_0.14   glue_1.4.1      graphics_4.0.2 
  grDevices_4.0.2 highr_0.8       htmltools_0.5.0 jsonlite_1.7.0  knitr_1.29     
  magrittr_1.5    markdown_1.1    methods_4.0.2   mime_0.9        rlang_0.4.7    
  rmarkdown_2.3.2 stats_4.0.2     stringi_1.4.6   stringr_1.4.0   tinytex_0.24   
  tools_4.0.2     utils_4.0.2     xfun_0.15       yaml_2.2.1     

Pandoc version: 2.7.3

[...]

Issue #457, opened 2015-06-19
Custom pandoc template with runtime:shiny not working


lmccusker:
I am unable to use my custom pandoc template in my ioslides presentation if runtime: shiny is included.

This works great:

```

---
title: "Presentation"
output: 
  ioslides_presentation:
    self_contained: false
    template: ./templates/template.html

---
```

But this does not:

```

---
title: "Presentation"
runtime: shiny
output: 
  ioslides_presentation:
    self_contained: false
    template: ./templates/template.html

---
```

I get the following error:

Error : pandoc document conversion failed with error 1

ignacio82:
I have the same issue. Is this a bug or are we both doing something wrong?

Thanks!

lmccusker:
I found a solution!

Remove the "template" call from the header:

    ---
    title: "Presentation"
    runtime: shiny
    output: 
    ioslides_presentation:
        self_contained: false
    ---

And save the template as "default.ioslides".  This seems to be working.

ignacio82:
@lmccusker workaround is not perfect. For example, it wont work if you try to publish to shinyapps.io

atusy:
`rmarkdown::run` sets `intermediates_dir` to temporary directory which does not contain template file.
Pandoc runs in the `intermediates_dir` and thus fails to find the template file.
A workaround is to give an absolute path.

PR #1865, opened 2020-07-19
Add template as output_render_files of discover_rmd_resources()

atusy:
solves #457﻿ and #1864

Please see the discussion in #1864 for the reason of changes.


cderv:
Yes this is now the good fix ! 👍  Thank you ! 

I think the source of the issue is the discovery of template as an external resource. see https://github.com/rstudio/rmarkdown/issues/1864#issuecomment-660706526
I think the NEWS should be that `rmarkdown::find_external_resources()` now discovers external template file (and this fix also the shiny runtime, and intermediate_dir set to something, because the run find_external_ressources)

Maybe we should also take the opportunity to add some tests ? Not sure if there are already some about this, and if this is easy to add.

atusy:
@cderv 
Thanks for your insightful comments and smaller reprex!!
I'll update NEWS and consider about tests.

atusy:
It seems difficult to unit test  `rmarkdown:::discover_rmd_resources` because it's second argument requires a complex function, which is defined internally by `find_external_resources`...
Unit testing `find_external_resources` is also difficult because it `render`s input file and requires valid resources (e.g., `reference_docx` must be a valid `docx` file).

atusy:
A first step to unit test `discover_rmd_resources` would be to pull out `discover_single_resource` from `find_external_resources`, and unit test `discover_single_resource`.
As `discover_single_resource` requires some hyper parameters, we also have to define a function that defines `discover_single_resource`...
I think such a heavy steps should be done separate PR(s).

```r
generate_discover_single_resource <- function(input_dir, .envir = parent.frame()) {
  force(input_dir)
  force(.envir)
  discover_single_resource <- function(path, explicit, web) {

    if (!(is.character(path) && length(path) == 1 && path != "." && path != ".." &&
          file.exists(file.path(input_dir, path))))
      return(FALSE)

    ext <- tolower(tools::file_ext(file.path(input_dir, path)))

    if (identical(ext, "r")) {
      # if this is a .R script, look for resources it contains, too
      discover_r_resources(file.path(input_dir, path), discover_single_resource)
    } else if (identical(ext, "css")) {
      # if it's a CSS file, look for files it references (e.g. fonts/images)
      discover_css_resources(file.path(input_dir, path), discover_single_resource)
    }
    # if this is an implicitly discovered resource, it needs to refer to
    # a file rather than a directory
    if (!explicit && dir_exists(file.path(input_dir, path))) {
      return(FALSE)
    }

    # this looks valid; remember it
    .envir$discovered_resources <<- rbind(.envir$discovered_resources, data.frame(
      path = path, explicit = explicit, web = web, stringsAsFactors = FALSE
    ))
    TRUE
  }
  return(discover_single_resource)
}
```

cderv:
I see. thanks for trying. I know that some steps need to be done in order to improve the "adding unit test easily" in rmarkdown. 

However, I think we can just for now add to the existing one, don't you think ?  
https://github.com/rstudio/rmarkdown/blob/ff1b279e795b62b1ffeabcc9aa5bf7386a7ebb83/tests/testthat/test-resources.R

this would require:
* create a `empty2.html`
* add `template: empty2.html` to the yaml header in https://github.com/rstudio/rmarkdown/blob/master/tests/testthat/resources/rmarkdown.Rmd

What do you think ?

atusy:
@cderv That is a good idea, and glad to hear aboiut `test-resources.R`.
Thanks!! Will try.

My fault is that I ran `usethis::use_test()`, and found no tests in `test-html_resources.R`.

atusy:
@cderv I don't understand why, but I face a warning and an error by updating test.

```
✓ |  OK F W S | Context
x |  13 1 24   | resource discovery 
[WARNING] Could not parse YAML metadata at line 18 column 1: :11:35: Unexpected '
  '

test-resources.R:20: failure: R Markdown resource discovery finds expected resources
`resources` not equal to `expected`.
Attributes: < Component “row.names”: Numeric: lengths (6, 9) differ >
Component “path”: Lengths (6, 9) differ (string compare on first 6)
Component “path”: 6 string mismatches
Component “explicit”: Lengths (6, 9) differ (comparison on first 6 components)
Component “explicit”: 1 element mismatch
Component “web”: Lengths (6, 9) differ (comparison on first 6 components)
Component “web”: 3 element mismatches
```

If I print `resources` and `expected`, they are I see some missing files in `resources`.

```r
> resource
         path explicit   web
6   empty.csv    FALSE FALSE
1    empty.md    FALSE FALSE
2   empty.png    FALSE FALSE
5   empty.Rmd    FALSE FALSE
4   empty.tsv     TRUE FALSE
3 empty2.html    FALSE FALSE

> expected
         path explicit   web
5   empty.css    FALSE  TRUE
9   empty.csv    FALSE FALSE
7  empty.html    FALSE  TRUE
6   empty.jpg    FALSE  TRUE
1    empty.md    FALSE FALSE
2   empty.png    FALSE FALSE
4   empty.Rmd    FALSE FALSE
3   empty.tsv     TRUE FALSE
8 empty2.html    FALSE  TRUE
```

cderv:
This is indeed tricky.
In fact `discover_rmd_resources` will run `discover_html_resources` that will render the Rmd document to html. And as an empty template is provided, some resources previously found in the resulting html are not found now. (Like the css file)

This template argument is a tricky one.

atusy:
I see thanks. I'll copy default template instead.

atusy:
GREAT!! WE PASSED TESTS!!

Huge thanks to @cderv!!

yihui (approved):
The discussions seem lengthy to me so I didn't read them very carefully, but the change seems to be simple enough (essentially one line of change), so I'll just trust you two and merge this PR. Thanks a lot!

PR #1871, opened 2020-07-27
Added alias for pptx and docx

jooyoungseo:
Added alias `pptx_document()` for `powerpoint_presentation()` and `docx_document()` for `word_document()`.

Given that we have `[file_extension]_document()` for output formats (e.g., pdf, html), it would be great if we had alias for pptx and docx as well following the convention.


jooyoungseo:
Oh, I get the picture now! Thanks very much for the detailed explanation. I will close this PR then.

yihui:
Thanks for your understanding!

PR #1874, opened 2020-08-04
Updating site Google Analytics

sarahemlin:
Hey @apreshill --

I tested the updating the GA code with Jonathan McPherson on the IDE Admin docs and this combination of an HTML file with the new GA standard script and then pointing to that file from the yaml files has worked beautifully, if you would please merge this pull request for rmarkdown.rstudio.com.  We've got several things related to GA that are working sub-optimally (or not working) because the ga.js setup was deprecated by Google a while back.

Thanks!
Sarah

### DIFF
--- a/_site.yml
+++ b/_site.yml
@@ -8,7 +8,7 @@ output:
     toc_float:
       collapsed: false
     includes:
-      in_header: include_head.html
+      in_header: [include_head.html, google_analytics.html]
       before_body: include_header.html
       after_body: include_footer.html
     css: css/rmarkdown.css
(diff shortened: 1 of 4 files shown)


apreshill:
Hi @yihui and @cderv -

When we pulled the Google Analytics for the current site, we noticed some issues with the tracking. This PR from @sarahemlin fixes it. Thank you!

Alison

sarahemlin:
@yihui --

Given that I have 19 subdomains to manage GA scripts in, it would make things much simpler for me to have a standard setup and files to edit when I need to (which at this point seems to be at least once per quarter).  With the ga.js file going away, it's the same number of files to manage.  

Thank you for finding that extra line of code to delete; I'll push an updated file by noonPT today.

Sarah

PR #1878, opened 2020-08-25
incorporate CSS styles from latest Pandoc's tempalte

atusy:
This will enable styles such as underlines, smallcaps, and multi-column layouts...

CSS is added to templates of 

- html_document
- slidy

If it is preferred to added to ioslides as well, let me know.

I suggest the same changes to revealjs as well (https://github.com/rstudio/revealjs/pull/82).

### DIFF
--- a/inst/rmd/h/default.html
+++ b/inst/rmd/h/default.html
@@ -27,6 +27,24 @@ $for(header-includes)$
 $header-includes$
 $endfor$
 
+<style type="text/css">
+  code{white-space: pre-wrap;}
+  span.smallcaps{font-variant: small-caps;}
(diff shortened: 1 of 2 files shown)


yihui (approved):
Looks good to me. Thanks!

> If it is preferred to added to ioslides as well, let me know.

I don't see why ioslides shouldn't also have these CSS styles. I'll merge this PR for now, and please feel free to open another one.

atusy:
I totally agree we'd better add these CSS to ioslides as well.
I skipped it just because the official template does not contain ioslides (https://github.com/jgm/pandoc-templates).
I'll do it tonight!

PR #1880, opened 2020-08-25
Add documentation about when render sets output_format options from YAML parameters

jonathan-g:
See https://github.com/rstudio/bookdown/pull/933 and https://github.com/rstudio/bookdown/issues/930

Document behavior of `output_format` argument to `render()`: It sets options from YAML parameters in the YAML header or `_output.yml` only if `output_format` is a character string. If it's a constructed `output_format` object, the YAML material is ignored and parameters must be set by passing arguments to the constructor.


jonathan-g:
Superseded by #1907.

PR #1894, opened 2020-09-08
Add missing lua filter vignettes to articles in pkgdown

cderv:
Don't know if it was not included on purpose but I think it is worth adding


cderv:
Oups... wrong branching... 😭

yihui (approved):
I'm okay with that.

PR #1898, opened 2020-09-11
Discover site generators in index.Rmd files in parent directories

jjallaire:
Note this requires the site generator to add a `subdirs = TRUE` field to opt-in to this behavior

Also note this code is speculative (has not been tested at all) so you'll want to try it out before merging :-)

This should in principle allow the IDE to call `render_site()` for Rmds below the root site directory.


jjallaire:
Have you managed to test this out w/ blogdown yet?

apreshill:
@jjallaire I just tested this with the default lithium theme. I was able to successfully knit to HTML a `.Rmd` post (and later `serve_site()` with no errors), but if I click the knit button on a `.Rmarkdown` post it knits straight to `.html` (expected behavior is to knit to a `.markdown` file).

jjallaire:
Hmm...this should require an update to blogdown (addition of `subdirs = TRUE` to the site generator) in order to work at all. @yihui May know more about what is happening with `.Rmarkdown`.

jjallaire:
Okay, I've now pushed the changes and tested that if a site generator adds `subdirs = TRUE` to it's definition that it will be correctly discovered even for files in subdirectories. I've also verified that the IDE picks this up and calls `render_site()` on these files. LMK if there are other things I can do to make this work better.

yihui (approved):
This works great! I'll merge this PR and push the corresponding (minimal) changes required in **blogdown** in a minute. Thank you!

PR #1899, opened 2020-09-14
Add lua_filters to pandoc_options to expose them in the output_format

### LINKED ISSUE #1897: Improve pandoc lua filters handling to make it more usable from other package
We are currently adding some lua filter also in https://github.com/rstudio/bookdown/pull/940 

There are several pieces missing to make things easier to extend what is done in **rmarkdown**

1. `pandoc_lua_filters` helper is currently only working for **rmarkdown** filter. We currently create the same helper for each 📦 

2. The lua filters are not including the same way in all 📦 so it is not easy to make another package filter run before or after another one. 

About 2, sometime it is add inside the `pre_processor` part of the function https://github.com/rstudio/rmarkdown/blob/284b685c2feb68083b3434c6ca38717cd61cbeea/R/html_document_base.R#L109-L110

and in others outside of the pre preprocessor in the main format function, **before** `pandoc_args` argument
https://github.com/rstudio/rmarkdown/blob/284b685c2feb68083b3434c6ca38717cd61cbeea/R/pdf_document.R#L145-L149

We could start fixing this by being more aligned in all format. 

And to improve, I think we need a mechanism for choosing the order of the filter (pandoc will run in the order of the command line). This is necessary to avoir conflicts. There are several ways to that IMO - one is to make lua filters activate with a switch.

cderv:
This will partly resolves #1897 and this is needed for https://github.com/rstudio/bookdown/pull/940 where we add a lua filter in bookdown. 

Adding `lua_filters` to `pandoc_options` allows exposes the filters in the output format and that will allow to manipulate them to change the order or even remove some if a format use another as its base. 

We need this in bookdown because we want to be sure to run the bookdown lua filter first. 

Currenlty, the PR assumes lua filters are appended by default, i.e if a format depend on an other, the lua filter from the former will execute after the latter. We could consider to inverse this maybe: always run lua filter of a format before the filters of its base format. 

With this change, the addition of the filters at the pandoc command line append in `render` directly and will be always put at the end of the command line.  

I have changed the helpers function for lua but they where not exposed, only accessible by `:::`. I wonder if that is fine. I can do otherwise but it seems better this way.


atusy:
If we export `pkg_file_lua`, I would ask for adding `package` argument.
In this way, `pkg_file_lua('foo.lua', package = 'mypkg')` is possible, and downstream packages do not have to reinvent a similar function.
Also, this encourages consistent directory structures for R Markdown related package.

cderv:
> If we export `pkg_file_lua`, I would ask for adding `package` argument.
> In this way, `pkg_file_lua('foo.lua', package = 'mypkg')` is possible, and downstream packages do not have to reinvent a similar function.
> Also, this encourages consistent directory structures for R Markdown related package.

Yes this was idea, but this means we impose the directory structure. I am more inclined to export `pandoc_lua_filters_args()` which IMO is closer to the other from its family `pandoc_*_args` and which leaves the directory structure open for 📦 developers. I don't think there is already one structure consistent for html resources and pandoc template.

atusy:
I see. I am fine with not exporting `pkg_file_lua`, and that is because I said "if we export" rather than "I want to export".
This function is small enough to neglect the cost for reinvention.

R/render.R
```diff
@@ -847,9 +847,16 @@ render <- function(input,
       utf8_input <- path.expand(utf8_input)
       output     <- path.expand(output)
 
+      pandoc_args <- output_format$pandoc$args
+
+      # if lua filters are provided, add the command line switch
+      if (!is.null(lua_filters <- output_format$pandoc$lua_filters)) {
+        lua_filters <- pandoc_lua_filter_args(lua_filters)
+      }
+      pandoc_args <- c(pandoc_args, lua_filters)
```
atusy, line 856: Considering https://github.com/rstudio/rmarkdown/issues/1897#issuecomment-692117149, I'd suggest a change so that user-specified lua filters comes after document-builtin filters.

```suggestion
      pandoc_args <- c(lua_filters, pandoc_args)
```

cderv: Yes I think you're right. I was hesitating on this one because not sure if it is easier to add a filter that modify first the AST or last, but to be sure our filters are working, they should be first run.

cderv:
This is ready I think. I'll test it with bookdown now.

R/util.R
```diff
@@ -41,6 +41,10 @@ pkg_file_arg <- function(...) {
   pandoc_path_arg(pkg_file(...))
 }
 
+pkg_file_lua <- function(...) {
+  pkg_file("rmd", "lua", ...)
+}
```
yihui, line 46: I tend to follow @atusy's suggestion to export this function to encourage a consistent folder structure for Lua fitlers in other packages. Even if we export it, users don't have to follow this convention. They can choose to use a different structure.

However, if we were to export this, we probably should put the `lua` folder under the `inst/rmarkdown` directory instead of `inst/rmd`. This is to follow the convention of R Markdown templates in packages: https://bookdown.org/yihui/rmarkdown/template-structure.html

As an example, in **pagedown**, we may move the `resources/lua/` folder to `rmarkdown/`: https://github.com/rstudio/pagedown/tree/master/inst/resources/lua Then we can use `rmarkdown::pkg_file_lua('foo.lua', package = 'pagedown')`.

Anyway, this is not super important to me, and doesn't need to be done in this PR.

yihui (approved):
Looks perfect! You are all set. I'll tweak the wording in NEWS and merge this PR soon. The `pkg_file_lua` stuff could be done in a new PR if necessary. Thanks!

R/output_format.R
```diff
@@ -290,6 +291,10 @@ knitr_options_pdf <- function(fig_width,
 #'   chooses default based on \code{to}). This is typically used to force
 #'   the final output of a latex or beamer conversion to be \code{.tex}
 #'   rather than \code{.pdf}.
+#' @param lua_filters  Character vector of file paths to lua filters to use with
```
yihui, line 294: Nitpicking: for consistency with the naming in Pandoc manual https://pandoc.org/lua-filters.html, I'd capitalize `l`, i.e. `Lua filter`. Similarly, I often use `Pandoc` when referring to the software, and `pandoc` only to refer to the command. Not really a big deal. If you want to make such changes, feel free to push to the master branch directly.

cderv: Oh ok. Sure, naming things is indeed important. Maybe we can share somewhere an index of all words and names that we want to be consitent ?

yihui: At the moment, I can't think of any other words besides "Pandoc" and "Lua filter", so perhaps we don't need a checklist for this small problem :)

R/util.R
cderv: I can see why it is useful. I'll do that; issue opened in #1903

R/output_format.R
cderv: On my list it adds up with TinyTeX, and TeX Live 😄 But yes it is small ! I just wondered if it was the tip of the iceberg 😆

yihui: The only spelling that bothers me is `TinyTex` (similarly, `LaTex`). I can tolerate any other words!

PR #1900, opened 2020-09-14
only install pkgdown for build

cderv:
This should fix the travis build for R 3.3. It reverts 97142d9453e91c225cd60e8ef786967ba92009f2 but we can revert again when new systemfonts hits CRAN.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ jobs:
         - "[[ -z ${NETLIFY_AUTH_TOKEN} ]] && exit 0"
         - nvm install stable
         - npm install netlify-cli -g
+        - Rscript -e 'xfun::pkg_load2("pkgdown")'
         - Rscript -e 'pkgdown::build_site(".", document = FALSE, preview = FALSE)' && netlify deploy --prod --dir docs
     - r: devel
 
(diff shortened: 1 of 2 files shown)


cderv:
This will prevent build failure until **systemfonts** hits CRAN. 
However, we really don't need to install pkgdown for all R versions. Only the one we use for the pkgdown building.

Travis is not easy to control cache or installation behavior. We can't also easily use RSPM binaries. This could change if we swtich to GHA.

yihui (approved):
Great! Thank you!

yihui:
I didn't really want **pkgdown** to be in Suggests. Let's see if `before_cache` works.

cderv:
We should try also to be more precise with the `if` , as pkgdown is required only in specific case
https://github.com/rstudio/rmarkdown/blob/1a96753b91510a873ac0399154d6e1033070de10/.travis.yml#L35-L36

no need to install it each time. But if caches works, I guess it is the same.  I lost my travis skills now that I recently played with GHA 😅

yihui:
> We should try also to be more precise with the `if`

As usual, you are absolutely correct! :)

cderv:
> But if caches works, I guess it is the same

That is what I thought too. Sometimes I am being a bit to rigourous 😄 
Thanks !

PR #1904, opened 2020-09-23
Export pkg_file_lua for package author to use

### LINKED ISSUE #1903: Export pkg_file_lua and promote common organisation of lua filters in packages
Following https://github.com/rstudio/rmarkdown/pull/1899#discussion_r492834759, we could export `pkg_file_lua` so that there is a helper that goes with a common location of lua filters inside packages. 

`inst/rmarkdow/lua` would be a good place.

cderv:
This will close #1903

Once merged we could use it in other packages like **bookdown** and **pagedown**


R/util.R
```diff
@@ -33,16 +33,44 @@ pandoc_output_ext <- function(ext, to, input) {
...
+#'   \code{\link{pandoc_path_arg}} to be used by Pandoc.
+#' @export
+#' @examples
+#' # get all filters
+#' pkg_file_lua(package = "rmarkdown")
+#' # get a specific filter
+#' pkg_file_lua("pagebreak", package = "rmarkdown")
+#' pkg_file_lua(c("pagebreak", "latex_divs"), package = "rmarkdown")
+#' pkg_file_lua("pagebreak.lua", package = "rmarkdown")
+pkg_file_lua <- function(filters = NULL, package = NULL) {
```
yihui, line 67: We can make our own life easier by giving the `package` a default:

```suggestion
pkg_file_lua <- function(filters = NULL, package = "rmarkdown") {
```

R/util.R
```diff
@@ -33,16 +33,44 @@ pandoc_output_ext <- function(ext, to, input) {
...
+  if (is.null(package) || length(package) > 1)
+    stop("One package name in which to look for Lua filters must be provided.", call. = FALSE)
+  lua_folder <- pkg_file("rmarkdown", "lua", package = package)
+  if (is.null(filters)) filters <- list.files(lua_folder, pattern = "*\\.lua$")
```
yihui, line 71: I think the `*` is unnecessary:

```suggestion
  if (is.null(filters)) filters <- list.files(lua_folder, "\\.lua$")
```

R/util.R
```diff
@@ -33,16 +33,44 @@ pandoc_output_ext <- function(ext, to, input) {
...
+  filters <- xfun::with_ext(filters, "lua")
```
yihui, line 72: I'm not sure if we want to be clever here... This trick doesn't seem to save much typing effort for users, and the cleverness also requires extra documentation.

yihui (changes requested):
Looks good overall. Thanks!

R/util.R
cderv: I wanted to do that at first, but as the function is now exported I did not because I thought it would strange for the user to have this function default to the filter from inside **rmarkdown**. I may be overthinking it ?

R/util.R
cderv: ok. I need to learn to be less clever in my feature then 😅

cderv: Simplified in 716bf90

cderv: But I may have misunderstood: by trick did you mean only the "adding missing extension" ? 
I also removed the "list all the filters available". Was also too clever maybe. 
One could now do `list.files(pkg_file_lua(package="rmarkdown"))`

R/util.R
yihui: It's also important to make our own code simpler :) In the current PR, you had to use `package = "rmarkdown"` in many places. If this is the default, you wouldn't need to add these `package` arguments repetitively.

This function belongs to the **rmarkdown** package, and I don't feel it's strange that the package name defaults to **rmarkdown**. For other package authors, they can easily write their own wrapper function, e.g.,

```r
pkg_file_lua <- function(...) {
  rmarkdown::pkg_file_lua(..., package = "foo")
}
```

R/util.R
yihui: Originally I mainly meant "adding the missing extension" (typing `"foo.lua"` is not significantly more difficult than `"foo"`, and our price to pay for this cleverness is the need for extra documentation, which is a little high in my opinion).

Listing all filters via the special value `filters = NULL` sounds okay to me and can be convenient for users. Feel free to bring it back and document it.

R/util.R
cderv: I see. It make complete sense. I'll remember that. 
Done f8da23b

R/util.R
cderv: > and our price to pay for this cleverness is the need for extra documentation, which is a little high in my opinion).

I see the logic. I'll remember that too

Reverted in c7ae478

cderv:
Review was taken into account and test is passing. I'll merge. You can modify if I missed anything. 

I'll be porting this to other packages using lua filter

yihui:
Sounds good. Thanks!

cderv:
So at the end, this is not as useful as I thought because it won't work _as expected_ in another package development workflow. 

* It work well from inside rmarkdown package, because `devtools::load_all` will add a shim for `system.files`
* but this shim does not apply when `system.files` is used in another function from a package. That means using `rmarkdown::pkg_file_lua` inside **bookdown** will not allow to use the system.file shim. That means, when doing `load_all` on bookdown, it won't find the lua files. You need to install the development package first!

It is a bit annoying while developping and you are used to devtools workflow. **testthat** is using load_all also. So test won't work with this function. 

However, if you use last version of **testit** no issue... because it install the package before testing 😉

Anyway, I note this here so that we don't forget what happens. I don't know if the shims in **pkgload** could work differently - I'll may open an issue there.

PR #1907, opened 2020-09-24
Add documentation about when render sets output_format options from YAML parameters

jonathan-g:
See rstudio/bookdown#958 and rstudio/bookdown#930

Document behavior of output_format argument to render(): It sets options from YAML parameters in the YAML header or _output.yml only if output_format is a character string. If it's a constructed output_format object, the YAML material is ignored and parameters must be set by passing arguments to the constructor.


yihui (approved):
Great. Thank you very much!

NEWS.md
```diff
@@ -25,6 +25,8 @@ rmarkdown 2.4
 
 - Fixed a bug that a chunk with a class `fold-hide` hides the rest of the chunks even the output format setting `html_document(code_folding = "show")` (thanks, @atusy, #1906).
 
+- Updated documentation for `render` to make it clearer how options are set for the `output_format` parameter (See <https://github.com/rstudio/bookdown/issues/930>) (thanks, @jonathan-g, #1907)
```
yihui, line 28:
```suggestion
- Updated documentation for `render()` to make it clearer how options are set for the `output_format` parameter (thanks, @jonathan-g, #1907 and rstudio/bookdown#930).
```

PR #1908, opened 2020-09-29
incorporate CSS styles from latest Pandoc's tempalte (ioslides)

atusy:
reflecting https://github.com/rstudio/rmarkdown/pull/1878#pullrequestreview-498085930

### DIFF
--- a/inst/rmd/ioslides/default.html
+++ b/inst/rmd/ioslides/default.html
@@ -92,23 +92,23 @@ $endfor$
 /* https://github.com/ropensci/plotly/pull/524#issuecomment-468142578 */
 slide:not(.current) .plotly.html-widget{
   display: block;
-
-  code{white-space: pre-wrap;}
-  span.smallcaps{font-variant: small-caps;}
-  span.underline{text-decoration: underline;}
-  div.column{display: inline-block; vertical-align: top; width: 50%;}
-  div.hanging-indent{margin-left: 1.5em; text-indent: -1.5em;}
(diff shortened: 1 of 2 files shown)


inst/rmd/ioslides/default.html
```diff
@@ -92,6 +92,22 @@
...
+  ul.task-list{list-style: none;}
+  $if(quotes)$
+  q { quotes: "“" "”" "‘" "’"; }
+  $endif$
+  $if(highlighting-css)$
+  $highlighting-css$
+  $endif$
+  $if(displaymath-css)$
+  .display.math{display: block; text-align: center; margin: 0.5rem auto;}
+  $endif$
```
cderv, line 110: Are you sure this has to be added inside this selector ? 
I believe those styles is for the whole document instead, right ? 

Just a bracket issue I think

inst/rmd/ioslides/default.html
```diff
@@ -92,6 +92,22 @@
...
 }
```
cderv, line 111: Are you sure this has to be added inside this selector ? 
I believe those styles is for the whole document instead, right ? 

Just a bracket issue I think

atusy: Oops!! Thanks for the review. Should be fixed now via [`438579c` (#1908)](https://github.com/rstudio/rmarkdown/pull/1908/commits/438579c9a84eb3b387cd38a70552d2d06f263bd3)

yihui (approved):
Thanks!

PR #1926, opened 2020-10-16
ui.Rmd as alternative to ui.R for defining shiny UI

jjallaire:
This PR implements support for defining Shiny UI via a ui.Rmd file as an alternative to ui.R. It takes advantage of the existing `runtime: shiny_prerendered` capability (which now has a friendlier alias in `runtime: shinyrmd`), but rather than encouraging use of multiple contexts (e.g. `{r context="setup"}`, `{r context="server"}`, etc.) it uses the Rmd file strictly for user-interface (you then create a server.R and optionally global.R just like a conventional shiny application).

There are a couple of motivations for this:

1) Some people feel that it would be much easier to teach Shiny to new users if the UI could be created laid out in markdown rather than with `htmltools` tags (this would be especially true if we created Lua filters to access bs4 layouts/components easily from markdown);

2) This enables taking a primarily static Rmd (which may be expensive to render) and adding just a bit of Shiny for interactivity. For example,  you could convert a static flexdashboard into a "dynamic" dashboard with a single interactive shiny visualization (but still keep all of the static bits you have).

Here's what the code would look like:

#### ui.Rmd

````markup
---
title: "Hello Shiny"
runtime: shinyrmd
---

```{r, echo=FALSE}
sliderInput("bins", "Number of bins:", min = 1, max = 50, value = 30)
plotOutput("distPlot")
```
````

#### server.R

```r
function(input, output) {
  output$distPlot <- renderPlot({
    x <- faithful[, 2]  # Old Faithful Geyser data
    bins <- seq(min(x), max(x), length.out = input$bins + 1)
    hist(x, breaks = bins, col = 'darkgray', border = 'white')
  })
}
```

You could run this app with:

```r
rmarkdown::run()
```

Note that b/c we are using shiny prerendered it's also possible to deploy the app on a server and skip the render step (just use the previously built .html file). On RSC, you'd like render on deploy, and then server the rendered app from .html when users visit the app.

@aronatkins This has some changes to `rmarkdown::run()` (principally looking for a ui.Rmd file in the same places that we now look for an index.Rmd file -- both work) that we'll want to make work w/ RSC invocation of `rmarkdown::run()`. Also note that `runtime: shinyrmd` is now an alias for `runtime: shiny_prerendered` so anywhere that you are looking specifically for "shiny_prerendered" you'll need to also look for "shinyrmd".

@jcheng5 and @wch this does not implement support for auto-reloading of global.R and server.R. I think this is the correct behavior because the ui.Rmd file also not reloaded when it changes. Making global.R and server.R reload would be easy based on similar code you already have in Shiny, but making ui.Rmd re-load / re-render would be pretty tricky as far as I can tell.


yihui (approved):
Looks good to me!

aronatkins:
With shinyapps.io deployments (not customer installations), the presence of either `server.R` or `app.R` implies that we should use `shiny::runApp` to launch the application. When the content is not a Shiny application, we then look for the presence of Rmd files and decide to use `rmarkdown::run`. We will need to change this logic in Connect to look at both R and Rmd files before deciding which path to follow. We will also need to adjust how we handle root requests to this content; the lack of `index.Rmd` and the presence of `index.htm` (created by `rsconnect`) probably means that we will route folks to the HTML file rather than straight to the app.

With customer deployments of Connect, we will need to adjust `rsconnect` to detect this style of content as ShinyRmd. Once we have the correct metadata in the `manifest.json`, the remainder of the deploy will likely just work.

I'll reach out privately to coordinate this work.

yihui:
@jjallaire Oh I forgot to mention one minor thing: `shiny_rmd` is probably a better name than `shinyrmd`, given that we have `shiny_prerendered`.

jjallaire:
@yihui we are going to deprecate the use (or at least the documentation of
the use) of shiny_prerendered so consistency w/ that isn’t critical. Joe
and I tried it
with a dash and underscore but liked the one w/o better (maybe b/c many
shiny derivative packages use that style)

yihui:
Got you. I just felt a full name `shiny` and an acronym `rmd` don't match without a separator, but never mind.

Issue #1487, opened 2018-12-03
replacing paged_table_type_sum with tibble::type_sum


sebkopf:
I'm curious if this is still the plan as noted in the comments for `paged_table_type_sum`? It would be extremely useful to help standardize the display of column types between interactive RStudio use and knitted documents displaying paged tables, especially for new classes that implement their own `type_sum.new_class` methods.

https://github.com/rstudio/rmarkdown/blob/34d2fd0ebceb91705a54692ee5cfb46f0f09e28e/R/html_paged.R#L2

yihui:
Cc @javierluraschi I think you are the right person to ask.

sebkopf:
Following up on this issue - pillar is moving away from the `type_sum()` generic for succinct column headers and towards the new `vctrs::vec_ptype_abbr()` which provides easy-to-implement data type abbreviations as part of the excellent [vctrs](https://github.com/r-lib/vctrs) framework. At present, new `vctrs` data types still show up as generic `<S3: class name>` in paged tables - would you be willing to consider a move towards `vctrs::vec_ptype_abbr()` in `rmarkdown:::paged_table_type_sum` for S3 classes that extend `vctrs_vctr`? This would make it possible to have consistent column headers for custom data types in interactive RStudio and knitted documents with paged tables. I'd be happy to provide a suggestion via pull request if you're okay with adding `vctrs` to the `Imports`?

PR #1927, opened 2020-10-16
support vctrs abbreviations

sebkopf:
Suggested fix for #1487 , this follows the equivalent use of `vctrs::vec_ptype_abbr` in the pillar package (see [code](https://github.com/r-lib/pillar/blob/08a96a94acdad5c69f524b518ace9287c6e1a463/R/type-sum.R#L38)). 

Note that it would be possible to replace the `paged_table_type_sum()` function almost entirely with what is implemented broadly by the `vctrs::vec_ptype_abbr` generic (see e.g. implementation for [atomic data types](https://github.com/r-lib/vctrs/blob/e1d02ed99a0a22292018e03adf2af7c8c2aad213/R/ptype-abbr-full.R#L62), [date time](https://github.com/r-lib/vctrs/blob/f5f57e0219a010dc09d8228e690969e6845050f5/R/type-date-time.R#L114), and [factors](https://github.com/r-lib/vctrs/blob/09aabd71d3f79bfda8b22489ead69b44d28bfab0/R/type-factor.R#L65)) but it would introduce small differences (e.g. the factor abbreviation would become `fct` instead of `fctr`) and I wasn't sure if this is something you'd be interested in.

ps: since this is my first pull request to rstudio, I have sent the signed contributor agreement to jj@rstudio.com

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -91,14 +91,14 @@ Imports:
     tinytex (>= 0.11),
     xfun (>= 0.15),
     methods,
-    stringr (>= 1.2.0),
-    vctrs (>= 0.3.2)
+    stringr (>= 1.2.0)
 Suggests:
     shiny (>= 0.11),
     tufte,
     testthat,
     digest,
     dygraphs,
+    vctrs,
     tibble,
     fs,
     rsconnect
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -91,7 +91,8 @@ Imports:
     tinytex (>= 0.11),
     xfun (>= 0.15),
     methods,
-    stringr (>= 1.2.0)
+    stringr (>= 1.2.0),
+    vctrs (>= 0.3.2)
 Suggests:
     shiny (>= 0.11),
     tufte,
--- a/R/html_paged.R
+++ b/R/html_paged.R
@@ -1,6 +1,7 @@
 # paged_table_type_sum and paged_table_obj_sum should be replaced
 # by tibble::type_sum once tibble supports R 3.0.0.
 paged_table_type_sum <- function(x) {
+  if (vctrs::vec_is(x)) return(vctrs::vec_ptype_abbr(x))
   type_sum <- function(x)
   {
     format_sum <- switch(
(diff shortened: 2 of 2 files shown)


DESCRIPTION
```diff
@@ -91,7 +91,8 @@ Imports:
     tinytex (>= 0.11),
     xfun (>= 0.15),
     methods,
-    stringr (>= 1.2.0)
+    stringr (>= 1.2.0),
+    vctrs (>= 0.3.2)
```
yihui, line 95: I'm hesitant on introducing a new dependency only for the purpose of supporting a specific feature `html_document(df_print = "paged")`. I'm okay if it is in `Suggests`, and we call it conditionally, e.g., `if (xfun::loadable('vctrs')) return(vctrs:: vec_ptype_abbr(x))`.

If the difference between the output of `rmarkdown:::paged_table_type_sum()` and `vctrs::vec_ptype_abbr()` is minor, I tend to fix the difference to make them consistent.

yihui (commented):
Thanks!

DESCRIPTION
sebkopf: Thank you, I moved `vctrs` to suggests and included the check whether the package is installed. This should work well since anyone who wants to make use of vctrs in a paged table will have it installed anyways. Could also skip the suggests altogether since `tibble` is already on the suggests list for `rmarkdown` and `tibble` depends on `vctrs`. 

With this, these kind of `vctrs`-type vector implementations now print similarly in standard output and paged tables:

```{r}
library(vctrs)
percent <- function(x = double()) {
  vec_assert(x, double())
  new_vctr(x, class = "percent")
}
format.percent <- function(x, ...) {
  paste0(formatC(signif(vec_data(x) * 100, 3)), "%")
}
vec_ptype_abbr.percent <- function(x, ...) "prcnt"

df <- tibble::tibble(x = percent(runif(4)))
```
`df` prints as:

```
## # A tibble: 4 x 1
##         x
##   <prcnt>
## 1   14.1%
## 2   51.5%
## 3   65.5%
## 4   95.5%
```

and `rmarkdown::paged_table(df)` now also as 

screenshot

yihui (approved):
Great. Thanks!

cderv:
Is this a confidential feature or should we add a bullet to NEWS ? 😄

yihui:
Not confidential but I thought it was a rather minor change and probably not worth mentioning :) Please feel free to add a bullet.

PR #1930, opened 2020-10-20
Add mention of !expr use in df_print about custom function.

### LINKED ISSUE #1093: Unable to pass custom function to df_print in YAML header
Per the rmarkdown documentation, `df_print` can be passed an arbitrary function for printing. This works fine when passed directly to `render()`, but [`resolve_df_print()` inside `render()`](https://github.com/rstudio/rmarkdown/blob/master/R/render.R#L819) is not set up to handle the names of arbitrary functions passed via YAML headers in .Rmd documents. Could a case be added to the `resolve_df_print()` function to check that the passed string from YAML:

- is a valid function name
- that function returns the right kind of output

and if so, set that function to `df_print`. Step two I would imagine is optional, depending on its difficulty to define vs how you want the user to see an error. Example code for step one below. Apologies, I'm not sufficiently well versed in R development and the rmarkdown package to suggest code step two.  

    else if (is.function(get(df_print)))
            df_print <- get(df_print)

Ideally I'd like to do this with an R Notebook, but per issue #964, `df_print` has no effect when rendering an HTML notebook, which is suboptimal. Why not make `paged_table_html_asis` the default for `df_print`, instead of the exclusive, hardcoded option? 

Thanks for all your work on a great package!

cderv:
Small doc fix. Opening PR if you want to review the wording. closes #1093

I added paragraph to the doc because it is a long one and giving some space makes it easier to read.


yihui (commented):
This is great. It's not specific to `df_print`, but applies to any YAML values. We should probably have a recipe in the Cookbook, although I feel a little hesitant on it, because `!expr` is essentially `eval(parse())`, and in general, I feel scary when mentioning anything that boils down to `eval(parse()`, but you could argue that `source()` is also `eval(parse())` and we use it all the time :)

cderv:
I agree this is not specific but regarding `df_print`, I find that it was not clear how to pass a function using the yaml, when the doc says you can use a custom function. hence the precision for only that here.

>  We should probably have a recipe in the Cookbook,

It has been suggested. https://github.com/yihui/rmarkdown-cookbook/issues/277

I believe we should have a recipe about the two ways to pass R code to the yaml 
````yaml
---
title: "Test"
date: "`r Sys.Date()`"
output: 
  html_document: 
    df_print: !expr pander::pander
---
````

* Using knitr inline code to pass some code to be evaluated by knitr before pandoc rendering
* Using `!expr` to pass expression to yaml field used by rmarkdown or other tool during the rendering process.

> I feel a little hesitant on it, because !expr is essentially eval(parse()),

I agree. But we support `yaml::yaml.load(..., eval.expr = TRUE)` in **rmarkdown**, so I guess it should be documented correctly with all the warning. We could also provide an option (`rmarkdown.yaml.eval` for example), allowing anyone to deactivate this feature completly (for security consideration for example). Just an idea.

File: man/beamer_presentation.Rd [hunk 1]
```diff
@@ -59,16 +59,22 @@ individual bullet you can precede it with \code{>}. For example:
...
+  for formats that produce HTML).
+
+  In addition to the named methods you can also pass an arbitrary function to
+  be used for printing data frames. If specifying this within a YAML header,
+  arbitrary functions must be prefaced by \samp{!expr}, i.e \samp{df_print:
+  !expr knitr::kable} is equivalent to the method "kable". See
+  \code{\link[yaml:yaml.load]{yaml::yaml.load}} for details.
+
+  You can disable the \code{df_print} behavior entirely by setting the option
+  \code{rmarkdown.df_print} to \code{FALSE}.}
```
yihui, line 77: I don't think `\item{}` supports multiple paragraphs, so you need to make this a single paragraph. Perhaps just add a link to the book instead of explaining all the details here? https://bookdown.org/yihui/rmarkdown/html-document.html#data-frame-printing (this page will be updated after I merge https://github.com/rstudio/rmarkdown-book/pull/77)

yihui (commented):
> I believe we should have a recipe about the two ways to pass R code to the yaml

Yes, I agree. The distinction is quite subtle. I've used this example several times in the past: https://stackoverflow.com/q/40801633/559676

man/beamer_presentation.Rd
cderv: It outputs like this 
image

So it seemed ok to me. 

But I guess you are right, now I've suggest addition in the book I should link to that. I have done these improvments the other way around 😅

cderv: Done in https://github.com/rstudio/rmarkdown/pull/1930/commits/6961ddd3a8a0e61b2054081a3688f3406574ae85

cderv:
Reminder: To merge after https://github.com/rstudio/rmarkdown-book/pull/77

cpsievert:
This is slightly off-topic, but I'm likely going to need this `eval(parse())` trick in https://github.com/rstudio/rmarkdown/pull/1706 to support something like:

```yaml
html_document:
  theme:
      bg: "#202123"
      fg: "#B8BCC2"
      primary: "#EA80FC"
      base_font: !expr bslib::font_google('Fira Sans')
``` 

And while looking into this, I noticed the inline knitr code trick doesn't work (is there a good reason why it only works for top-level params like `date`? Or is that a bug?)

```yaml
html_document:
  theme:
      bg: "#202123"
      fg: "#B8BCC2"
      primary: "#EA80FC"
      base_font: "`r bslib::font_google('Fira Sans')`"
```

cderv:
Hi @cpsievert , 

This `!expr` trick is from **yaml** package and it will work for any fields as long as we support `eval.expr = TRUE` in `yaml::yaml.load()`. (which we do). It is very useful but as Yihui said it is evaluating, so it is always something to be aware of and I guess that is why the change to FALSE by default in **yaml**. (which we change to TRUE in **rmarkdown**). 

So, the trick you need using `!expr` will work as is without needing anything else apart from better documentation on it. 
Which currently is very few as it is considered an advanced usage because one need to understand YAML and know this will be evaluated only by R when the yaml header is imported - pandoc will see it as is,  with `!expr`, and not the evaluted result. For example, this won't give te expected result: 
```yaml
---
title: "inline R chunk"
date: !expr Sys.Date()
output:
  html_document: default
---
```
That is why we are trying at best to provide new fields in YAML to pass values instead of relying to this `!expr` feature. But for parameter that needs function (like df_print) it is a must have.

This is the main difference with using "knitr way" in YAML header: These are inline R chunks, which are processed by knitr when creating md from Rmd so the YAML header will contain the resulting value that Pandoc could access when importing the metadata. (That is why it is used in fields like date or title, or any other Pandoc variables for templates.). 

As with any knitr inline R code, code must evaluated to character (or be coercible to) so that it is inserted as text in the document. I don't think `bslib::font_google('Fira Sans')` is a correct candidate to be used in a inline R chunk as it is a list, and I believe you need this full list in `base_font`. It is not a yaml header issue as it would not work either as an inline chunk in the body of an Rmd. The error you get should indicates that inline result are coerced to character.

So yes, using knitr inline R code won't work for your usage. Only yaml `!expr` feature will be useful. 

I don't really know about **bslib** yet or `font_google`, but what about having some more levels ? 
```yaml
html_document:
  theme:
      bg: "#202123"
      fg: "#B8BCC2"
      primary: "#EA80FC"
      base_font: 
          font_google: 'Fira Sans'
---
```

You would get this value to pass to `font_google` and not rely on users knowing how to use `!expr` ? 

Anyway, this is not the place to discuss this. Did I answer your question in the first place ?

cpsievert:
> Did I answer your question in the first place ?

Yes, thanks!

yihui (approved):
Thanks!

PR #1931, opened 2020-10-22
pkg_file_lua now works with devtools

cderv:
The fact that it does not work with `devtools::load_all`, and so **testthat** is a bit annoying. 

I reused a trick from **pkgdown** to make it work - hope it is fine. 

With this PR, this now works from **bookdown*** for example

```r
devtools::load_all()
lua_filter()
```


yihui (approved):
Thanks!

PR #1932, opened 2020-10-23
fix exceptions on pandoc_convert(citeproc = TRUE)

atusy:
When citeproc is TRUE, but natbib or biblatex are under use, they are removed from the `options` variable.
However, this variable is no more used after the modification.
Instead, we should modify the `args` variable.

https://github.com/rstudio/rmarkdown/blob/bd2369f9f965c744588a0b056b35dd92e47902cb/R/pandoc.R#L83-L89

Note that the modification of `option` variable was valid when the exception is initially introduced by https://github.com/rstudio/rmarkdown/commit/bbabea82c1bfd3461fc7f675ac674c21560654db

### DIFF
--- a/R/pandoc.R
+++ b/R/pandoc.R
@@ -83,7 +83,7 @@ pandoc_convert <- function(input,
   # citeproc filter if requested
   if (citeproc) {
     # --natbib/--biblatex conflicts with '--filter pandoc-citeproc'
-    args <- c(args[!args %in% c("--natbib", "--biblatex")],
+    args <- c(setdiff(args, c("--natbib", "--biblatex")),
               pandoc_citeproc_args())
   }
 
--- a/R/pandoc.R
+++ b/R/pandoc.R
@@ -81,9 +81,10 @@ pandoc_convert <- function(input,
   args <- c(args, options)
 
   # citeproc filter if requested
-  if (citeproc && !any(c("--natbib", "--biblatex") %in% options)) {
+  if (citeproc) {
     # --natbib/--biblatex conflicts with '--filter pandoc-citeproc'
-    args <- c(args, pandoc_citeproc_args())
+    args <- c(args[!args %in% c("--natbib", "--biblatex")],
+              pandoc_citeproc_args())
   }
 
   # build the conversion command
--- a/R/pandoc.R
+++ b/R/pandoc.R
@@ -81,11 +81,9 @@ pandoc_convert <- function(input,
   args <- c(args, options)
 
   # citeproc filter if requested
-  if (citeproc) {
-    args <- c(args, pandoc_citeproc_args())
+  if (citeproc && !any(c("--natbib", "--biblatex") %in% options)) {
     # --natbib/--biblatex conflicts with '--filter pandoc-citeproc'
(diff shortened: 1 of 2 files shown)


R/pandoc.R
```diff
@@ -82,10 +82,9 @@ pandoc_convert <- function(input,
 
   # citeproc filter if requested
   if (citeproc) {
-    args <- c(args, pandoc_citeproc_args())
     # --natbib/--biblatex conflicts with '--filter pandoc-citeproc'
-    i <- stats::na.omit(match(c("--natbib", "--biblatex"), options))
-    if (length(i)) options <- options[-i]
+    args <- c(args[!args %in% c("--natbib", "--biblatex")],
+              pandoc_citeproc_args())
```
yihui, line 87: Is it a little cleaner to use `setdiff()`?

```suggestion
    args <- c(setdiff(args, c("--natbib", "--biblatex")),
              pandoc_citeproc_args())
```

yihui (approved):
Thanks a lot for catching this problem!

R/pandoc.R
atusy: Absolutely!!

cderv:
This look good to me too. 

However, I wondered why we did not have any issues with this because this is supposed to not work anymore since this change: https://github.com/rstudio/rmarkdown/commit/07da2572770568234b89ee2951a0b592f46e490f ... in 2016 !! 

At first I guess **pandoc** itself knows what to do when both are provided has it does not error when you provide both. 

But I think this part of code is never used because we already set `citeproc = FALSE` for `pandoc_convert()` iif `--natbib` or `--biblatex` is present in the option

* We set the argument in `pdf_document()` if natbib or biblatex asked
https://github.com/rstudio/rmarkdown/blob/2b924304485d474b110b322df472541aa8bcac2f/R/pdf_document.R#L265-L273
* We check in `render()` if an argument is present
https://github.com/rstudio/rmarkdown/blob/4e1661dde00f844e0d8d6e860bdf9643b177bbe8/R/render.R#L838-L839
* We check in `render()` if citation processing (no matter which) is required
https://github.com/rstudio/rmarkdown/blob/4e1661dde00f844e0d8d6e860bdf9643b177bbe8/R/render.R#L942
* For pdf, we set `citeproc` for `pandoc_convert` to FALSE if `need_bibtex` is TRUE
https://github.com/rstudio/rmarkdown/blob/4e1661dde00f844e0d8d6e860bdf9643b177bbe8/R/render.R#L946-L948

IMO, for Rmarkdown internal behavior that means `--biblatex` or `--natbib` can't be in the `args` if `citeproc`  is TRUE for `pandoc_convert`. 

That is why we did not have any errors on this despite this code being incorrect since 2016.

@atusy, just to be sure, what made you spot this ? did you encountered an issue ?

atusy:
@cderv I did not face any issue by running `pandoc_convert`.
I am currently implementing citation feature on https://github.com/atusy/ftExtra by using Pandoc.
For that, I took a look at how `pandoc_citeproc_args` is used in rmarkdown package, and noticed the issue.

cderv:
Great, awesome ! If you had not seen it, we would have never found this incorrect piece of code as it does not cause error in **rmarkdown** workflow.

Thank you !

Issue #580, opened 2015-12-28
New update deletes .bib file


RobertMyles:
Hi RMarkdown folks,

I just updated to the new version (9) and now something rather strange is happening. Everytime I knit a pdf, it seems to delete the .bib file and the .csl file that I was using (both named in the YAML header). Quite naturally, I thought I must be imagining things, but this happened on two separate computers, one a Mac and one using Windows, about six or seven times now. It was fine before (the pdf still gets produced, btw). I know it sounds like the craziest thing, but...

yihui:
I cannot reproduce this issue. Could you provide a small reproducible example? Mine is below:

test.Rmd

```
---
output: pdf_document
bibliography: test.bib
---

@R-base is cool.
```

test.bib

```
@Manual{R-base,
  title = {R: A Language and Environment for Statistical Computing},
  author = {{R Core Team}},
  organization = {R Foundation for Statistical Computing},
  address = {Vienna, Austria},
  year = {2015},
  url = {https://www.R-project.org/},
}
```

BTW, do you have `latexmk` installed, and what is your LaTeX version? You may check them in R:

```r
system('latexmk -v')
system('pdflatex -v')
```

RobertMyles:
Hi Yihui,

Thanks for the quick reply. When I run a small test example like yours above, this error doesn't happen. However, when I use the full document I was using earlier (I'm writing a phd thesis in RMarkdown), then it happens every time, deleting both the .bib file and the .csl I'm using (on both mac and Windows). I didn't include a reproducible example earlier because this was all working fine until I updated RMarkdown on both the computers. As you can imagine, having to redo my entire bibliography in the last few weeks of my phd is making me 'slightly' mad.

The versions of latexmk and pdflatex are:

Latexmk, John Collins, 5 February 2015. Version 4.43a
pdfTeX 3.14159265-2.6-1.40.16 (TeX Live 2015)

Is there any other information I could include to help us figure out the problem?

yihui:
[...]

PR #1933, opened 2020-10-23
never use the original input directory as the intermediate directory ever

### LINKED ISSUE #1248: using intermediates_dir deletes sourced file
If I source a file in markdown and use the intermediates_dir option when rendering, the sourced file gets deleted.

Simple example.Rmd: 

````
```{r example}
source("blank.R") 
```
````

then:

```r
rmarkdown::render("example.Rmd",intermediates_dir=".")
```

blank.R is deleted. If I do not use intermediates_dir, it does not get deleted.

I am using R-3.4.0 and I installed rmarkdown from devtools before submitting this issue.

cderv:
Fixes #1248, that was a more generic issue of #580, fixed by #582. 

So I think we can apply the same fix but to all types of input, not just inputs with shell chars.

More precisely, if intermediates_dir is set to "." or set to the same as input directory, it will be set to NULL. I added a test for this specific case.


yihui (approved):
Looks good. Thanks!

Issue #1308, opened 2018-03-30
render() fails to delete intermediate tex file if called from non-input working directory


vqv:
`render()` fails to delete the intermediate tex file (when `keep_tex = FALSE`) if it is called from a working directory which is different from that of the `input`.  It appears that there is a pseudo-race condition:

https://github.com/rstudio/rmarkdown/blob/8cfc8783636b37876a6dc929dac2395df080e7ac/R/render.R#L154

is called before

https://github.com/rstudio/rmarkdown/blob/8cfc8783636b37876a6dc929dac2395df080e7ac/R/render.R#L671

So by the time the exit hook evaluates `unlink(texfile)`, the working directory has already been changed back.

jtr13:
Thanks for explaining what the issue was... that stubborn .tex file was driving me nuts.

likanzhan:
Has this problem been resolved? Because I still got the same issue.

cderv:
Here is a reproducible example 
``` r
dir.create(tmp_dir <- tempfile())
old <- setwd(tmp_dir)
dir.create("rmd")
xfun::write_utf8("# just a title",
                 "rmd/test.Rmd")
fs::dir_ls(recurse = TRUE)
#> rmd          rmd/test.Rmd
rmarkdown::render("rmd/test.Rmd", output_format = rmarkdown::pdf_document(keep_tex = FALSE), quiet = TRUE)
fs::dir_ls(recurse = TRUE)
#> rmd          rmd/test.pdf rmd/test.Rmd rmd/test.tex
setwd(old)
unlink(tmp_dir)
```

<sup>Created on 2020-10-26 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

This is indeed related to the path of the tex file.

cderv:
@vqv thanks for the bug report ! Finally, it should be fixed now.

PR #1935, opened 2020-10-26
correctly delete tex file if keep_tex = FALSE

cderv:
This will fix https://github.com/rstudio/rmarkdown/issues/1308

File was not deleted because of relative path and sometimes, `unlink()` is run from another working dir than the input directory.


yihui (approved):
Perfect. Thanks!

Issue #1364, opened 2018-06-03
Bibliography not rendered if given in YAML other than main header


bwiernik:
My package includes a function to programmatically build and render an RMarkdown file, including a bibliography.

If I place the `bibliography: ` parameter in the main YAML header of the document, it renders correctly (see 
[rmarkdown_bib_header_test.zip](https://github.com/rstudio/rmarkdown/files/2065823/rmarkdown_bib_header_test.zip) for .R, .Rmd, and .bib files for this examples).

### Renders correctly:
```r
rmarkdown::render("bib_in_header.rmd")
```

However, if I place the `bibliography: ` parameter in a YAML block outside of the main header, rmarkdown doesn't pick this up and does not render the citations or bibliography.

### .bib file isn't found:
```r
rmarkdown::render("bib_not_in_header.rmd")
```

Other YAML metadata doesn't seem to behave like this (e.g., `nocite`, `csl`, and even `title` can all be placed in other YAML blocks and work correctly), so this seems like a bug. It would be really helpful if rmarkdown could be adjusted to accept `bibliography: ` metadata given outside the main header, as that would make it easier for users of my package to place the output of this function inside larger RMarkdown documents.

cderv:
This issue is related to the way **rmarkdown** is getting the metadata: it uses a yaml frontmatter only . It will only read meta from a top yaml header, and not yaml chunk inside the document. 

Pandoc however can handle metadata provided in several yaml blocks. That is why it works for fields used only by Pandoc. It does not work for fields that **rmarkdown** will use itself, like `bibliography`

Related to https://github.com/rstudio/rmarkdown/issues/1891

cderv:
@bwiernik we finally made a change that should answer your need. Pandoc citation processing will now be enable also if `bibliography:` is in another yaml block in the document in this form: 
```yaml
bibliography:
  - reference_list.bib
```

bwiernik:
Awesome thank you!

Issue #1891, opened 2020-09-05
Inconsistent behaviour for multiple YAML metadata blocks


fkohrt:
The [Pandoc MANUAL](https://pandoc.org/MANUAL.html#extension-yaml_metadata_block) says:

> A document may contain multiple metadata blocks. If two metadata blocks attempt to set the same field, the value from the second block will be taken.

This is the behaviour as of Pandoc 2.8 (jgm/pandoc@9c7f75a), earlier versions gave preference to value from the first block. But it appears this change has not been implemented in the `rmarkdown::metadata` object, as the example below demonstrates:

````Rmd
---
title: "Title 1"
output: html_document
---

```{r, echo = FALSE, results = 'asis'}
cat(rmarkdown::metadata$title)
```

---
title: "Title 2"
---

````

This leads to the following document:

![metadata](https://user-images.githubusercontent.com/12914806/92313704-f753ff00-efce-11ea-92f6-95156708b852.png)

Note that the title of the document is set according to the second YAML metadata block, but the `rmarkdown::metadata` object reports a different title.

<details>
<summary><code>xfun::session_info('rmarkdown')</code></summary>

```
R version 3.6.3 (2020-02-29)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 18.04.4 LTS, RStudio 1.2.5001

Locale:
  LC_CTYPE=de_DE.UTF-8       LC_NUMERIC=C               LC_TIME=de_DE.UTF-8        LC_COLLATE=de_DE.UTF-8    
  LC_MONETARY=de_DE.UTF-8    LC_MESSAGES=en_US.UTF-8    LC_PAPER=en_US.UTF-8       LC_NAME=C                 
  LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

Package version:
  base64enc_0.1.3 digest_0.6.25   evaluate_0.14   glue_1.4.2      graphics_3.6.3  grDevices_3.6.3 highr_0.8      
  htmltools_0.5.0 jsonlite_1.7.0  knitr_1.29      magrittr_1.5    markdown_1.1    methods_3.6.3   mime_0.9       
  rlang_0.4.7     rmarkdown_2.3.3 stats_3.6.3     stringi_1.4.6   stringr_1.4.0   tinytex_0.25    tools_3.6.3    
[...]

PR #1936, opened 2020-10-27
detect bibliography field outside of yaml header as Pandoc supports

cderv:
This would fix https://github.com/rstudio/rmarkdown/issues/1364 

This applies the same logic as for a `references` yaml block defined elsewhere in the document. 

It is a temporary fix while we are only parsing the yaml header (https://github.com/rstudio/rmarkdown/issues/1891)

We would only allow blocks of the form
```yaml
bibliography:
  - reference_list.bib
```

so that it is easier to match exactly but we could also use 
```r
length(grep("^bibliography\\:", input_lines)) > 0
```
so that this would also work
```yaml
bibliography: reference_list.bib
```

@yihui I let you see if that is a good fix or if we should wait for better multi yaml block support. Or if we won't support this.


yihui (approved):
Looks good. Thanks!

PR #1941, opened 2020-11-04
Add more highlighting support for HTML output

### LINKED ISSUE #1881: Make it easy to use downlit from html_document()
So you can easily get R powered syntax highlighting and autolinking.

(And probably also worth supporting in `github_document()` and `html_document()`)

cderv:
This PR aim to close #1881 by adding an option support for **downlit** highlighting using a new argument in the format, named `highlight_downlit` after the one chose in **distill**

#### Support in `html_document` 

* post processing step on the resulting html document
* All document based on `html_document` could benefit from this by activating the feature. 

- [x]  Test it is working in gitbook)
- [x] Choose the default highlight theme (see below)

#### Support in markdown output 

* `downlit::downlit_md_path()` could be used, but only support Chroma for now
* Also it adds span tags for highlighting but we want only autolink ? 🤔 Pandoc may remove them depending on the extensions, so pre process ? 

- [x] ~Support in `github_document`~
- [x] ~Support in `md_document`~

Edit: Won't add it here. See https://github.com/rstudio/rmarkdown/issues/1881#issuecomment-722284965

### About highlighting

* By default, **rmarkdown** uses highlightjs with the default them in `html_document`
* If downlit is enabled, default (`highlight = "default"`) will be a pandoc style and not highlightjs.
* downlit requires to add a specific css code so that the code has links but the highligting stay the one expected from pandoc the highlight.

- [x] improve documentation regarding highlighting, and explain difference between engine and color scheme. (see https://github.com/rstudio/rmarkdown/pull/1941#discussion_r518031334)
- [x] Update the `a11y.theme` file after its update (see https://github.com/rstudio/rmarkdown/pull/1941#discussion_r518029940)


cderv:
default syntax highlighting **when downlit is used** is now a11y.theme like in **distill** and `bookdown::bs4_book()`

R/html_document.R
```diff
@@ -434,6 +449,20 @@ html_document <- function(toc = FALSE,
     args
   }
 
+  # post-processor that use the output file from pandoc
+  post_processor <- function(metadata, input_file, output_file, clean, verbose) {
```
hadley, line 453: This won't work with output formats that extend `html_document()` but just replace the post processor (instead of wrapping it). I don't know if that's common but I think it's worth noting and thinking about.

R/pandoc.R
```diff
@@ -551,25 +551,34 @@ unix_mathjax_path <- function() {
 
 
 pandoc_html_highlight_args <- function(template,
```
hadley, line 553: I think this function would be marginally more clear if `highlight` was the first argument because is `template` only used when `highlight` is "default" (and then you could make it `template = "default"`)

tests/testthat/test-pandoc.R
```diff
@@ -0,0 +1,35 @@
...
+    else
+      c("--highlight-style", name)
+  }
+  downlit  <- pandoc_variable_arg("highlight-downlit=1")
+  highlightjs <- pandoc_variable_arg("highlightjs=1")
+  a11y_theme <- highlight_style(pkg_file_highlight("a11y.theme"))
+  # check logic
+  expect_equal(pandoc_html_highlight_args("default", NULL), highlight_style())
+  expect_equal(pandoc_html_highlight_args("dummy.html", NULL), highlight_style())
+  expect_equal(pandoc_html_highlight_args("dummy.html", "default"),
```
hadley, line 15: The wrapping makes it a little harder to see that you've covered all combinations of cases. It _might_ be worth making a local helper for `pandoc_html_highlight_args` that isn't quite so long. I also suspect re-ordering the arguments as I suggest above will make the logic more clear.

R/html_document.R
```diff
@@ -193,6 +197,7 @@ html_document <- function(toc = FALSE,
                           self_contained = TRUE,
                           theme = "default",
                           highlight = "default",
+                          highlight_downlit = FALSE,
```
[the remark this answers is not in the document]
cderv: yes that is the reason. Initially, I would have called `use_downlit` to not get the `highlight` word used twice but I stayed with the choice made for distill which choose `highlight_downlit`. 

I will improve the documentation regarding highlighting. 👍

R/pandoc.R
```diff
@@ -551,25 +551,34 @@ unix_mathjax_path <- function() {
...
-  }
-  else if (!identical(template, "default")) {
-    if (identical(highlight, "default"))
-      highlight <- "pygments"
+  } else if (highlight_downlit) {
+    # use pandoc highlight and add custom css
+    if (is_highlightjs(highlight)) {
+      # default to accessible theme a11y - copied from distill
+      # https://github.com/rstudio/distill/blob/c98d332192ff75f268ddf69bddace34e4db6d89b/inst/rmarkdown/templates/distill_article/resources/a11y.theme
+      highlight <- pkg_file_highlight("a11y.theme")
```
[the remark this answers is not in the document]
cderv: Yes I think we could. 
Would it be interesting to add the Rstudio theme also added in `distill` ?

inst/rmarkdown/highlight/a11y.theme
```diff
@@ -0,0 +1,211 @@
+{
+    "text-color": null,
```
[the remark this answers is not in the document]
cderv: Good to know ! 
Would it be a good idea to store those `.theme` (and `.css`  equivalent ?) somewhere online ? That would be a lot easier to maintain across packages, using a script to use in dev to update the file. But warning each other of an update works.

(Or this is maybe the kind of resource to live in one package as a dependency for other to use to find the file. )

R/html_document.R
cderv: you mean a format doing something like 

```r
new_format <- function(...) {
    base_format <- html_document(...)
    base_format$post_processor <- new_postproc() 
    base_format
}
```` 
instead of `output_format` and its `base_format` argument ?

File: inst/rmarkdown/highlight/a11y.theme (hunk 6 above)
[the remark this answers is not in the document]
Comment (hadley) [reply inst/rmarkdown/highlight/a11y.theme:2]: Well there are online in the GitHub repo 😄. But there's some chance we'll move into bslib, since there's general need to match the syntax highlight colour scheme to the rest of the template.

File: R/html_document.R (hunk 1 above)
Comment (hadley) [reply R/html_document.R:453]: Right.

**Comment by cderv** [comment]
Just a not that this is also working with `gitbook` as `highlight_downlit = TRUE` will be correctly pass to `html_document`

File: R/pandoc.R (hunk 5 above)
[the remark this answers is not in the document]
Comment (cderv) [reply R/pandoc.R]: While rewrite keeping the same logic, I added the possibility to use `highlight = "a11y"`

File: inst/rmarkdown/highlight/a11y.theme (hunk 6 above)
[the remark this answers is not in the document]
Comment (cderv) [reply inst/rmarkdown/highlight/a11y.theme:2]: It still feel weird to have pandoc highlight files stored in `bslib`. Maybe not all format that will use theme will use bslib. 
Maybe they should live in the "pandoc" package in the future.

Will see.

**Comment by cderv** [comment]
All this highlighting stuff started by downlit support in rmarkdown made me think. I found it was not clear enough in the current code, so I rewrote the whole function on the base of highlighting engine + syntax highlight theme. 

* Same logic as before - I added some tests to check that
* Rewrote the if-else code to be better align with engine + theme
* Added support for `*.theme` file passed in the `highlight = ` arguement (no more `match.args` on pandoc highgligth style name) 
* Added support for `highlight = "a11y"` and `highlight = "rstudio"`, based on distill `.theme` file
* This allow also to be more aligned with recent change in distill (we may be able to refactor as distill uses `html_document`)

I did this only for the html format, but this could be apply also to other format (not downlit, but the use of `.theme` file). 

I still need to update the documentation and look at the other reviews.

File: R/html_document.R [hunk 7]
```diff
@@ -52,6 +52,10 @@
 #'@param highlight Syntax highlighting style. Supported styles include
 #'  "default", "tango", "pygments", "kate", "monochrome", "espresso", "zenburn",
 #'  "haddock", and "textmate". Pass \code{NULL} to prevent syntax highlighting.
+#'@param highlight_downlit \code{TRUE} to use the \pkg{downlit} package to
+#'  highlight R code (including providing hyperlinks to function documentation).
```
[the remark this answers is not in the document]
cderv: Done in 509ca7a

R/html_document.R
```diff
@@ -52,6 +52,13 @@
 #'@param highlight Syntax highlighting style. Supported styles include
 #'  "default", "tango", "pygments", "kate", "monochrome", "espresso", "zenburn",
 #'  "haddock", and "textmate". Pass \code{NULL} to prevent syntax highlighting.
+#'@param highlight_downlit \code{TRUE} to use the \pkg{downlit} package as
+#'  syntax highlighting engine to highlight inline code and R code chunks
+#'  (including providing hyperlinks to function documentation). Only Pandoc
+#'  color schemes are supported with this engine. With \code{highlight}
+#'  "default", it will use the accessible theme called "a11y". To know more
```
hadley, line 65:
```suggestion
#'  "default", it will use the accessible theme called "a11y". To learn more
```

R/html_document.R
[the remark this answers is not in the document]
cderv: Documented in b6e5c4c

R/html_document.R
```diff
@@ -49,9 +49,22 @@
 #'  "darkly", "readable", "spacelab", "united", "cosmo", "lumen", "paper",
 #'  "sandstone", "simplex", or "yeti"). Pass \code{NULL} for no theme (in this
 #'  case you can use the \code{css} parameter to add your own styles).
-#'@param highlight Syntax highlighting style. Supported styles include
-#'  "default", "tango", "pygments", "kate", "monochrome", "espresso", "zenburn",
-#'  "haddock", and "textmate". Pass \code{NULL} to prevent syntax highlighting.
+#'@param highlight Syntax highlight engine and style. "default" (and "textmate")
+#'  will select highlightjs as syntax highlight engine instead of Pandoc.\cr Any
```
hadley, line 53: Why do you use `\cr` here? (instead of just a newline)

hadley, line 53:
```suggestion
#'  will use highlightjs as syntax highlighting engine instead of Pandoc.\cr Any
```

R/html_document.R
```diff
@@ -49,9 +49,22 @@
...
+#'  other value will be passed as
+#'  \href{https://pandoc.org/MANUAL.html#syntax-highlighting}{highlight style to
+#'  Pandoc}. Pandoc's supported styles include "tango", "pygments", "kate",
```
hadley, line 56:
```suggestion
#'  other value will be passed to Pandoc's
#'  \href{https://pandoc.org/MANUAL.html#syntax-highlighting}{highlighting style}. 
#' Pandoc's built-in styles include "tango", "pygments", "kate",
```

R/pandoc.R
```diff
@@ -551,34 +551,58 @@ unix_mathjax_path <- function() {
...
+  if (highlight_downlit) {
     if (is_highlightjs(highlight)) {
-      args <- c(args, "--no-highlight")
-      args <- c(args, "--variable", "highlightjs=1")
-    }
-    else {
-      args <- c(args, "--highlight-style", highlight)
+      stop(
+        sprintf(
+          "%s theme is for highlightjs highlighting engine",
```
hadley, line 581:
```suggestion
          "'%s' theme is for highlightjs highlighting engine",
```

R/pandoc.R
```diff
@@ -551,34 +551,58 @@ unix_mathjax_path <- function() {
...
+          "and can't be used with downlit.",  highlight
+        ),
+        call. = FALSE
+      )
     }
+    args <- c(
+      pandoc_highlight_args(highlight, default = resolve_highlight("a11y")),
+      pandoc_variable_arg("highlight-downlit")
```
hadley, line 589: It might worth a comment that these variables are included primarily for testing?

hadley (commented):
This is looking great!

R/html_document.R
cderv: If someone overwrite the post processor AND use `highlight_downlit = TRUE`, there won't be any downlit processing but the highlight would still default to "a11y" and not highlightjs. But not sure we can detect this specific case.
I am not sure to see what we could do here. If someone extend `html_document` but replace the post processor. Currently, `html_document_base` post processor will also be replaced - and some feature like `copy_resource` are lost. 
And if some extend `html_document` AND use `highlight_downlit = TRUE`, we could think the post processor would have been kept. 

I believe, if someone extend on the feature of `html_document`, it would be the responsibility of the author to keep the post processor if needed. Different formats not using `output_formats` mechanism are doing it already with : 
```r
new_format <- function(...) {
    base_format <- html_document(...)
    post <- base_format$post_processor
    base_format$post_processor <- function(..., output_file) {
        if (is.function(post)) output_file <- post(...)
        new_proc(..., output_file)
    }
    base_format
}
```

I believe a vignette (or [recipe](https://github.com/yihui/rmarkdown-cookbook/issues/285)) on how to extend rmarkdown format would be necessary to share the good practices and the pitfall to watch out for.

R/html_document.R
cderv: Because newline won't create a newline in the html help page. I believe you need `\cr` to break a line. (unless you are using markdown and you leave two space ?

```r
#' @param highlight Syntax highlight engine and style. "default" (and "textmate")
#'  will select highlightjs as syntax highlight engine instead of Pandoc.
# ' Any other value
```
will lead to no break line between "Pandoc" and "Any". 

Am I missing something ?

R/pandoc.R
cderv: For testing ? It is not only for testing.

`highlight-downlit` is a pandoc variable I need to introduced in order to have a piece of css inserted (same mechanism as for highlightjs). 

Here : https://github.com/rstudio/rmarkdown/pull/1941/files#diff-91b59569aab1b937e1e5a6e1559a62de19e04f4c4bad52bfb08b910281e0f2dfR65

The css is for having the color scheme of highlighting style applies on links created by downlit. Otherwise, they are like any other hyperlinks (blue and underlined)

It is a small piece of css but it seemed necessary to a good visual output.

but I could add a comment about this.

R/html_document.R
hadley: Can't you do this?

```
#' @param highlight Syntax highlight engine and style. "default" (and "textmate")
#'  will select highlightjs as syntax highlight engine instead of Pandoc.
#'  
#'  Any other value
```

I'm pretty sure that works in plain Rd.

R/html_document.R
hadley: I agree with your points; I think it's important just to note that this has the potential to break downstream formats. I don't think this is particularly important here, but I think it's the sort of analysis that's necessary when making changes to old and popular functions.

R/pandoc.R
hadley: Ah ok, I missed that. I think it's probably important to note somewhere that `downlit_highlight` will insert  new css, but it's not clear to me where that should be.

R/html_document.R
cderv: Oh yes I can do that with en empty line, but that would make several paragraphs. It does not seem to be the style in the package currently. Every params in help file are in one paragraph. 

Let's begin to do that then - It is easier to read.

cderv: Done in 7f11dc9

R/pandoc.R
cderv: It may be worth adding a section on highlighting in the help page. That would allow to give this sort of details about 

* highlightjs which insert css and js in the template
* Custom theme included in the package in `.theme` file
* downlit which will replace pandoc highlighting modifying the html nodes and inserting css in the template

It is better than nowhere.

R/html_document.R
cderv: You're right, we need to check that closely. But I am worried we do not have any counter measure if you extend based on a format but replace its internal processing. That is why the `output_formats` function exists and is exported in the first place I believe, because it contains a merging strategy with the base format provided (so you can't overwrite by mistake)

R/pandoc.R
cderv: Added a whole section on Highlighting if you want to look at it: https://github.com/rstudio/rmarkdown/pull/1941/files#diff-93d1e6747a8a5a21e4668d9a67b3889d481426747e8a0f18d7b2ebfb32ee8413R101

cderv:
I now have documented everything, new optional downlit support but also what a user should expect from the current value of `highlights`, including mention of highlightjs. I took the opportunity to warn that the default could change in future version. 

Also, you can now pass a theme file now to `highlight` argument in `html_document`. This is used to support `"a11y"` theme and `"rstudio"` theme. I just need maybe to check pandoc version and add a `pandoc_available()` as I believe [this is a Pandoc 2 only feature](https://pandoc.org/releases.html#pandoc-2.0-2017-10-29) and error or fallback accordingly. (always the same need to support very old version). 

After this PR we could make `"a11y"` theme and `"rstudio"` theme available to other format too.

@yihui it is not ready to merge yet because of this Pandoc-2-only thing, but if you haven't look at it yet, you can review from now so that I include your feedbacks as soon as possible.

cderv:
Just waiting for the new updated "a11y.theme" before this PR is ready to merge. I let it as draft until then. 
@apreshill waiting for your signal 😉

cderv:
This PR is now quite ready. 

I started testing with different output formats. This now works correctly with `html_document` and `bookdown::html_document2`. It seems there is an issue with `bookdown::gitbook()` 🤔 - looking into this.

cderv:
Ok this is because `downlit::downlit_html_path()` applied before splitting will mess up with the bookdown's special tokens used for splitting the book. `xml2::read_html()` then `xml2::write_html()` will not re-write in the same format as it has read. Newlines are messed up and special comments tokens won't be on there own line anymore, which will prevent bookdown for splitting the book correctly. 

No issue with `bs4_book()` because every tweaks happens after splitting, include applying `downlit::downlit_html_path()`.

cderv:
If we want to support downlit in `bookdown::gitbook()` it will require another process. Currently, I'll need to deactivate it : `highlight_downlit = TRUE` can't be used in `bookdown::gitbook()`

With `pagedown::html_paged()`, the links in autolinking does not render correctly. This is because links are dealt specifically so that they are correctly printed. downlit autolinking makes no sense with printing. So I guess downlit should get deactivated (or autolinking only if possible)

With `xaringan::moon_reader()`, nothing is done as highlighting is done inside xaringan (`highlight = NULL` in `html_document`

With `learnr::tutorial`, it doesn't work either - `highlight_downlit = TRUE` will cause error. I believe same issue than with `bookdown` - the html rewritten after downlit post processing will mess up the formatting and parts with `<!--/html_preserve-->` won't be parsed correctly by shiny prerendered which happens after downlit post processing.

It seems post processing output html with `downlit_html_path` is not in fact the correct way to go for downlit support in `html_document` to be enable 'as is' in others formats. The other solution was `downlit_md_path` to process before pandoc and apply downlit only the code part using pandoc AST. could have been more generic.

R/html_document.R
```diff
@@ -49,9 +49,30 @@
...
+#'  engine instead of Pandoc.
+#'
+#'  Any other value will be passed to Pandoc's
+#'  \href{https://pandoc.org/MANUAL.html#syntax-highlighting}{highlighting
+#'  style}. Pandoc's built-in styles include "tango", "pygments", "kate",
+#'  "monochrome", "espresso", "zenburn", "haddock" and "breezedark". Two custom
+#'   are also included, "arrow", for accessible color scheme and "rstudio",
+#'  to replicate RStudio editor theme. A path to a \samp{.theme} file can also be
+#'  passed to be used as a custom highlighting style for Pandoc. Note that custom
+#'  style are available for Pandoc 2.0+.
```
hadley, line 65:
```suggestion
#'   are also included: "arrow", an accessible color scheme, and "rstudio",
#'   which mimics the default IDE theme. 
#' 
#'  Alternatively, supply a path to a \samp{.theme} to use a custom Pandoc style.
#'  (This requires Pandoc 2.0+)

File: R/html_document.R [hunk 14]
```diff
@@ -49,9 +49,30 @@
...
+#'
+#'  Pass \code{NULL} to prevent syntax highlighting.
+#'
+#'@param highlight_downlit \code{TRUE} to use the \pkg{downlit} package as
```
hadley, line 69: Mention that this requires downlit to be installed?

R/html_document.R
```diff
@@ -49,9 +49,30 @@
...
+#'  syntax highlight engine to highlight inline code and R code chunks
+#'  (including providing hyperlinks to function documentation). Only Pandoc
+#'  color schemes are supported with this engine. With \code{highlight}
```
hadley, line 72: Maybe: If \code{highlight = "default"}, ...

R/html_document.R
```diff
@@ -78,6 +99,54 @@
 #'@param extra_dependencies,... Additional function arguments to pass to the
 #'  base R Markdown HTML output formatter \code{\link{html_document_base}}
 #'@return R Markdown output format to pass to \code{\link{render}}
+#'@section Highlighting:
+#'
+#'  Different highlighting engines can be used with a R Markdown HTML documents:
```
hadley, line 104: I suggest making this a bulleted list. Maybe something like:

There are three highlighting engines available to HTML documents:

* highlightjs, which does highlighting in the browser, using javascript.
* pandoc, which uses pandoc
* downlite, which uses an R package (and only applies to R code)

R/html_document.R
```diff
@@ -78,6 +99,54 @@
...
+#'  highlightjs and Pandoc for all languages they support, \pkg{downlit} only
+#'  for R code.
+#'
+#'  highlightjs engine can only be used with the default template (i.e
+#'  \code{template = "default"}) and it has two styles ("default" and
+#'  "textmate"). When activated, a JS script and a CSS file are inserted into
+#'  the output HTML to make it works. For now, this is the default engine for
```
hadley, line 111:
```suggestion
#'  "textmate"). When activated, it adds two additional dependencies to the
#'  output file: a JS script and a CSS file. For now, this is the default engine for
```

R/html_document.R
```diff
@@ -78,6 +99,54 @@
...
+#'  the default template - this could change in the future.
+#'
+#'  Pandoc built-in highlighting engine works with any template, default or
```
hadley, line 114: Either "The pandoc" or "pandoc's"

R/html_document.R
```diff
@@ -78,6 +99,54 @@
...
+#'  custom, and style can be chosen among the built-in ones ("tango", "pygments",
+#'  "kate", "monochrome", "espresso", "zenburn", "haddock" and "breezedark") or
+#'  among some custom theme passed as ".theme" files (see Details in the
```
hadley, line 117:
```suggestion
#'  a path to a custom theme ".theme" file (see Details in the
```

R/html_document.R
```diff
@@ -78,6 +99,54 @@
...
+#'  \href{https://pandoc.org/MANUAL.html#syntax-highlighting}{Pandoc Manual}).
+#'  \pkg{rmarkdown} includes two custom themes to select with \code{highlight}
+#'  parameter:
+#'  \itemize{
+#'  \item{"arrow", an accessible color theme derived from \href{https://www.a11yproject.com/}{optimized for
```
hadley, line 122: Not sure it's still true to say is derived from the a11y project?

R/html_document.R
```diff
@@ -78,6 +99,54 @@
...
+#'  accessibility and color contrast}}
+#'  \item{"rstudio", a color scheme close to RStudio's default highlighting and
+#'  highglightjs's textmate.}
+#'  }
+#'  Custom theme are only available for Pandoc 2.0 and above.
```
hadley, line 127:
```suggestion
#'  Custom themes are only available for Pandoc 2.0 and above.
```

R/html_document.R
```diff
@@ -78,6 +99,54 @@
...
+#'
+#'  \href{https://downlit.r-lib.org/}{\pkg{downlit}} is an R package that
+#'  provides a syntax highlighting engine in R. It will also do automatic
+#'  linking of R code (requires internet connectivity). It is activated only if
+#'  \code{highlight_downlit} is \code{TRUE} and will pass after Pandoc engine to
+#'  replace only R code syntax highlighting and will use any selected Pandoc's
+#'  style. Default color scheme is the accessible theme "arrow". If default
+#'  template is used, some CSS is added to get the link introduced by
+#'  autolinking to use the color scheme chosen. If you want to use with a custom
+#'  template, add this to your template:
```
hadley, line 137:
```suggestion
#'  linking of R code (requires internet connectivity). It is activated only if
#'  \code{highlight_downlit} is \code{TRUE}. It only affects R code, leaving
#'  highlighting for other languages unchanged. The default color scheme is 
#'  the accessible theme "arrow". 
#'
#'  It requires some CSS in the template to correctly style links. This is included
#'  in the default template, but if you want to use with a custom template, you will
#'  need to add this to your template:
```

R/md_document.R
```diff
@@ -22,7 +22,7 @@
 #'   widely used retina displays, but will also result in the output of
 #'   \code{<img>} tags rather than markdown images due to the need to set the
 #'   width of the image explicitly.
-#' @param ext Extention of the output document (defaults to ".md").
+#' @param ext Extension of the output document (defaults to ".md").
```
hadley, line 25:
```suggestion
#' @param ext Extension of the output file (defaults to ".md").
```

R/html_document.R
cderv: I regrouped in one paragraph as `arrow` and `rstudio` also require pandoc 2.0+

R/html_document.R
cderv: I used `\describe` to create the list instead of just bullet. That is clearer to separate the 3 engines.

R/html_document.R
cderv: Yes I forgot to remove after having reworded following Alison PR in distill

apreshill (commented):
hi @cderv - this all looks good - updated `arrow.theme` tweaks now here: https://github.com/rstudio/distill/pull/229/files#diff-e7d4aff16237a3758288199b65102c15bfd3034508428516a7ddadabf37d87f2

cderv:
One thing I missed to handle here is that there are some cases where we want to use **downlit** to produce highlighting for Chroma CSS. This is the case for `blogdown::html_page()`. The current support in **rmarkdown** would allow to have **downlit** applies but will not produce the correct HTML tags for Chroma as it currently uses `classes_pandoc()` only and can only be activated if Pandoc highlighting style is used. 

`blogdown::html_page()` would require `highlight = NULL` and `highlight_downlit = TRUE` with `downlit::classes_chroma()`

Either we should support both in here (using an option or another switch), or this is specific to **blogdown** and should be handle there directly.

yihui (approved):
Okay, I didn't review this PR carefully or read the conversations above. Please feel free to merge if/whenever you feel it's ready. Thanks!

PR #1948, opened 2020-11-12
Support 'href' dependencies

### LINKED ISSUE #794: validate requires file but does not check href

matthewstrasiotto:
Resolves #794

Based on #1805 but no longer includes href reachability tests to validate URLs (They're either online, or they're not at runtime, but the server rendering the document doesn't actually need to use the frontend resources). 

Also no longer includes the RCurl dependency introduced in #1805 . 
If this is merged, #1805 should be closed.

Squash of:

- Support 'href' dependencies ( @crazycapivara )
- Use 'RCurl' to check whether url exists ( @crazycapivara )
- Revert "Use 'RCurl' to check whether url exists"
  This reverts commit 4894bd9f87b523f113a4d0ea7b19abbd9a4d18c1.
  ( @matthewstrasiotto )
- Remove href dependency url check from render side
  Build time reachability is not a requirement for a referenced dependency
  to be available at runtime. ( @matthewstrasiotto )

Co-authored-by: skuethe <crazycapivara@gmail.com>
Co-authored-by: Matthew Strasiotto <39424834+matthewstrasiotto@users.noreply.github.com>

### DIFF
--- a/R/html_dependencies.R
+++ b/R/html_dependencies.R
@@ -213,7 +213,7 @@ html_dependencies_as_string <- function(dependencies, lib_dir, output_dir) {
 
 filter_dependencies <- function(dependencies, src_type = "file") {
   deps <- lapply(dependencies, function(dep) if(!is.null(dep$src[[src_type]])) return(dep))
-  deps[!sapply(deps, is.null)]
+  deps[!vapply(deps, is.null, logical(1))]
 }
 
 # check class of passed list for 'html_dependency'
--- a/R/html_dependencies.R
+++ b/R/html_dependencies.R
@@ -194,16 +194,26 @@ html_reference_path <- function(path, lib_dir, output_dir) {
 # return the html dependencies as an HTML string suitable for inclusion
 # in the head of a document
 html_dependencies_as_string <- function(dependencies, lib_dir, output_dir) {
+  # Only copy deps with path
+  dependencies_href <- filter_dependencies(dependencies, "href")
+  dependencies <- filter_dependencies(dependencies, "file")
 
   if (!is.null(lib_dir)) {
     dependencies <- lapply(dependencies, copyDependencyToDir, lib_dir)
     dependencies <- lapply(dependencies, makeDependencyRelative, output_dir)
   }
-  return(renderDependencies(dependencies, "file", encodeFunc = identity,
+  #return(
+  tags <- renderDependencies(dependencies, "file", encodeFunc = identity,
                             hrefFilter = function(path) {
                               html_reference_path(path, lib_dir, output_dir)
                             })
-  )
+  #)
+  HTML(paste0(c(tags, renderDependencies(dependencies_href, "href")), collapse = "\n"))
+}
+
+filter_dependencies <- function(dependencies, src_type = "file") {
+  deps <- lapply(dependencies, function(dep) if(!is.null(dep$src[[src_type]])) return(dep))
+  deps[!sapply(deps, is.null)]
 }
 
 # check class of passed list for 'html_dependency'
@@ -223,15 +233,20 @@ validate_html_dependency <- function(list) {
     stop("name for html_dependency not provided", call. = FALSE)
   if (is.null(list$version))
     stop("version for html_dependency not provided", call. = FALSE)
-  list <- fix_html_dependency(list)
-  if (is.null(list$src$file))
-    stop("path for html_dependency not provided", call. = FALSE)
-  file <- list$src$file
-  if (!is.null(list$package))
-    file <- system.file(file, package = list$package)
-  if (!file.exists(file)) {
-    utils::str(list)
-    stop("path for html_dependency not found: ", file, call. = FALSE)
+  list <- fix_html_dependency(list) 
+
+  # check src path or href are given
+  if (is.null(list$src)) {
+    stop("neither path nor href are provided for html_dependency", call. = FALSE)
+  }
+  if (!is.null(list$src$file)) {
+    file <- list$src$file
+    if (!is.null(list$package))
+      file <- system.file(file, package = list$package)
+    if (!file.exists(file)) {
+      utils::str(list)
+      stop("path for html_dependency not found: ", file, call. = FALSE)
+    }
   }
 
   list


matthewstrasiotto:
@yihui , @cderv please review this when you have time.

matthewstrasiotto:
@yihui @cderv please review this - as of https://github.com/rstudio/htmltools/pull/188 , href dependencies may be safely fetched with SRI checks, and these APIs diverge.

R/html_dependencies.R
```diff
@@ -224,14 +234,19 @@ validate_html_dependency <- function(list) {
...
-  if (is.null(list$src$file))
-    stop("path for html_dependency not provided", call. = FALSE)
-  file <- list$src$file
-  if (!is.null(list$package))
-    file <- system.file(file, package = list$package)
-  if (!file.exists(file)) {
-    utils::str(list)
-    stop("path for html_dependency not found: ", file, call. = FALSE)
+  #if (is.null(list$src$file))
+  #  stop("path for html_dependency not provided", call. = FALSE)
```
matthewstrasiotto, line 238: Commented code- will remove

matthewstrasiotto: Resolved by 340594e

matthewstrasiotto:
:eyes: @yihui @cderv

cderv:
Thanks for the ping. We'll look at that quickly. Thanks for your patience - we've dealt with our delay on other packages before **rmarkdown** but it will come next.

R/html_dependencies.R
```diff
@@ -194,16 +194,26 @@ html_reference_path <- function(path, lib_dir, output_dir) {
...
   }
-  return(renderDependencies(dependencies, "file", encodeFunc = identity,
+  #return(
+  tags <- renderDependencies(dependencies, "file", encodeFunc = identity,
                             hrefFilter = function(path) {
                               html_reference_path(path, lib_dir, output_dir)
                             })
-  )
+  #)
+  HTML(paste0(c(tags, renderDependencies(dependencies_href, "href")), collapse = "\n"))
```
cderv, line 211: I believe the current implementation means that all `href` deps will be loading **after** `file` deps. 
I would think **rmarkdown** should not touch the order of dependencies. Is it safe to change the order ? 
Could there not be side effect ?

R/html_dependencies.R
```diff
@@ -194,16 +194,26 @@ html_reference_path <- function(path, lib_dir, output_dir) {
...
+}
+
+filter_dependencies <- function(dependencies, src_type = "file") {
+  deps <- lapply(dependencies, function(dep) if(!is.null(dep$src[[src_type]])) return(dep))
+  deps[!sapply(deps, is.null)]
```
cderv, line 216:
```suggestion
  deps[!vapply(deps, is.null, logical(1))]
```

`vapply` is safer than `sapply`

cderv (changes requested):
Thanks for this. 

This PR is simpler than #1805. I don't think either than `RCurl` is needed and that `href` url shoud be checked.

Did you test this PR with several examples ? 

I have one question regarding dependency orders. The current implementation should change the order of dependencies (order of the list different than the string added in header). Is this safe ?

A note from **htmltools** behavior. `copyDependencyToDir` and `makeDependencyRelative` have a `mustWork` argument that we could set to FALSE so that `href` dependencies are just returned as is and `file` one are processed and copied. 
However, we currently the filter because we need to apply `html_reference_path()` as a `hrefFilter` function in `renderDependencies` only to `file` dependencies. 

It would be interesting to process both differently but keep the order maybe. 

Am I worrying too much about this order ?

matthewstrasiotto:
Just saw this, thank you for the review.

I'm not sure about the order, but I'd like whatever changes this imposes on the output files to be as minimal and specific and possible, so tomorrow I'll have another look at it, reason with your concerns about order, and try to make it preserve order

R/html_dependencies.R
matthewstrasiotto: Looking at this now, I can't see where the makeDependencyRelative / copyDependency logic is reimplemented. 
I'll have to check this out properly

R/html_dependencies.R
```diff
@@ -194,16 +194,26 @@ html_reference_path <- function(path, lib_dir, output_dir) {
...
                             hrefFilter = function(path) {
                               html_reference_path(path, lib_dir, output_dir)
                             })
-  )
+  #)
+  HTML(paste0(c(tags, renderDependencies(dependencies_href, "href")), collapse = "\n"))
+}
+
+filter_dependencies <- function(dependencies, src_type = "file") {
+  deps <- lapply(dependencies, function(dep) if(!is.null(dep$src[[src_type]])) return(dep))
```
matthewstrasiotto, line 215: consider changing this to accept a vector of src_types, like:

```r
function(dep) if (any(!vapply(dep$src[src_types], is.null, logical(1)))) return(dep)
```

I'd probably write this lambda as multiple lines for readability:
```r
function(dep) {
  undefined_types <- vapply(dep$src[src_types], is.null, logical(1))
  if ( ! all(undefined_types) ) return(dep)
}
```

The above should circumvent reordering dependencies.

R/html_dependencies.R
matthewstrasiotto: & I agree about the order, I will probably modify filter_dependencies to take a vector of `src_types` to cover both types at once.

R/html_dependencies.R
cderv: We can also not filter and process dependencies in a for-loop as `renderDependencies` is doing internally. 
This would allow to apply the correct `renderDependencies` function, because only difference we need to make currently between href and files are about `encodeFunc=` and `hrefFilter=` in `renderDependencies()`. 
We could also revisit that to simplify maybe. 

This is considering my comment above regarding href deps being ignored (https://github.com/rstudio/rmarkdown/pull/1948#pullrequestreview-628985594) which confirmed by @jcheng5 in https://github.com/rstudio/rmarkdown/pull/1805#discussion_r608191632

cderv: Something like this based on your change. 
````diff
diff --git a/R/html_dependencies.R b/R/html_dependencies.R
index 9f92dacb..6a81d9ee 100644
--- a/R/html_dependencies.R
+++ b/R/html_dependencies.R
@@ -194,26 +194,27 @@ html_reference_path <- function(path, lib_dir, output_dir) {
 # return the html dependencies as an HTML string suitable for inclusion
 # in the head of a document
 html_dependencies_as_string <- function(dependencies, lib_dir, output_dir) {
-  # Only copy deps with path
-  dependencies_href <- filter_dependencies(dependencies, "href")
-  dependencies <- filter_dependencies(dependencies, "file")

   if (!is.null(lib_dir)) {
-    dependencies <- lapply(dependencies, copyDependencyToDir, lib_dir)
-    dependencies <- lapply(dependencies, makeDependencyRelative, output_dir)
+    dependencies <- lapply(dependencies, copyDependencyToDir,
+                           output_dir = lib_dir, mustWork = FALSE)
+    dependencies <- lapply(dependencies, makeDependencyRelative,
+                           basepath = output_dir, mustWork = FALSE)
   }
-  #return(
-  tags <- renderDependencies(dependencies, "file", encodeFunc = identity,
-                            hrefFilter = function(path) {
-                              html_reference_path(path, lib_dir, output_dir)
-                            })
-  #)
-  HTML(paste0(c(tags, renderDependencies(dependencies_href, "href")), collapse = "\n"))
-}
-
-filter_dependencies <- function(dependencies, src_type = "file") {
-  deps <- lapply(dependencies, function(dep) if(!is.null(dep$src[[src_type]])) return(dep))
-  deps[!sapply(deps, is.null)]
+  html <- c()
+  for (dep in dependencies) {
+    tags <-
+      if (is.null(dep$src[["file"]])) {
+        renderDependencies(list(dep), "href")
+      } else {
+        renderDependencies(list(dep), "file", encodeFunc = identity,
+                           hrefFilter = function(path) {
+                             html_reference_path(path, lib_dir, output_dir)
+                           })
+      }
+    html <- c(html, tags)
+  }
+  HTML(paste(c(html, collapse = "\n")))
 }
````

I am still wondering if we really need to differentiate `renderDependencies` call. 🤔 
it is a pretty old code (since **rmarkdown** 0.6 maybe) so hard to tell.

cderv:
@matthewstrasiotto do you want to continue this PR ? 

I now get a good sense of how it should work. I can work on it next week based on your work here and our discussion above. So happy to take it from here if it is good for you. You tell me. 

Thanks!

matthewstrasiotto:
Hi @cderv I'm happy to leave this in your capable hands to complete (making this a three coauthor PR 🙃 ) 
thank you for your consideration

cderv (approved):
I'll work on finishing this and adding some tests. Thanks for the contribution on this one !

R/html_dependencies.R
cderv: Doing this https://github.com/rstudio/rmarkdown/pull/1948/files#r608469286 should keep the order of dependencies. Trick is to use `mustWork = FALSE` in htmltools functions

R/html_dependencies.R
cderv: I kept the differentiation for now and went with the code change above

cderv:
@matthewstrasiotto It seems I can't push your branch in your fork. Just a sanity check: Did you allow edit from maintainer on this PR ? https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

It could be another issue on my side though but I wanted to check with you

cderv:
But that is possible I can't because you created your fork from another repo which is a fork of this one. 
Thinking this from the doc: https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/committing-changes-to-a-pull-request-branch-created-from-a-fork

I may do another branch and yet another PR if I can't update this one.

matthewstrasiotto:
I forked from a fork. You can fork from mine if that's the answer 😛
Otherwise i can try to fix this on my end

matthewstrasiotto:
I'll fix this when I'm in front of a computer

cderv:
I did a new PR to unblock me. #2151 is now merged. I'll close this PR then

Issue #1956, opened 2020-11-20
Pandoc 2.11.2 now default to ATX header 


cderv:
This is a change in the [new 2.11.2 release](https://pandoc.org/releases.html#pandoc-2.11.2-2020-11-19). 
> Default to using ATX (##-style) headings for Markdown output (#6662, Aner Lucero). Previously we used Setext (underlined) headings by default for levels 1–2.
>
> Add option --markdown-headings=atx|setext, and deprecate --atx-headers (#6662, Aner Lucero).

That means `rmarkdown::md_document` won't output the same now with Pandoc 2.11.2
``` r
tmp_file <- tempfile(fileext = ".Rmd")
xfun::write_utf8(c(
  "# Header 1",
  "",
  "## Header 2",
  "",
  "Some content"), tmp_file
)

# Pandoc 2.7.3
rmarkdown::find_pandoc(dir = "C:/Users/chris/Logiciel/pandoc/pandoc-2.7.3", cache = FALSE)
#> $version
#> [1] '2.7.3'
#> 
#> $dir
#> [1] "C:/Users/chris/Logiciel/pandoc/pandoc-2.7.3"

res <- rmarkdown::render(tmp_file, "md_document")
#> processing file: file9c046df1eed.Rmd
#> output file: file9c046df1eed.knit.md
#> "C:/Users/chris/Logiciel/pandoc/pandoc-2.7.3/pandoc" +RTS -K512m -RTS file9c046df1eed.utf8.md --to markdown_strict --from markdown+autolink_bare_uris+tex_math_single_backslash --output file9c046df1eed.md --standalone
#> 
#> Output created: file9c046df1eed.md
xfun::file_string(res)
#> Header 1
#> ========
#> 
#> Header 2
#> --------
#> 
#> Some content

# Pandoc 2.11.2
rmarkdown::find_pandoc(version = "2.11.2", cache = FALSE)
#> $version
#> [1] '2.11.2'
#> 
#> $dir
#> [1] "C:/Users/chris/scoop/shims"

res <- rmarkdown::render(tmp_file, "md_document")
#> processing file: file9c046df1eed.Rmd
#> output file: file9c046df1eed.knit.md
#> "C:/Users/chris/scoop/shims/pandoc" +RTS -K512m -RTS file9c046df1eed.utf8.md --to markdown_strict --from markdown+autolink_bare_uris+tex_math_single_backslash --output file9c046df1eed.md --standalone
#> 
[...]

PR #1957, opened 2020-11-20
apply the flag --atx-headers only for pandoc version before 2.11.2

### LINKED ISSUE #1919: investigate differences in gfm output w pandoc 2.11
I've noticed that now GFM behaves a bit differently. For example, in Pandoc 2.9.2.1 

```bash
echo '## <https://www.google.com>' | pandoc --to gfm
```
Gives:

```
## <https://www.google.com>
```

Whereas in Pandoc 2.11 it gives:

```text
<a href="https://www.google.com" class="uri">https://www.google.com</a>
-----------------------------------------------------------------------
```

cderv:
closes #1919)

I started to do a fix which would have been more complicated, that is creating a wrapper function `pandoc_atxheaders_args(activate = TRUE)` to deal with all the versions of Pandoc

* Before Pandoc 1.9, everything was ATX headers
* `--atx-headers` exists since Pandoc 1.9 because they change the default for markdown writer to setex for level 1 and 2
*  Pandoc 2.11.2 resets the default to ATX headers, and change the flag to `--markdown-headings`

So a function like 
```r
pandoc_atxheaders_args <- function(activate = TRUE) {
  if (pandoc_available("2.11.2")) {
    # atx headers are the default in Pandoc 2.11.2+
	if (!activate) "--markdown-headings=setext"
  } else if (pandoc_available("1.9")) {
    # atx headers were removed as default in 1.9
    if (activate) "--atx-headers"
  }
  # before Pandoc 1.9 that was only ATX headers
}
```

But it does not seem we do that much conditional in **rmarkdown** and we try to have minimal scoped fix. 

@yihui do you think a wrapper like would be interesting ?

### DIFF
--- a/R/github_document.R
+++ b/R/github_document.R
@@ -48,9 +48,8 @@ github_document <- function(toc = FALSE,
 
   # atx headers are the default in pandoc 2.11.2 and the flag has been deprecated
   # to be replace by `--markdown-headings=atx|setx`
-  pandoc_args <- c(
-    if (!pandoc_available("2.11.2")) "--atx-headers",
-        pandoc_args
+  if (!pandoc_available("2.11.2")) pandoc_args <- c(
+    "--atx-headers", pandoc_args
   )
 
   format <- md_document(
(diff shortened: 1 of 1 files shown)


cderv:
Check failure is due to #1956

yihui (approved):
The minimal fix looks good to me. Thanks!

PR #1961, opened 2020-11-24
Revert change about formatting for Pandoc headerless simple table

### LINKED ISSUE #1954: Incorrect format for HTML tables
Here is a minimal reprex
````md
---
title: "table format"
output: html_document
---

```{r}
knitr::kable(iris, "html")
```
````

This is caused by the change in https://github.com/rstudio/rmarkdown/pull/1895 and the fact that `kable` html output does not include odd and even on rows. 

Discovered by a user while loading kableExtra.
````md
---
title: "table format"
output: html_document
---

```{r}
library(knitr)
library(kableExtra)
knitr::kable(iris)
```
```` 

**kableExtra** is modifying the format for `knitr::kable` that default to markdown pipe output. By auto formating to html, just loading the package modify the behavior of kable. 

Two solutions: 
* We could revert and fix another solution for #1893 
* Modify **knitr** html table to include the `odd` and `even` class.

For now the former seems better, as there is a JS solution for #1893 (see https://github.com/rstudio/rmarkdown/issues/1893#issuecomment-688877872)

cc @apreshill

### PR DESCRIPTION
**Pull request by cderv**
Close #1954 by revertgin the change in #1895 for #1894
This reverts commit e1aebefffcc89d0f71b74805ee7f7aeb7f89a620

Should we found another JS solution for #1894 or just document in NEWS.md that one could use 
````md
```{js, echo=FALSE}
$('table').addClass('table table-condensed'); 
```
````
to get the correct formatting for headerless table ?

### DIFF
--- a/inst/rmd/h/default.html
+++ b/inst/rmd/h/default.html
@@ -602,7 +602,7 @@ $endif$
 
 // add bootstrap table styles to pandoc tables
 function bootstrapStylePandocTables() {
-  $$('tr.odd').parent('tbody').parent('table').addClass('table table-condensed');
+  $$('tr.header').parent('thead').parent('table').addClass('table table-condensed');
 }
 $$(document).ready(function () {
   bootstrapStylePandocTables();

### CODE REVIEW

**Comment by cderv** [comment]
in fact this is not the issue with #1954 - Closing. I'll reopen if needed.

File: inst/rmd/h/default.html [hunk 1]
```diff
@@ -602,7 +602,7 @@ <h2 id="$idprefix$toc-title">$toc-title$</h2>
 
 // add bootstrap table styles to pandoc tables
 function bootstrapStylePandocTables() {
-  $$('tr.odd').parent('tbody').parent('table').addClass('table table-condensed');
+  $$('tr.header').parent('thead').parent('table').addClass('table table-condensed');
```
yihui, line 605: How about finding tables that have either `thead > tr.header` or `tbody > tr.odd` as their children, and add the classes to them? Would that fix both problems?

PR #1964, opened 2020-11-26
Rework anchor_sections to change default behavior

### LINKED ISSUE #1946: Re-consider a more user-friendly anchor symbol, or make the default FALSE for anchor_sections
Re: https://github.com/rstudio/rmarkdown/pull/1884

I think the default `anchor_sections = TRUE` is hard as a default given the `#` symbol. It isn’t intuitive or self-explanatory, and most people don’t know that those are called anchors, so finding the argument `anchor_sections` to turn it off is hard to figure out. Also, because the hashtag symbol has a special meaning as a header in markdown syntax, it looks like a weird markdown rendering error when you see it- I even knew this feature was in place and STILL had a freak out and initially thought I was seeing an odd markdown parsing artifact at first.

On the other hand, precisely because many users may not be familiar with the concept of anchor headers (even if they enjoy them!), it will be hard to figure out that you can turn it _ON_ as well if the default is `FALSE`. And I think the utility is extremely useful for users (adding anchor symbols to my HTML-based R Markdown outputs is often the first thing I do, and the first thing I asked my intern Desiree to do!- thank you @atusy for recognizing that need and adding the functionality 🌟 ).

But, the more I use it in my own recent work across output formats and see users attempting to trouble-shoot this new feature, I think that if we want to keep `TRUE` as the default, we should switch to an svg symbol. This is familiar to many users because it is how GitHub `README`s look, `pkgdown` websites, and now `distill` as well, and there is even a link emoji :link:. All of the above boils down to:

+ If we keep the `#` symbol, I lean towards setting `FALSE` as the default across the board.
+ If we are open to an svg (single file, open source like the link icon here: https://material.io/resources/icons/?search=link&style=baseline), I think `TRUE` can stay as the default (with slide decks and perhaps pagedown setting it to `FALSE`)

## Some examples with svg icons

Screen Shot 2020-11-10 at 1 21 54 PM

Screen Shot 2020-11-10 at 12 01 47 PM
Screen Shot 2020-11-10 at 12 02 18 PM
Screen Shot 2020-11-10 at 12 02 55 PM
Screen Shot 2020-11-10 at 12 03 24 PM
Screen Shot 2020-11-10 at 1 19 34 PM

---

cderv:
This will close #1946 

This PR is a two step process

## Reverting the default

First step shared in this PR is to revert the default and use FALSE instead of TRUE. Some examples are also added in `?html_document` about how to customize. The style has not changed and `#` is still used. I did not modified the JS - even if we don't need to check for existing anchor I believe as it will not be activated by default.

### Changes: 

https://github.com/rstudio/rmarkdown/pull/1964/files/f26da730418d4b62de3c72b6dd7ba0bc745a31c8

### New help section: 
image

## Simplifying customization

Second step is experimenting with more helping way of customization. I believe we can make the choice easier for users who wants an icon instead of the dash. I have something like that in mind
````yaml
output:
  html_document:
    anchor_sections:
      style: 'icon'
````

I don't think this would be difficult to maintain, and it would be a lot easier for helping non advanced user (who don't know CSS) to customize there outputs. 

It is in progress hence the draft PR.

/cc @atusy @apreshill @yihui

### CODE REVIEW

**Comment by cderv** [comment]
We can now do some customization easily 

* Using `#` is the default when TRUE
```yaml
output: 
  html_document:
    anchor_sections: TRUE
```
* But you could specify the style now. This would be the same. 
```yaml
output: 
  html_document:
    anchor_sections: 
      style: hash
```

And we have predefined  a symbol, using `\01F517\00FE0E` - it will be different on where it is seen (as explained here https://github.com/rstudio/rmarkdown/issues/1946#issuecomment-725520453)
```yaml
output: 
  html_document:
    anchor_sections: 
      style: symbol
```
image

and 

```yaml
output: 
  html_document:
    anchor_sections: 
      style: icon
```
which uses an svg icon
image


Do you think it is useful ? 
Currently the implementation forces me to use several css files. I don't think it is the best, but it helps me test the API. I would have prefer to have a logic of templating but I guess that would require lua filter and pandoc template. Or ability to apply JS differently based on information in R. (I should look in bookdown maybe...)

I am thinking also of adding something to easily control the level you want to apply - only level 1 for example
```yaml
output: 
  html_document:
    anchor_sections: 
      style: icon
      levels: 1
```
but it would definitely changing the way it works - currently it would require passing parameters from R to JS. Not sure how and if it is a good idea.

Still WIP but feedbacks are welvome.

cderv:
We could support something like this 
```yaml
output: 
  html_document:
    anchor_sections: 
      style: icon
      maxlevel: 1
```

if we revert to using a lua filter as @atusy first implementation. We could pass the value as metadata to pandoc (`-M rmd_anchor_maxlevel=1`) and use a lua filter like this: 
```lua
local anchor_maxlevel

-- retrieve metadata
local function get_metadata(el)
  anchor_maxlevel = tonumber(el.rmd_anchor_maxlevel)
  if not anchor_maxlevel then
    anchor_maxlevel = 6
  end
end

-- insert anchor link
function insert_anchor(el)
  if el.identifier ~= "" and el.level <= anchor_maxlevel then
    el.classes:insert("hasAnchor")
    table.insert(el.content,
      pandoc.Link("", "#"..el.identifier, "", {class = "anchor-section"})
    )
  end
  return(el)
end


return {
  { Meta = get_metadata },
  { Header =  insert_anchor}
}
```

With the current implementation in JS, I am not sure how we could modify the levels to apply to (unless building js script from a template in a temp folder)
https://github.com/rstudio/rmarkdown/blob/8cc31e5e48658cca2f590474cdc9ec783a72ac03/inst/rmd/h/anchor-sections/anchor-sections.js#L8

In the first place we switch to JS implementation because we where thinking of defaulting to FALSE and requiring checking is anchors where already set (https://github.com/rstudio/rmarkdown/pull/1884#discussion_r496809335). As we default to FALSE now this would be required anymore IMO. 

Do you think this would be interesting to support this max level header to apply anchor on ? 
It is less important than styling but I can see the need.

atusy:
@cderv I think sass will be a nice solution.

- Generate one CSS file conditionally, instead of combining multiple CSS files conditionally.
- Hide anchors with headers whose levels exceeds `anchor_maxlevel`

cderv:
@atusy thanks for your feedback.

yes sass would definitely help. I just don’t want yet to introduce this dependency. But when it will be one, we could switch. 

About ˋanchor_maxlevel` with the solution above there is no more JS. Only a lua filter which adds the html code and css to make the styling. Do you think this would be a useful feature or only anchor_sections for all or non makes more sense ? 

If you have any comment as review on this PR, you’re welcome to share!

apreshill:
Hi @cderv -

I really like this approach, and the added flexibility of max level. It reminds me of the `toc_depth` argument too, so intuitive. And being able to pick header levels for anchors mirrors a great feature of Hugo's Goldmark rendering approach to adding anchors, see: https://gohugo.io/getting-started/configuration-markup/#heading-link-example

cderv:
> In the first place we switch to JS implementation because we where thinking of defaulting to FALSE and requiring checking is anchors where already set (#1884 (comment)). As we default to FALSE now this would be required anymore IMO.

As we decided `anchor_sections` is now optional, I think we can avoid using the JS script that was mainly there in the first place to check is anchor section was already in use. As it is FALSE by default, it will be the user responsibility to not activate it. 

Not using JS means using a Lua filter instead. That allows us to add a feature I think will be appreciated: selected the depth on which to apply the anchor on. Feedbacks from @apreshill and @mine-cetinkaya-rundel confirmed that there is some potential use case.

So a TL;DR on this PR: 

* `anchor_sections: FALSE` is the default now. 
* `anchor_sections: TRUE` will add an anchor link on all headers (h1 to h6) using the default style with a `#`, underlined on hover. 
* There is two levels of customization available: 
  ```yaml
  output:
    html_document:
      anchor_sections:
        style: symbol # style of the anchor ("dash", "symbol", "icon")
        depth: 2 # max depth to apply anchor on (default to max which is 6)
  ```
  * 3 styles are predifined and anchor can be set only on some levels starting by h1 header.
* `?html_document` has detailed documentation about this feature now.

* Drawback: Anchor sections is implemented with a Lua filter, so it requires Pandoc 2.0+. As it is optional, I don't think that is an issue. And most of users should be using Pandoc 2+ already. 

* I also added some tests on all this, and I tried to factor out in a function a logic which is usually inserted inside `html_document()` body directly (see [`aee4eb9` (#1964)](https://github.com/rstudio/rmarkdown/pull/1964/commits/aee4eb99d219440b31c79decdfe74594a98ae905) for this factoring-out step)
    * I think this helps for the readability of the code - maybe easier maintenance in the future
    * It clarifies test allowing to write smaller tests, closer to unit test than integration test. 
I was also thinking about tests for each feature using Rmd file to render and compare - something like knitr-example, using snapshot test, but I am not sure yet if it should live here or in its own repo. So I added on the TODO / IDEAS list for next time.

* I also added some logic to check user inputs. **rmarkdown** does not use yet  **assertthat** or any other - should we ? For now, some if condition and `stop()` with message are the least we could do. The aim was to help fail fast if someone does not provide the correct yaml entry for `anchor_sections`. 

IMO this PR is ready. Reviews welcome @yihui @apreshill @atusy !

inst/rmd/h/anchor-sections/anchor-sections.css
```diff
@@ -1,5 +1,4 @@
 /* Styles for section anchors */
 a.anchor-section {margin-left: 10px; visibility: hidden; color: inherit;}
-a.anchor-section::before {content: '#';}
```
atusy, line 3: What about keep this line so that people can easily find the default behavior?
We can override the rule by including other CSS files.

man/figures/link-black-18dp.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="black" width="18px" height="18px"><path d="M0 0h24v24H0z" fill="none"/><path d="M3.9 12c0-1.71 1.39-3.1 3.1-3.1h4V7H7c-2.76 0-5 2.24-5 5s2.24 5 5 5h4v-1.9H7c-1.71 0-3.1-1.39-3.1-3.1zM8 13h8v-2H8v2zm9-6h-4v1.9h4c1.71 0 3.1 1.39 3.1 3.1s-1.39 3.1-3.1 3.1h-4V17h4c2.76 0 5-2.24 5-5s-2.24-5-5-5z"/></svg>
```
atusy, line 1: What about move this file to `inst/rmd/h/anchor-sections`.
We can include this file as an HTML dependency.
In this way, we can render the file correctly even if offline.

atusy (commented):
Good job! Let me add some suggestions.
BTW, you've solved two other suggestions I was about to make (rename 'maxlevel' and refactor `html_document` by implementing `add_anchor_secions`). Nice!!

inst/rmd/h/anchor-sections/anchor-sections-hash.css
```diff
@@ -0,0 +1,2 @@
+/* Styles for section anchors */
+a.anchor-section::before {content: '#';font-size: 80%;}
```
atusy, line 2: We can remove this file if we accept:
https://github.com/rstudio/rmarkdown/pull/1964#discussion_r533428044

inst/rmd/h/anchor-sections/anchor-sections.css
cderv: an remove `anchor-section-hash.css` ? 
I thought regrouping everything was clearer (I added a font-size rule too)

Regarding "finding behavior", I expect it to be easier with the documentation I made (in NEWS and in `?html_document`) that looking through the files. 

No strong opinion on this, even if I prefer separating the styling in files.

man/figures/link-black-18dp.svg
cderv: I looked for that but wasn't sure how to add an svg file (or more generally a resource file) for a CSS usage, as html dependency. Do you have example of that somewhere ?

inst/rmd/h/anchor-sections/anchor-sections.css
atusy: Yes, your document is clear.
I just prefer self-explaining code.
Mine is also not a strong opinion.
You can ignore.

inst/rmd/h/anchor-sections/anchor-sections-hash.css
cderv: Ok. but there is the `font-size` that I added only for the dash style. 
It would apply also on other style if we put it in the default css file right ?

inst/rmd/h/anchor-sections/anchor-sections.css
cderv: By self explaining you mean having the default `anchor-sections.css` contains the whole default style for hash character ? instead of having a default which is a combination of two files ?

man/figures/link-black-18dp.svg
atusy: Yes, the example is my initial implementation of anchor sections:
https://github.com/atusy/rmarkdown/blob/323135ef2baea445976427865d92c9d7d19aec35/inst/rmd/h/anchor-sections/anchor-sections.css#L17

inst/rmd/h/anchor-sections/anchor-sections-hash.css
atusy: Ah, yes. Let's keep the current manner then. Alhough I see the other CSS defines `font-size` by themselves if needed, this will be a better choice that reduces unexpected conflicts with the additional styles in the future.
https://github.com/rstudio/rmarkdown/pull/1964/files#diff-a354a361dd9b24096377c197932c131ed2d45352ebd40d80cff96855829b278aR4

inst/rmd/h/anchor-sections/anchor-sections.css
atusy: Yes.

inst/rmd/h/anchor-sections/anchor-sections-hash.css
cderv: yes there is non when I added the icon. I think I'll keep the 3 files
* Common rules in `anchor-sections.css`
* Specific rule for style argument in `anchor-sections-<style>.css`

man/figures/link-black-18dp.svg
cderv: I missed that - sorry. I was not aware this was correctly copied with the link being modified. I should have trusted **rmarkdown** and **htmltools** to be clever enough. 

It could indeed be interesting to include it. 

Re. offline usage, `html_document()` resulting HTML file requires internet access anyway as I think we load mathjax from a url for now. When you have math expression obviously. Otherwise, everything is fine.

Should we use the link icon that I missed from your PR ? Where does it come from ? The one I used is Apache 2 licensed.

inst/rmd/h/anchor-sections/anchor-sections.css
cderv: https://github.com/rstudio/rmarkdown/pull/1964#discussion_r533472286

man/figures/link-black-18dp.svg
atusy: Don't worry.

- MathJax requires internet access but falls back to TeX math if offline, right? On the other hand, the svg just disappears if offline.
- The icon comes from pkgdown (<https://github.com/r-lib/pkgdown/blob/master/inst/assets/link.svg>)
- I just noticed that we can add alt text to the CSS `content` (https://developer.mozilla.org/en-US/docs/Web/CSS/content). Will it be helpful if in terms of accessibility? (cc. @jooyoungseo )

cderv: I'll try to include it. That will also prevent any issue with the link no more working. Thanks for all the information !

jooyoungseo: > * I just noticed that we can add alt text to the CSS `content` (https://developer.mozilla.org/en-US/docs/Web/CSS/content). Will it be helpful if in terms of accessibility? (cc. @jooyoungseo )

Absolutely! That would be very instrumental. `content` alt text is getting supported by and for modern browsers and assistive technologies. Thanks for highlighting that aspect, @atusy!

As a side note concerning Math content, TeX Math is not rendered to be accessible for assitive technologies whereas MathJax is fully accessible. I know it'd be an offline fallback, but having a local copy of MathJax would be ideal when offline for accessibility purpose. (No pressure)

cderv:
@atusy I found one drawback with the lua filter: the `hasAnchor` class I add to the header is in fact added by Pandoc to the div and not the header because of `section-divs`. 

That lead to all anchor sections link of the same h1 part to become visible 😅  not so bad but different behavior as before. 

I guess there is no trick for that on the lua side... With JS, you were able to add `hasAnchor` to the parent of the added link, hence the header node.

man/figures/link-black-18dp.svg
cderv: Thanks ! 
What would be a useful alt text for that ? Should we use empty alt text so that it is "skipped" ? or should we had a property to ignore the anchor link in screen reader ?

atusy:
@cderv

> all anchor sections links of the same h1 part to become visible on hover

Yes. That is why my initial implementation did not use the `hasAnchor` class.
Refer to https://github.com/atusy/rmarkdown/blob/323135ef2baea445976427865d92c9d7d19aec35/inst/rmd/h/anchor-sections/anchor-sections.css#L1-L9 .

cderv:
I see. Sorry for all the trouble redoing part of what you have done... hope you don't mind.
Switching back to lua was only to provide this `depth` feature. Seems useful though... 

I'll keep thinking about what is best here.

man/figures/link-black-18dp.svg
cderv: @jooyoungseo I am not sure what we should do for this anchor link on sections. 

* What would you expect as alt text for the generated content ? Just a description (like "Anchor svg icon")
* This `content:` alt text is not supported on all browser it seems. 
See https://developer.mozilla.org/en-US/docs/Web/CSS/content#Browser_compatibility
    * Safari and Firefox seem to not support it. I don't know yet if they ignore or if this create an error in the content insertion.

cderv:
This is now working as expected : Anchor link is visible when header (not section div) are hovered. 

The API has not changed - only the internals
```yaml
output:
  html_document:
    anchor_sections:
      style: symbol # style of the anchor ("dash", "symbol", "icon")
      depth: 2 # max depth to apply anchor on (default to max which is 6)
```

* I ended up using a JS script only if `section_divs = TRUE` to move the `hasAnchor` class used by CSS on the header, and remove it from the section div. 
* This allows to have a Lua filter that will do the job alone when `section_divs = FALSE`
* It also enables some trick to work: Setting anchor manually on headers (is that useful ? 🤔 )
````md
---
title: manual anchor
output:
  html_document:
    anchor_sections:
      depth: 0
---

# A header without anchor

## A header with a manual anchor {.hasAnchor}
```

I am still waiting for comments regarding the alt text for CSS content (https://github.com/rstudio/rmarkdown/pull/1964#discussion_r537411471)

Test coverage is ok, but test is failing on Windows due to a broken TinyTeX installation. Also, previous conflict has been resolved.

man/figures/link-black-18dp.svg
cderv: @jooyoungseo friendly ping if you can help me on this one ? Thank you

DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: rmarkdown
 Type: Package
 Title: Dynamic Documents for R
-Version: 2.6.0001
+Version: 2.7.1
```
cderv, line 4: To update to correct version before merging

R/html_document.R
```diff
@@ -81,28 +83,50 @@
...
+#'  output:
+#'    html_document:
+#'      anchor_sections: TRUE
+#'  }
+#'  There is currently two options to modify the default behavior
+#'
+#'  \describe{
+#'  \item{\code{style}}{It allows to select a predefined style:
+#'  \itemize{
+#'  \item{\code{style = "dash"}, the default, is using a \samp{#}, a minimalist choice referring to the id selector in HTML and CSS.}
```
hadley, line 97:
```suggestion
#'  \item{\code{style = "dash"}, the default, uses \samp{#}, a minimalist choice that evokes the id selector from HTML and CSS.}
```

R/html_document.R
```diff
@@ -81,28 +83,50 @@
...
+#'  \item{\code{style = "symbol"} will use a \href{https://codepoints.net/U+1F517}{link
+#'  symbol} \if{html}{\out{(&#x1F517;&#xFE0E;)}}}
+#'  \item{\code{style = "icon"} will use an svg icon from \url{https://material.io/resources/icons/}
+#'  \if{html}{(\figure{link-black-18dp.svg}{options: alt="icon link"})}}
+#'  }
+#'  You can also easily change to any of your choosing using a css rule in your
```
hadley, line 103:
```suggestion
#'  You can also customise using a css rule in your
```

R/html_document.R
```diff
@@ -81,28 +83,50 @@
...
+#'  document. For example, to get a pictogram \if{html}{\out{(&#x1F517;)}}:
```
hadley, line 104: Isn't this the link symbol?

R/html_document.R
```diff
@@ -81,28 +83,50 @@
...
-#'  \if{html}{\out{(&#x1F517;)}}.
-#'
-#'  If you prefer an svg icon, you can also use one using for example a direct link or downloading it from
-#'  \url{https://material.io/resources/icons/}.
+#'  About how to apply custom CSS in R Markdown document, see
+#'  \url{https://bookdown.org/yihui/rmarkdown-cookbook/html-css.html}
+#'  }
+#'  \item{\code{depth}}{It allows to select the minimum header level to apply
+#'  the anchor link on. For example, this will apply the symbol style
+#'  on level 1 and 2 only:
```
hadley, line 114:
```suggestion
#'  \item{\code{depth}}{Select the maximum header level to add the 
#'  anchor link to. For example, this yaml will use the symbol style and
#'  only with level 1 and 2 headings:
```

R/html_document.R
```diff
@@ -81,28 +83,50 @@
...
-#'     Licence: https://www.apache.org/licenses/LICENSE-2.0.html */
-#'  a.anchor-section::before {
-#'    content: url(https://fonts.gstatic.com/s/i/materialicons/link/v7/24px.svg);
+#'  output:
+#'    html_document:
+#'      anchor_sections:
+#'        style: icon
+#'        depth: 2
+#'  }
+#'  By default, it will be applied on all headers (equivalent of
```
hadley, line 122:
```suggestion
#'  If omitted, anchor will be added to all headers (equivalent of
```

R/html_document.R
```diff
@@ -81,28 +83,50 @@
...
+#'  \code{depth=6}). You can also set anchors manually using \samp{# my header
+#'  {.hasAnchor}} with \code{depth = 0}.
 #'  }}
 #'
-#'  About how to apply custom CSS, see
-#'  \url{https://bookdown.org/yihui/rmarkdown-cookbook/html-css.html}
+#'  Using anchor sections will add some CSS to your document output for the
+#'  styling, and JS script if \code{section_divs = TRUE}. The anchor link itself
```
hadley, line 128:
```suggestion
#'  styling, and a JS script if \code{section_divs = TRUE}. The anchor link itself
```

R/html_document.R
```diff
@@ -81,28 +83,50 @@
...
+#'  is added using a Lua filter - it requires Pandoc 2.0+
```
hadley, line 129:
```suggestion
#'  is added using a Lua filter, and hencerequires Pandoc 2.0+
```

R/html_document.R
```diff
@@ -689,4 +719,49 @@ navbar_link_text <- function(x, ...) {
...
+  # default style used
+  style <- "hash"
+
+    # allowed customization
+  allowed_args <- c("style", "depth")
+  if (is.list(anchor_sections)) {
+    # check list elements
+    all_allowed <- all(names(anchor_sections) %in% allowed_args)
+    if (xfun::isFALSE(all_allowed)) {
+      stop("`anchor_sections` should be TRUE or one of [",
```
hadley, line 741:
```suggestion
      stop("`anchor_sections` should be TRUE or a list with names [",
```
?

R/html_document.R
```diff
@@ -689,4 +719,49 @@ navbar_link_text <- function(x, ...) {
     tagList(x$text, ...)
 }
 
+add_anchor_sections <- function(anchor_sections, section_divs = FALSE) {
+
+  res <- list(args = NULL, lua_filters = NULL, extra_dependencies = NULL)
 
+  if (xfun::isFALSE(anchor_sections)) return(res)
```
hadley, line 726: You also check this above?

R/html_document.R
hadley: It can also be `FALSE` — overall I find the logic of this function a little hard to parse. I think it would be clearer if organised around the three permitted inputs:

* `FALSE`
* `TRUE`
* A list with names in `allowed_args`

This might require extracting out some helper functions to reduce duplication

inst/rmd/h/anchor-sections/README.md
```diff
@@ -0,0 +1,15 @@
...
+From: https://material.io/
+Source: https://github.com/google/material-design-icons
+
+## Icon used
+
+* `link-black-48dp.svg`
+
+## Licence
+
+Apache-2.0 - https://www.apache.org/licenses/LICENSE-2.0.html
```
hadley, line 12: Should also mention in `LICENSE.note` and include the Author(s) in `DESCRIPTION`

inst/rmd/h/anchor-sections/anchor-sections-icon.css
```diff
@@ -0,0 +1,7 @@
+/* Styles for section anchors content */
+/* From https://material.io/resources/icons/
+   Licence: https://www.apache.org/licenses/LICENSE-2.0.html */
+
+a.anchor-section:before {
+  content: url(./link-black-24dp.svg);
```
hadley, line 6: Is it worth inlining the svg here?

R/html_document.R
```diff
@@ -689,4 +719,49 @@ navbar_link_text <- function(x, ...) {
...
+    if ("depth" %in% names(anchor_sections)) {
+      # Add metadata only if required
+      res$args <- pandoc_metadata_arg(
+        "rmd_anchor_depth", anchor_sections[["depth"]]
+      )
+    }
+    # change style if required
+    style <- anchor_sections[["style"]] %||% style
+    if (length(style) != 1) {
+      stop("`anchor_sections` can be a list containing only a single `style` element.",
```
hadley, line 754: Can't you leave argument validation to `html_dependency_anchor_sections()`? 

Currently this error implies that you have a list like `list(style = 1, style = 2)` but I don't think that's actually what you're checking for.

tests/testthat/test-html_dependencies.R
```diff
@@ -0,0 +1,15 @@
+test_that("Correct anchor_sections style is used", {
+  deps <- html_dependency_anchor_sections
+  expect_s3_class(deps(), "html_dependency")
+  expect_error(deps("dummy"))
```
hadley, line 4: Good practice to check the error is the error you expect

tests/testthat/test-html_dependencies.R
```diff
@@ -0,0 +1,15 @@
...
+  expect_match(deps()$stylesheet[[2]], "hash")
+  expect_null(deps()$script[[1]])
+  expect_equal(deps(section_divs = TRUE)$script[[1]], "anchor-sections.js")
+  css <- function(style, id = 2) {
+    deps(style)$stylesheet[[id]]
+  }
+  expect_equal(css("hash", 1), "anchor-sections.css")
+  expect_equal(css("icon",1), "anchor-sections.css")
+  expect_match(css("icon"), "icon")
+  expect_match(css("symbol"), "symbol")
```
hadley, line 14: Overall, given the complexity of `html_dependency_anchor_sections()` this feels like you are mostly testing other people's code.

tests/testthat/test-html_document.R
```diff
@@ -0,0 +1,22 @@
+test_that("anchor_sections adds correct component", {
+  skip_if_not(pandoc2.0())
+  expect_equivalent(add_anchor_sections(FALSE), list(NULL, NULL, NULL))
```
hadley, line 3: This would be a good place for snapshot tests, if you can use them.

R/html_document.R
```diff
@@ -22,8 +22,10 @@
 #'  options that control the behavior of the floating table of contents. See the
 #'  \emph{Floating Table of Contents} section below for details.
 #'@param number_sections \code{TRUE} to number section headings
-#'@param anchor_sections \code{TRUE} to show section anchors when mouse hovers.
-#'  See \link[rmarkdown:html_document]{Anchor Sections Customization section}.
+#'@param anchor_sections \code{TRUE} to show section anchors when mouse hovers
+#'  for all headers. A list can also be passed with \code{style} and/or
+#'  \code{depth} to customize the behavior. See
+#'  \link[rmarkdown:html_document]{Anchor Sections Customization section}.
```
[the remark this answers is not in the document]
cderv: I added that because if case other formats import params for `html_document`. Only this text would be imported but not the below section. So you would need to know to look there for futher details. The link seemed useful in this case.

Not a good idea ?

R/html_document.R
cderv: It is but use a pictogram that could differ according to the browser. We use the symbol version instead by using the special variation selector 
> Emoji-specific variation selector: text, signals intention for preceding glyph to be rendered with basic line stroke rather than complex pictogram.
Source: 
* https://graphemica.com/%EF%B8%8E
* https://codepoints.net/U+1F517

Example: 
image

The documentation was a way to show how to use the pictogram if one desires

R/html_document.R
cderv: Oh yes.... 🤔 must be a duplicate introduced when I refactored the code out of  `html_document`. Thanks!

R/html_document.R
cderv: At first `anchor_sections = FALSE` was meaning "do not run the code of adding anchor section because all was in `html_document` as we do in other places in the body. 

Later on I find it better to refactor in a special helper `add_anchor_sections` but still need to check FALSE before for adding the components (this code: https://github.com/rstudio/rmarkdown/pull/1964/files#diff-93d1e6747a8a5a21e4668d9a67b3889d481426747e8a0f18d7b2ebfb32ee8413R382) 

So I need to choose where to check FALSE (related to your review above) 

1. in `html_document` body
2. in the helper itself 

Maybe 2 is clearer and I should restructure as you are suggesting

R/html_document.R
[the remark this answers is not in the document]
hadley: Oh yeah, we don't have a good solution for this case 😢

R/html_document.R
hadley: Ah, I didn't spot that you were using a specific technical unicode term. Given the additional complexity of explaining all these details, I'd be tempted to pick a different example. (Especially since I think most people would want 🔗).

inst/rmd/h/anchor-sections/README.md
cderv: oh ok. Thanks for the advice.  I was not sure what to do for using this type of content. 

Maybe I should find another icon to use to avoid adding all that just for one icon.
Did you design it yourself for **pkgdown** ?

inst/rmd/h/anchor-sections/anchor-sections-icon.css
cderv: Do you mean not have it as an external file but directly put its content in one line i.e `<svg>...</svg>` ?

Just to be sure of the meaning on "inlining"

tests/testthat/test-html_dependencies.R
cderv: That confirms my feeling. I wondering if snapshot test would be better for testing this kind feature with Rmarkdown file

tests/testthat/test-html_document.R
cderv: I am just looking into testthat 3 for rmarkdown. I think it is important so yes I hope I can use them. 

I understand `expect_equivalent` is deprecated, so I should change it.

inst/rmd/h/anchor-sections/anchor-sections-icon.css
hadley: Well, using a data url, but yes — include all of the svg data in the css file to avoid having a separate file.

inst/rmd/h/anchor-sections/README.md
hadley: Hmmm, no idea 😞 I'm sure I would've got it from a openly licensed source, but I failed to record that information.

R/html_document.R
cderv: I want to check against  this
```r 
expect_error(add_anchor_sections(list(style = c("icon", "dash")), "one of"))
````

My wording is incorrect then.

> Can't you leave argument validation to html_dependency_anchor_sections()?

I put it here at first but thought it was a logic to but outside to keep the dependency function simple as others. 

I'll rethink that.

File: inst/rmd/h/anchor-sections/README.md (hunk 14 above)
Comment (hadley) [reply inst/rmd/h/anchor-sections/README.md:12]: But given that you're only including a small single file, I think you could just put this info in the css/svg file.

File: R/html_document.R (hunk 20 above)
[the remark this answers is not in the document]
Comment (cderv) [reply R/html_document.R:28]: This links I added works to move to `html_document` help page but it does not show the section directly, you would still have to look down to find it. 

Documenting in section is useful to describe our formats' features but we have this difficulties to navigate to them. Maybe I should spend time on vignettes instead, like one per format, that we could link to.

Do you think that would be better than my solution ?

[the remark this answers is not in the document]
Comment (hadley) [reply R/html_document.R:28]: There's no great way to link to a section within a vignette either; except to link to link to the html on the rendered website (which you could also if it's in function documentation). In general, I don't have a great sense for when you should use function documentation vs a vignette. The primary disadvantage of a vignette is that it's quite far from the function so its easy to forget to update it.

**Comment by jooyoungseo** [comment]
@cderv, I am very sorry for this delayed response.

> What would you expect as alt text for the generated content ? Just a description (like "Anchor svg icon")

Yes, "Anchor svg icon" would suffice and that would be better than empty "" alt value.

> This content: alt text is not supported on all browser it seems.

I don't think there should be an error in non-supporting browsers against the alt text; however, we need to test. I was wondering if we could then use a more general solution (e.g., aria-label). If that works, I would go with that.

File: man/figures/link-black-18dp.svg (hunk 2 above)
Comment (cderv) [reply man/figures/link-black-18dp.svg:1]: I used `aria-label`

File: tests/testthat/test-html_document.R (hunk 19 above)
Comment (cderv) [reply tests/testthat/test-html_document.R]: I'll introduce testthat 3 later and not introduce it in this PR. I change for `expect_equal` here.

File: R/html_document.R (hunk 7 above)
Comment (cderv) [reply R/html_document.R:104]: > Especially since I think most people would want 🔗

As it can be rendered differently depending on OS (Ex. of differences https://www.utf8icons.com/character/128279/link-symbol), I wanted to use a default for `symbol` with less variations. If someone wants the unicode icon it can be added using CSS as in example. But for icon, I thought the svg one was a better fit.

Now I am not sure which default is better and I wonder if offering the choice between `symbol` and `icon` is really interesting - they are similar.

Comment (hadley) [reply R/html_document.R:104]: I realised that the other downside of the emoji is that it won't get coloured by css, so I think the best solution really is the svg.

**Comment by cderv** [comment]
BTW @jooyoungseo I added `aria-label` to the link. So it should work as expected when hovering the link.

**Comment by atusy** [comment]
@cderv 

I just come up with the an idea to control `max_level` with JS.

With the `head` argument of the `htmltools::htmlDependency` function, we can insert arbitrary lines of HTML to insert into the document head.
So for example, we can add a metadata about `max_level`.

```r
htmltools::htmlDependency(
  ...,
  head = sprintf("<meta itemprop='anchor-sections-max-level' content='%s'>", max_level)
)
```

Then, we can catch the value with JS by

```js
document.querySelector("meta[itemprop='anchor-sections-max-level']").content
```

atusy:
Oh, the `script` element would be cleaner way than the `meta` element.
An example from MDN:

```html
<!-- Generated by the server -->
<script id="data" type="application/json">{"userId":1234,"userName":"John Doe","memberSince":"2000-01-01T00:00:00.000Z"}</script>

<!-- Static -->
<script>
  const userInfo = JSON.parse(document.getElementById("data").text);
  console.log("User information: %o", userInfo);
</script>
```

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#embedding_data_in_html

cderv:
There is also [anchorjs](https://github.com/bryanbraun/anchorjs) which is used in **distill** already. Did not know that one so sharing it here for later reference.

yihui (approved):
@cderv This PR looked more complicated than I expected, but after another review, I think it looks good overall. I guess you meant "hash" instead of "dash". Other than that, please feel free to merge it. Thanks!

PR #1968, opened 2020-12-01
Default anchor_sections to FALSE and document how to change content.

### LINKED ISSUE #1946: Re-consider a more user-friendly anchor symbol, or make the default FALSE for anchor_sections

cderv:
@yihui as discussed, here is the PR for the upcoming patch release. 

I cherry picked the relevant commit in #1964 to only set `anchor_sections` to FALSE by default, and document better how to customize.

This will be enough for now and we'll then can merge #1964 once finished in a new feature release, after the next CRAN release.

it will close #1946 better as #1964 is rather adding feature.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,7 +15,7 @@ rmarkdown 2.6
   rather than preservation tokens when pandoc >= v2.0. Note that this option will
   have the intended effect only for versions of htmltools >= 0.5.0.9003.
 
-- `anchor_sections` in `html_documents()` now default to `FALSE`. It was introduced in previous version with a default to `TRUE` but it is reverted now after hearing feedbacks from the community. (Thank you!). The `#` is still used as the character for the anchor but you can easily change that using css rule. Examples have been added in `?html_document`. 
+- `anchor_sections` in `html_documents()` now defaults to `FALSE`. It was introduced in previous version with a default to `TRUE`, but it is reverted now after hearing feedbacks from the community (thank you!). The `#` is still used as the character for the anchor but you can easily change that using CSS rules. Examples have been added to the help page `?html_document`.
 
 rmarkdown 2.5
 ================================================================================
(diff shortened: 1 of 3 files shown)


yihui (approved):
Thanks!

PR #1970, opened 2020-12-03
Use Pandoc's default for --email-obfuscation

### LINKED ISSUE #1969: Why is email obfuscation made impossible in html_document?
---

I am referring to this line code: https://github.com/rstudio/rmarkdown/blob/master/R/html_document_base.R#L38. I am making an html template that depends on `html_document()` and I would like to obfuscate linked email addresses in the document. Why is this option always turned off for `html_document()`? Would you accept a change that makes it possible to enable email obfuscation?

cderv:
closes #1969 

Currently, `--email-obfuscation none` is set by `html_document_base()` for Pandoc. However, it can take different value and not only `none`

This is not needed per-se because this works already as last command line flag in Pandoc overwrites previous ones.
```r
  html_document: 
    pandoc_args: ["--email-obfuscation", "javascript"]
```

But it will be cleaner to not have `--email-obfuscation none` as it is the default for Pandoc 1.17.2+

@yihui if you're ok, we could add that for next patch release with no harm

Also, just added a small test to keep the habit of having a good test coverage.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -21,6 +21,15 @@ rmarkdown 2.6
 
 - `anchor_sections` in `html_documents()` now defaults to `FALSE`. It was introduced in previous version with a default to `TRUE`, but it is reverted now after hearing feedbacks from the community (thank you!). The `#` is still used as the character for the anchor but you can easily change that using CSS rules. Examples have been added to the help page `?html_document`.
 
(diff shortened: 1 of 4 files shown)


yihui (commented):
Personally I'd just get rid of `--email-obfuscation` completely in our code base to make it cleaner to maintain. For users who are still using Pandoc <= 1.17.2 (I don't think there are many now), they will have to use `pandoc_args` if they prefer the previous behavior. This will be a tiny breaking change for them, but I don't feel it is going to hit many users.

cderv:
Oh ok that is fine with me. I was under the impression we tried to support without breaking Pandoc version since 1.14
https://github.com/rstudio/rmarkdown/blob/b5e4b9cc30d70fefcb30b67bbef16e13f72a1440/DESCRIPTION#L105

I'll remove it.

cderv:
@yihui I made the change per your review so merging it now

yihui:
Sounds good. Thanks! I think this feature is too small to be worth bumping the Pandoc version in `SystemRequirements`.

PR #1972, opened 2020-12-04
use GHA runner's temp directory as R temp dir

cderv:
This will fix the issue (https://github.com/rstudio/rmarkdown/commit/9635483451dc4f357b0fe220bf482c450d8dd42c#r44719348)  and all subsequent potential ones regarding the use of `tempfile()` on Windows and LaTeX rendering.

By changing the temp directory, we insure that there will be no space in the filename, so no short path created which makes latex failed to compile. (BTW I can now reproduce this issue locally with the ~ in filename) - not sure why it happens because it seems it should have been resolved (https://tug.org/pipermail/tex-live/2014-June/035507.html)

This will also avoid to deal with this issue directly in our test (like in https://github.com/rstudio/rmarkdown/commit/8cc31e5e48658cca2f590474cdc9ec783a72ac03)

I wonder if `r-lib/actions/setup-r` should consider setting the R temp directory to GHA $RUNNER_TEMP. It is already where the R library is set ([example](https://github.com/cderv/rmarkdown/runs/1499096453?check_suite_focus=true#step:8:10)) 🤔

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -33,6 +33,7 @@ jobs:
       R_REMOTES_NO_ERRORS_FROM_WARNINGS: true
       RSPM: ${{ matrix.config.rspm }}
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+      TMPDIR: ${{ runner.temp }}
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


cderv:
For future us, this would also work to set the env var for all following R steps:
```yaml
      - name: Set TMPDIR to change R temp directory
        run: echo "TMPDIR=${{ runner.temp }}" >> $GITHUB_ENV
        shell: bash
```

yihui:
That's great! Thanks!

PR #1973, opened 2020-12-08
clean_site default to preview = TRUE

### LINKED ISSUE #1971: clean_site should default to preview TRUE to prevent deletion without notice
Will close https://github.com/rstudio/blogdown/issues/518

`rmarkdown::clean_site(preview = TRUE)` should be the default with a nice message showing the files that would be deleted and informing what command to run to remove thoses files. 

Example of cleaning function with message: `bookdown::clean_book()`

cderv:
This will close #1971 

* `preview = TRUE` by default now - this will affect the behavior of Clean All button in website project Build pane
* A message is output to console to inform about the file that could be removed. 
* `clean_site(preview = FALSE)` will effectively remove those files
* Nothing is done if no file returned by the generator clean function
* if `quiet = FALSE` and `preview = FALSE`, a message is also output to console - useful for knowing what happens in the Build pane

I also 

* refactored tests and add some for better coverage. 
    * Added **withr** as Suggests because it is really useful for clean test helper.
* adapted `mark_dirs` from **bookdown** - will do a PR in **xfun**

One question remaining: 

* Should `clean_site()` check for existence of the files and directories mentioned in message ? 
* Or should the `generator$clean()` function only return filepath that exists ? 

I think the latter would be best but currently it does not - for example **blogdown**: 
```r
==> rmarkdown::clean_site()

These files and folders can probably be removed:

* blogdown
* public/
* content/_index.markdown
* content/post/2015-07-23-r-rmarkdown/index.html
* static/rmarkdown-libs

Use `rmarkdown::clean_site(preview = FALSE)` to remove them.
```

`blogdown` folder does not exist, `static/rmarkdown-libs` neither (because page bundle I think). I find it puzzling that it is output in the message

What do you think ?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,7 @@
 rmarkdown 2.6
 ================================================================================
 
-- `clean_site()` now default to `preview = TRUE` and will no more remove files without notice. This change will affect the 'Clean All' button in the 'Build' pane for website project. `clean_site(preview = FALSE)` must be run to effectively remove files.
+- `clean_site()` now default to `preview = TRUE` and will no more remove files without notice. This change will affect the "Clean All" button in the "Build" pane for website project. `clean_site(preview = FALSE)` must be run to effectively remove files (#1973).
 
 - The intermediate `.tex` file is now correctly deleted if `keep_tex = FALSE` when the R Markdown document is not rendered from the working directory (thanks, @vqv, #1308).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 rmarkdown 2.6
 ================================================================================
 
+- `clean_site()` now default to `preview = TRUE` and will no more remove files without notice. This change will affect the 'Clean All' button in the 'Build' pane for website project. `clean_site(preview = FALSE)` must be run to effectively remove files.
+
 - The intermediate `.tex` file is now correctly deleted if `keep_tex = FALSE` when the R Markdown document is not rendered from the working directory (thanks, @vqv, #1308).
 
 - Fix a bug causing certain resources files to be deleted as intermediate files when `intermediates_dir` is the same as the input (thanks, @bellma-lilly, #1248). 
--- a/R/render_site.R
+++ b/R/render_site.R
@@ -220,7 +220,6 @@ clean_site <- function(input = ".", preview = TRUE, quiet = FALSE,
   # get the files to be cleaned
   files <- generator$clean()
 
-  length(files) == 0
 
   if (length(files) == 0) {
     if (preview || !quiet) cat("Nothing to removed. All clean !\n")
--- a/R/render_site.R
+++ b/R/render_site.R
@@ -245,8 +245,8 @@ clean_site <- function(input = ".", preview = TRUE, quiet = FALSE,
 
 # TODO: Move to xfun - bookdown use it too.
 mark_dirs <- function(files) {
-  i = file_test("-d", files) & !grepl("/$", files)
-  files[i] = paste0(files[i], "/")
+  i <- file_test("-d", files) & !grepl("/$", files)
+  files[i] <- paste0(files[i], "/")
   files
 }
 
--- a/R/render_site.R
+++ b/R/render_site.R
@@ -230,13 +230,13 @@ clean_site <- function(input = ".", preview = TRUE, quiet = FALSE,
   # actually remove the files
   if (preview) {
     cat("These files and folders can probably be removed:\n",
-        paste0("* ", mark_dirs(files, input)),
+        paste0("* ", mark_dirs(files)),
         "\nUse `rmarkdown::clean_site(preview = FALSE)` to remove them.",
         sep = "\n")
   } else {
     if (!quiet) {
       cat("Removing files: \n",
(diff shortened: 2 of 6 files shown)


cderv:
Also I have opened https://github.com/rstudio/rmarkdown/issues/1974 for the idea we shared last Friday. 

If you find it interesting and good enough in its implementation, we could ship it with this. Otherwise, we have time to think about it.

R/render_site.R
```diff
@@ -220,19 +220,39 @@ clean_site <- function(input = ".", preview = FALSE, quiet = FALSE,
...
 
+# TODO: Move to xfun - bookdown use it too.
+mark_dirs <- function(files, dir) {
+  if (!missing(dir)) {
+    owd = setwd(dir); on.exit(setwd(owd))
+  }
+  i = file_test("-d", files) & !grepl("/$", files)
+  files[i] = paste0(files[i], "/")
+  files
+}
```
cderv, line 254: After some tests, I am not so sure the `dir` argument is useful because `files <- generator$clean()` does not pass the `input` directory to the function. 

That means I don't think `clean_site` can really be used with a directory different than the current working dir. 

Try bookdown-demo. After a first render, you won't get the same result between this
```
# Render book
xfun::in_dir("bookdown-demo/", bookdown::render_book("index.Rmd"))
# No file to remove ? 
rmarkdown::clean_site("bookdown-demo/", preview = TRUE)

# It seems there is
xfun::in_dir("bookdown-demo/", bookdown::clean_book())
```

R/render_site.R
```diff
@@ -204,9 +204,9 @@ render_site <- function(input = ".",
 
 #' @rdname render_site
 #' @param preview Whether to list the files to be removed rather than actually
-#'   removing them.
+#'   removing them. Defaulting to TRUE to prevent removing without notice.
 #' @export
-clean_site <- function(input = ".", preview = FALSE, quiet = FALSE,
+clean_site <- function(input = ".", preview = TRUE, quiet = FALSE,
```
cderv, line 209: Would it be interesting to have an option for `preview` to change the default behavior of the IDE button to FALSE ? 

Not so sure, so I did not add it

R/render_site.R
cderv: I removed the dir argument as discussed 33ef30f

We just need to remember that `clean_site()` only work within working directory really

R/render_site.R
yihui: Yes, that could be useful (e.g., `preview = getOption('rmarkdown.clean_site.preview', TRUE)`), but let's wait until someone asks for it.

R/render_site.R
```diff
@@ -220,19 +220,36 @@ clean_site <- function(input = ".", preview = FALSE, quiet = FALSE,
...
+          sep = "\n")
     }
     unlink(file.path(input, files), recursive = TRUE)
   }
 }
 
+# TODO: Move to xfun - bookdown use it too.
+mark_dirs <- function(files) {
+  i = file_test("-d", files) & !grepl("/$", files)
+  files[i] = paste0(files[i], "/")
```
yihui, line 249:
```suggestion
  i <- file_test("-d", files) & !grepl("/$", files)
  files[i] <- paste0(files[i], "/")
```

yihui (approved):
> * Should `clean_site()` check for existence of the files and directories mentioned in message ?
> * Or should the `generator$clean()` function only return filepath that exists ?

I'd prefer the latter. I'll fix the problem in **blogdown**. Thanks!

NEWS.md
```diff
@@ -1,6 +1,8 @@
 rmarkdown 2.6
 ================================================================================
 
+- `clean_site()` now default to `preview = TRUE` and will no more remove files without notice. This change will affect the 'Clean All' button in the 'Build' pane for website project. `clean_site(preview = FALSE)` must be run to effectively remove files.
```
```suggestion
- `clean_site()` now default to `preview = TRUE` and will no more remove files without notice. This change will affect the "Clean All" button in the "Build" pane for website project. `clean_site(preview = FALSE)` must be run to effectively remove files (#1973).
```

R/render_site.R
```diff
@@ -220,19 +220,36 @@ clean_site <- function(input = ".", preview = FALSE, quiet = FALSE,
...
+
   # if it's just a preview then return the files, otherwise
   # actually remove the files
-  if (preview)
-    files
-  else {
+  if (preview) {
+    cat("These files and folders can probably be removed:\n",
+        paste0("* ", mark_dirs(files)),
+        "\nUse `rmarkdown::clean_site(preview = FALSE)` to remove them.",
```
yihui, line 234:
```suggestion
        "\nUse rmarkdown::clean_site(preview = FALSE) to remove them.",
```

R/render_site.R
cderv: Oh I remember this method of yours now - not used to that: get ideas but wait for people asking ! 😄

R/render_site.R
cderv: arg... copy paste from xfun PR. Change in style guide is always tricky

R/render_site.R
yihui: You can definitely also proactively ask if anyone wants this feature in advance. I just don't have the energy to do this myself :)

Issue #1980, opened 2020-12-14
rmarkdown uses incorrect working directory if filename has spaces and output format is "github_document"


cboettig:
By filing an issue to this repo, I promise that

- [ x ] I have fully read the issue guide at https://yihui.org/issue/.
- [ x ] I have provided the necessary information about my issue.
    - If I'm asking a question, I have already asked it on Stack Overflow or RStudio Community, waited for at least 24 hours, and included a link to my question there.
    - If I'm filing a bug report, I have included a minimal, self-contained, and reproducible example, and have also included `xfun::session_info('rmarkdown')`. I have upgraded all my packages to their latest versions (e.g., R, RStudio, and R packages), and also tried the development version: `remotes::install_github('rstudio/rmarkdown')`.
    - If I have posted the same issue elsewhere, I have also mentioned it in this issue.
- [x ] I have learned the Github Markdown syntax, and formatted my issue correctly.

I understand that my issue may be closed if I don't fulfill my promises.

---

Consider the following reprex (which apparently cannot be parsed by the reprex package either, but that's a different issue)

```r

dir.create("subdir")
writeLines(
'
---
title: "bug report"
output: github_document
---
  
  
```{r}
stopifnot(file.exists("Bug Report.Rmd"))
```
',
"subdir/Bug Report.Rmd")

rmarkdown::render("subdir/Bug Report.Rmd")

```

As you can see, all this does is create a trivial Rmarkdown document in a subdirectory, and then attempt to knit it.  Because the RMarkdown document has a space in the name and we have requested `github_document`, the example errors by having the wrong working directory (the root dir, not the same dir from where the Rmarkdown file is found, as in the usual convention!)  Change the output format to `html_document`, or remove the space in the name, and the error is resolved.

atusy:
Reproduced. It seems the bug originally comes from `md_document` which is the base format of `github_document`.

atusy:
I see issue comes from here, which removes `dirname()` from `input` when input has shell characters.

https://github.com/rstudio/rmarkdown/blob/80f14b2c6e63dcb8463df526354f4cd4fc72fd04/R/render.R#L366

Then, if `intermediates_dir = NULL` as default, the following lines considers input file exists in the current directory.

https://github.com/rstudio/rmarkdown/blob/80f14b2c6e63dcb8463df526354f4cd4fc72fd04/R/render.R#L355-L356

A workaround is to explicitly specifying `intermediates_dir = "subdir"`.
[...]

PR #1982, opened 2020-12-15
Make intermediates_dir explicit earlier to fix 1980

atusy:
See https://github.com/rstudio/rmarkdown/issues/1980#issuecomment-745329485 for why I suggest the change.

### DIFF
--- a/R/render.R
+++ b/R/render.R
@@ -325,6 +325,8 @@ render <- function(input,
     if (!dir_exists(intermediates_dir))
       dir.create(intermediates_dir, recursive = TRUE)
     intermediates_dir <- normalize_path(intermediates_dir)
+  } else {
+    intermediates_dir <- normalize_path(dirname(input))
   }
   intermediates_loc <- function(file) {
     if (is.null(intermediates_dir))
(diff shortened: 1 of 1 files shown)


yihui (approved):
Looks good. Thank you!

atusy:
Oops the test is failing. I'll inspect tommorrow.

yihui (changes requested):
Oh, actually this seems to break some tests in the package (as shown in Github actions), so needs further investigation.

atusy:
OK, finally I passed the tests.

cderv:
Thanks @atusy ! 

Just sharing my thoughts here - not a review.

I did not have look closely yet to the initial issue and your fix but from what I can see these changes are targeting important part of the `render()` code. Paths management are always tricky in **rmarkdown**. 

I've seen in the past that side effects can happen. We should really make sure we cover how it is expected to work before the change to check it still work the same after. I know I added some tests already and I think we should add more to be sure. But there is a lot of edge cases (as the one in the issue - Github document in subdir with space in names). 

Maybe I am a bit too cautious but I find this part of the code very tricky and trying to fix a specific bug by a more generic solution could lead to unexpected behavior. And I got bitten by that already in the past.

I'll find time to look closer.

atusy:
@cderv Thanks. I agree with you. We have to be cautious.

cderv:
I commented in #1980 where I tried to better understand the underlying issue. I need to see clearer what happens, and what should happens. In the meantime, if you are ok I would like to hold this PR. thank you.

PR #1984, opened 2020-12-18
Improvements to latex-div

jjallaire:
- Allow "latex" attribute (as alternative to "data-latex"). Note that the "data-" prefix is an HTML convention and Pandoc already automatically adds a "data-" prefix to HTML attributes it doesn't recognize. Net: for markdown syntax like this we don't need to use "data-" prefixes.

- Allow slightly more intuitive syntax for a latex div with no options: `latex=true` or `latex=1`

- If the leading and trailing blocks in the div are paragraphs than write the environment begin/end flush up to the block text (the extra lines added when inserting raw blocks can make a difference for some environment implementations)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: rmarkdown
 Type: Package
 Title: Dynamic Documents for R
-Version: 2.6.0001
+Version: 2.6.0002
 Authors@R: c(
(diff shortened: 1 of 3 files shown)


cderv (approved):
Looks great, thank you ! `latex=true` or `latex=1` is indeed more intuitive. 

We just need to add a NEWS bullet and bump the version before merging.

jjallaire:
Okay, done!

PR #1993, opened 2020-12-22
Hide header from floating ToC if the header is unlisted and unnumbered

atusy:
To solve https://github.com/rstudio/bookdown/issues/894#issuecomment-749473767


inst/rmd/h/default.html
```diff
@@ -677,7 +677,7 @@ <h2 id="$idprefix$toc-title">$toc-title$</h2>
       hashGenerator: function (text) {
         return text.replace(/[.\\/?&!#<>]/g, '').replace(/\s/g, '_');
       },
-      ignoreSelector: ".toc-ignore",
```
cderv, line 680: I believe you added `.unlisted.unnumbered *` to handle the headers when classes is on the section divs. 

This is done for `toc-ignore ` by moving the class the header. See above (https://github.com/rstudio/rmarkdown/pull/1993/files#diff-91b59569aab1b937e1e5a6e1559a62de19e04f4c4bad52bfb08b910281e0f2dfL667) 

I don't know if that cause an issue to ignore everything under a div of class - I don't think so but this is not currently how it works with `toc-ignore` so I am wondering. I think it would be better to use the same trick no matter the classes. Easier to maintain. 

Also, BTW, the classes are not on the header because we remove it in `header-attrs.js` to insure compatibility with Pandoc<2.8 - at the end if seems odd in this kind of case but understandable to keep a similar HTML structure to manipulate no matter Pandoc version.

cderv (commented):
Currently the way to ignore a header using toc float if to use `.toc-ignore` class to the header. That ways this header and its sub headers will be ignored from the floating toc. 

So currently I see the feature as 

* `.unnumbered .unlisted` when using Pandoc's TOC
* `.toc-ignore` when using floating TOC (which disable Pandoc's TOC)

So I don't see that as an issue which need a fix but more as a Feature request: 
* Allow the same class to ignore header in TOC when using toc float or not. 

It does not seem so important, but thinking this way, as find that another solution would be to make those classes equivalent.

* If `toc_float: true`, detect Pandoc's way `.unnumbered .unlisted` and change to tocify's way `.toc-ignore`
* If `toc_float: false`, detect tocify's way `.toc-ignore` and change to `.unnumbered .unlisted`

This could be easy to do using a lua filter I think. Maybe it could also be easier to maintain.

This would make using one or the other usable. I would maybe do only the former so that we could advertise only Pandoc's way (https://bookdown.org/yihui/rmarkdown-cookbook/toc-unlisted.html) and `.toc-ignore` would work only when floatting toc is used. 

What do you think ?

inst/rmd/h/default.html
atusy: Sorry, `.unlisted.unnumbered *` comes from my misunderstanding. I confirmed things work well even if removing that selector.

cderv: Oh I was not sure it would work... 🤔 
Why do we need to move `.toc-ignore` to headers then ? 

I'll check something later

atusy:
Yes, this can be considered as a feature request. I'll change the NEWS.md.

> - toc_float: true, detect Pandoc's way .unnumbered .unlisted and change to tocify's way .toc-ignore
> - If toc_float: false, detect tocify's way .toc-ignore and change to .unnumbered .unlisted

I would also do the former and advertise only Pandoc's way.
I also considered about Lua filter, but decided not to because of simplicity.
This PR just tweaks the option of `tocify.js`.
If we employ Lua filter, then `tocify.js`-related codes become sporadic.
I think this will regress maintainability.

inst/rmd/h/default.html
atusy: Oops, sorry. It's my mistake...
I was using `.toc-ignore` instead of `.unlisted .unnumbered`.
Yes, we need to move `.toc-ignore` to headers.

atusy: Now I modified the JavaScript to add `.toc-inore` selector to headers if they are children of `div.unlisted.unnumbered`.

inst/rmd/h/default.html
```diff
@@ -677,7 +677,7 @@ <h2 id="$idprefix$toc-title">$toc-title$</h2>
...
+      ignoreSelector: ".toc-ignore, .unlisted.unnumbered",
```
cderv, line 680: You can just ignore `.toc-ignore` with the change you did above ? 

(Which is BTW a clever one and indeed more scoped than by lua ! Thanks)

atusy: We need to ignore `.unlisted.unnumbered` when `section_divs = false`.

cderv: Sorry I mistyped the first time. I mean is this still needed ?

cderv: Oh I see. That is why I found all those JS trick difficult to understand later on. We need to handle special cases in several places all the time. 
That means 

* section_divs = FALSE => we will have .unlisted.unumbered on all headers and no .toc-ignore
* section_divs = TRUE => we sill have .toc-ignore on all header and .unlisted .unnumbered in the sections divs above. 

All this does not help for future us !

cderv: Moreover because we loose `.unlisted .unnumbered` on headers because we remove it ourselves in `header-attr.js` for newer Pandoc. 🤕

atusy: OK, I simplified the trick by adding `.toc-ignore` to `.unlisted.unnumbered` with jQuery.
What do you think?

atusy: A limitation is when section div has `.unlisted.unnumbered.toc-ignore` prior to the `tocify.js`.
In this case, `.toc-ignore` becomes duplicated, and the one remains in the final DOM.

atusy: If we allow more complex selector, i.e. `.unlisted.unnumbered:not(.toc-ignore)`, we can avoid the limitation.

cderv: I was thinking about the cleanest way and that may be just commenting to explain more (about this section divs trick)

But this last solution is easier to understand I think! Also the diff is really minimal as it is just one line addition! We will add `.toc-ignore` to all headers in the HTML though instead of handling directly the classes as option in your first solution. 

Sorry for all the thinking, I like to search for the different solution to a problem. 😉 

This is fine by me. Let's see what Yihui thinks about this.

cderv (approved):

inst/rmd/h/default.html
atusy: Great. Thanks for your deep thoughts. It really helped me to come up with this solution. I like it too!

cderv:
@yihui we discussed this new feature with @atusy to come up with this last solution of just one line addition. 

The situation we want to improve is described in #1995

Several solutions: 

1. Configure tocify to also ignore `.unlisted .unnumbered` classes and not only `.toc-ignore`
1. Do not change configuration but detect `.unnumbered .unlisted` classes and add `.toc-ignore` class so that tocify will ignore the header 
1. Replace `.unnumbered .unlisted` classes by `.toc-ignore` class (similar as above)

We decided to do 2. and do it with JS directly. A lua filter would also have been possible (to handle class modification). 
Only drawback I see with solution 2. is that we will add `.toc-ignore` in addition of `.unnumbered .unlisted` instead of just dealing with the latter (which was first @atusy idea, modified following shared thoughts in dicussion above)

At last, this feature only try to make possible the use of `.unnumbered .unlisted` we [documented ](https://bookdown.org/yihui/rmarkdown-cookbook/toc-unlisted.html) with floating toc which currently only supports `.toc-ignore`. 
It does not make `.toc-ignore` works with non-floating toc. It is simpler than adding two classes but it is not really documented I think, whereas the other is documented in our doc AND in Pandoc's docs.

I let you merge in case you have a strong opinion on how this should be done.

yihui (approved):
Looks good to me. Thanks!

PR #1996, opened 2020-12-22
exclude renv folder from render_site() copied resources

jmbuhr:
This makes sure it follows the same behavior as e.g. for `packrat` and `rsconnect`.


cderv:
Thanks @jmbuhr ! It seems indeed something to support !

cderv (approved):

PR #1998, opened 2020-12-23
use only the TinyTeX-1 version in CI

### LINKED ISSUE #1986: Fix CI on windows
TinyTeX installation issue... 
Don't know why yet... 🤔

cderv:
maybe a lighter binary will be less prone to errors to fix #1986


cderv:
Let's merge and see if the issue in #1986 happens again with this bundle

PR #2001, opened 2020-12-30
use on.exit to remove file

MichaelChirico:
safer than executing at end of function

### DIFF
--- a/tests/testthat/test-formats.R
+++ b/tests/testthat/test-formats.R
@@ -4,13 +4,13 @@ test_that("formats successfully produce a document", {
 
   testFormat <- function(output_format, df_print = NULL) {
     output_file <- I(tempfile())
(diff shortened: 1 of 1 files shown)


tests/testthat/test-formats.R
```diff
@@ -4,13 +4,13 @@ test_that("formats successfully produce a document", {
 
   testFormat <- function(output_format, df_print = NULL) {
     output_file <- I(tempfile())
+    on.exit(unlink(output_file))
```
cderv, line 7:
```suggestion
    on.exit(unlink(output_file), add = TRUE)
```

cderv (approved):
Sure, we could have used `on.exit()` here in the first place. But as it is a temp file here, I believe it it will be removed anyway after tests. 

Thanks for the suggestion !

PR #2014, opened 2021-01-20
Add shiny to remotes

cpsievert:
In #1706, I was hoping that `bslib`&`shiny` would both be on CRAN by today, but that's highly unlikely (it might not be until next week that both are on CRAN).


yihui (approved):
Thanks!

PR #2017, opened 2021-01-21
Warn that cropping is disabled if pdfcrop or ghostscript are not found.

cderv:
here is a proposal to do that on **rmarkdown** side.

It is oriented specifically toward cropping tools for which we want the warning. I had the idea to make it more generic but we don't need it for now so I stop myself from doing it.

Other solutions could be by adapting `find_program()` so that it warns, or using (exporting ?) `knitr:::has_utility()` but the logic is not exactly the same as `find_program()` (which deals with OSX a specific way). Opportunity to refactor ? 

Anyway, here is a PR to have you thoughts before merging.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,6 +15,7 @@ rmarkdown 2.7
 
 - Do not force `options(htmltools.preserve.raw = TRUE)` when this option has been set, otherwise it is impossible for other packages to turn this option off, e.g., yihui/xaringan#293.
 
+- `knitr_options_pdf()` will now throw a warning when `fig_crop = TRUE` but is disabled because required tools _pdfcrop_ and/or _ghostscript_ are missing. (thanks, @netique, #2016)
 
 rmarkdown 2.6
 ================================================================================
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,7 +15,7 @@ rmarkdown 2.7
 
 - Do not force `options(htmltools.preserve.raw = TRUE)` when this option has been set, otherwise it is impossible for other packages to turn this option off, e.g., yihui/xaringan#293.
 
-- `knitr_options_pdf()` will now throw a warning when `fig_crop = TRUE` but is disabled because required tools _pdfcrop_ and/or _ghostscript_ are missing. (thanks, @netique, #2016)
+- `knitr_options_pdf()` will now throw a warning when `fig_crop = TRUE` but is disabled because required tools `pdfcrop` and/or `ghostscript` are missing (thanks, @netique, #2016).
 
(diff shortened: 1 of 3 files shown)


R/util.R
```diff
@@ -312,6 +312,23 @@ find_program <- function(program) {
   }
 }
 
+has_crop_tools <- function() {
+  tools <- list(
```
yihui, line 316:
```suggestion
  tools <- c(
```

R/util.R
```diff
@@ -312,6 +312,23 @@ find_program <- function(program) {
...
+  missing <- tools[tools == ""]
+  if (length(missing) > 0) {
+    x <- paste0(names(missing), collapse = ", ")
+    warning(
+      sprintf("\nTool(s) not installed or not in PATH: %s", x),
+      "\n-> As a result, cropping will be disabled."
+    )
+    return(FALSE)
+  }
+  TRUE
```
yihui, line 329:
```suggestion
  if (length(missing) == 0) return(TRUE)
  x <- paste0(names(missing), collapse = ", ")
  warning(
    sprintf("\nTool(s) not installed or not in PATH: %s", x),
    "\n-> As a result, figure cropping will be disabled."
  )
  FALSE
```

yihui (approved):
The PR looks good enough to me. Thanks!

I'm not sure if the peculiarity of `which` on macOS is still there. Someday we might be able to get rid of the hack https://github.com/rstudio/rmarkdown/blob/dbc4a68c7acff83ef3df0a628c6a637c60761863/R/util.R#L293-L309

I've been using `Sys.which()` myself and haven't run into this problem. My guess is that we originally had this hack mainly because of the nonstandard path of MacTeX (which is something like `/usr/texbin`). TinyTeX's binaries are symlinked to the standard `/usr/local/bin` so it won't suffer from this problem.

NEWS.md
```diff
@@ -15,6 +15,7 @@ rmarkdown 2.7
 
 - Do not force `options(htmltools.preserve.raw = TRUE)` when this option has been set, otherwise it is impossible for other packages to turn this option off, e.g., yihui/xaringan#293.
 
+- `knitr_options_pdf()` will now throw a warning when `fig_crop = TRUE` but is disabled because required tools _pdfcrop_ and/or _ghostscript_ are missing. (thanks, @netique, #2016)
```
yihui, line 18:
```suggestion
- `knitr_options_pdf()` will now throw a warning when `fig_crop = TRUE` but is disabled because required tools `pdfcrop` and/or `ghostscript` are missing (thanks, @netique, #2016).
```

PR #2040, opened 2021-02-09
set gfm_auto_identifiers to input format if toc is TRUE

### LINKED ISSUE #2039: github_document() should set +gfm_auto_identifiers to the input for TOC to work correctly
This is from an investigation following a question on Community
https://community.rstudio.com/t/github-document-table-of-contents-links-not-working-for-sections-that-begin-with-numbers/95539/3

[At first I thought](https://community.rstudio.com/t/github-document-table-of-contents-links-not-working-for-sections-that-begin-with-numbers/95539/2?u=cderv) it was an issue with Pandoc not supporting the correct algorithm even for GFM, but in fact I misunderstood how the auto identifier feature works. 

From [this anwser on Pandoc dicussion](https://groups.google.com/g/pandoc-discuss/c/c3LEmJEQCkk/m/5Oq6r2_SBwAJ), it seems that the auto identifier creation happens at read time, not write time. This seems counter intuitive but I put theses files togther in this gist: https://gist.github.com/cderv/589912cb000347cf6d22e91adfa852bc

As you see, not providing `markdown+gfm_auto_identifiers` will not produce the correct identifiers to work in the TOC for Pandoc autolink feature.

This means this works

````markdown
---
title: "Test"
output:
  github_document:
    md_extension: +gfm_auto_identifiers
    toc: true
---

# 1 Section

Sentence.
```` 

and this will produce a non working TOC
````markdown
---
title: "Test"
output:
  github_document:
    toc: true
---

# 1 Section

Sentence.
```` 

I think we should set the extension in **rmarkdown** if `toc = TRUE`

### PR DESCRIPTION
**Pull request by cderv**
This is required to have the correct identifier in the TOC corresponding to the link Github will create. 

This will fix #2039

### DIFF
--- a/tests/testthat/test-github_document.R
+++ b/tests/testthat/test-github_document.R
@@ -2,23 +2,20 @@
 test_that("toc has correct identifier", {
   local_edition(3)
   skip_on_cran() # avoid pandoc issue on CRAN
-  tmp_file <- withr::local_tempfile()
-  xfun::write_utf8(
-    c("# 10 Section","","Sentence.","", "# Header","","Sentence "),
-    tmp_file)
+  skip_if_not(pandoc_available("2.10.1"))
+  tmp_file <- local_rmd_file(
+    c("# 10 Section","","Sentence.","", "# Header","","Sentence ")
+  )
   res <- render(tmp_file, github_document(toc = TRUE, html_preview = FALSE))
-  if (pandoc_available("2.10.1")) {
-    expect_snapshot_file(res, "github-toc.md", binary = FALSE)
-  }
+  expect_snapshot_file(res, "github-toc.md", binary = FALSE)
 })
 
 
 test_that("github_document produces atx-header", {
   local_edition(3)
   skip_on_cran() # avoid pandoc issue on CRAN
-  tmp_file <- withr::local_tempfile()
   h <- paste0(Reduce(paste0, rep("#", 5), accumulate = TRUE), " title ", 1:5, "\n\n")
-  xfun::write_utf8(h, tmp_file)
+  tmp_file <- local_rmd_file(h)
   res <- render(tmp_file, github_document(html_preview = FALSE))
   expect_snapshot_file(res, "github-atx.md", binary = FALSE)
 })
--- a/tests/testthat/test-github_document.R
+++ b/tests/testthat/test-github_document.R
@@ -1,18 +1,20 @@
 
-local_edition(3)
-
 test_that("toc has correct identifier", {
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by cderv** [comment]
@yihui is it ok for you to start using [snapshot testing](https://testthat.r-lib.org/articles/snapshotting.html) using **testthat** ? 

Or do you prefer regex-testing like in https://github.com/rstudio/rmarkdown/blob/de0e2ec4ddaca455b6d67fea9553d2da1d2103c0/tests/testthat/test-lua-filters.R#L16 ? 

I wanted to try the former and found it was a good match for output that give us textual content to compare like markdown content.

File: tests/testthat/test-github_document.R [hunk 1]
```diff
@@ -0,0 +1,26 @@
+
+test_that("toc has correct identifier", {
+  local_edition(3)
+  skip_on_cran() # avoid pandoc issue on CRAN
+  tmp_file <- withr::local_tempfile()
```
hadley, line 5: Rather than the `if` statement below, I'd recommend `skip_if_not_pandoc("2.10.1")` (creating the function as necessary); that way the test will be recorded as being skipped so that you know it hasn't run.

tests/testthat/test-github_document.R
```diff
@@ -0,0 +1,26 @@
...
+  xfun::write_utf8(
```
hadley, line 6: If you use this pattern a lot, you might consider writing something like this:

```R
local_rmd <- function(..., .env = parent.frame()) {
  path <- withr::local_tempfile(envir = .env)
  xfun::write_utf8(c(...), path)
  path
}
```

tests/testthat/test-github_document.R
cderv: Oh indeed. At first I wanted to make two snapshots : one for Pandoc 2.10.1+ and one for earlier version. 
This would require a `if` or two `test_that()` with different `skip_*`
However it is currently not easy to generate snapshot for different pandoc versions.

tests/testthat/test-github_document.R
cderv: Is `tests/testthat/helpers.R` still a good place ? 
I see there is a new recommandation to use `R/test-helpers.R`.

hadley: It doesn't really matter, but I now prefer somewhere inside `R/`

cderv: Ok. I put it in the former as I was used to that. Thanks

yihui (approved):
Looks good. Thanks!

PR #2046, opened 2021-02-14
Defer JavaScript scripts in html_document

### LINKED ISSUE #2032: anchor_sections: true doesn't appear to work with runtime: shiny
For example

````yaml
---
title: "My First R Markdown Document" 
runtime: shiny
output:
  html_document:
    anchor_sections: true
---

# R Markdown

## Overview

R Markdown combines markdown (an easy to write plain text format) with embedded
R code chunks.
````

### PR DESCRIPTION
**Pull request by atusy**
This PR defers JavaScript scripts that are currently called as DOMContentLoaded event handlers.

Although this works fine in static `html_document`, it still fails in `runtime: shiny`.
This is probably because `src` is converted wrongly.
See highlighted line and the last line in the following image.

image

## Reprex

````
---
title: "My First R Markdown Document" 
output:
  html_document:
    anchor_sections: true
    self_contained: false
runtime: shiny
---

# fooo
````

(closes #2032)

### DIFF
--- a/R/html_dependencies.R
+++ b/R/html_dependencies.R
@@ -160,7 +160,10 @@ html_dependency_accessible_code_block <- function() {
     name = "accessible-code-block",
     version = "0.0.1",
     src = pkg_file("rmd/h/accessibility"),
-    script = "empty-anchor.js"
+    script = list(
+      src = "empty-anchor.js",
+      defer = NA
+    )
   )
 }
 
@@ -207,11 +210,13 @@ html_dependency_navigation <- function(code_menu, source_embed) {
 }
 
 html_dependency_anchor_sections <- function() {
-
   htmlDependency(name = "anchor-sections",
                  version = "1.0.1",
                  src = pkg_file("rmd/h/anchor-sections"),
-                 script = "anchor-sections.js",
+                 script = list(
+                   src = "anchor-sections.js",
+                   defer = NA
+                 ),
                  stylesheet = "anchor-sections.css")
 }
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by yihui** [comment]
> This is probably because `src` is converted wrongly.

```html
<script src="header-attrs-2.6.6/%5Bobject%20Object%5D"></script>
```

is this when decoded:

```html
<script src="header-attrs-2.6.6/[object Object]"></script>
```

It seems when the `defer` attribute is present, the `src` value is calculated by adding the string `"header-attrs-..."` to an object like

```js
> "header-attrs-.../" + { "src": "foo.js", "defer": null}

"header-attrs-.../[object Object]"
```

I don't know where this string concatenation occurs. We will need help from the Shiny team (perhaps @cpsievert).

cpsievert:
Ok, rstudio/shiny#3345 should be fixed on the main branch now, so the `<script defer>` approach shouldn't be a blocker anymore (although, it's not entirely clear to me that `defer` is the best way to solve this issue -- you might want to consider using `setTimeout()` instead)

atusy:
@cpsievert Thanks a lot!

> `setTimeout()`

Would you tell me some background of this suggestion?
I think it is quite difficult to predict the adequate value for it.

atusy:
Maybe the best way is to

1. check if `document.readyState` is `complete`
    - this is what happens by `runtime: shiny`
2. if true, run function immediately or use `setTimeout`
3. otherwise, `addEventListener` as usual.

PR #2049, opened 2021-02-18
Add shiny::bootstrapLib() as a dependency inside html_document_base()'s pre-processor

cpsievert:
In #1706, to make `bslib::bs_themer()` work with `runtime: shiny` I added `shiny::bootstrapLib()` as a dependency when `run()` is first called. This ends up being is too early for output formats that wish to pass a _modified_ `theme`  to `html_document_base()` (the new version of `flexdashboard::flex_dashboard` will do this).

If you care for an example, first install https://github.com/rstudio/flexdashboard/pull/308, then render a [flexdashboard example](https://github.com/rstudio/flexdashboard/pull/308). Note that with this PR will make it so the navbar's bg color reflects `$primary` (instead of `$dark`)


cpsievert:
Ahh, in order for `runtime: shiny_prerendered` to work with `bslib::bs_themer()`, we also need the `shiny::bootstrapLib()` dependency added at `run()`-time, so I'll be undo-ing those deletions momentarily.

yihui (approved):
@cderv Usually I don't review PRs related to shiny and would just merge them as is. If you want to take a closer look, please feel free to, otherwise I'm okay if you just merge it. Thanks!

R/html_document_base.R
```diff
@@ -89,10 +89,16 @@ html_document_base <- function(theme = NULL,
...
+      # theme was set globally pre-knit, and it may be modified during knit
+      if (is_bs_theme(theme)) {
+        theme <- bslib::bs_global_get()
+      }
+      bootstrap_deps <- if (is_bs_theme(theme) && is_shiny(runtime)) {
+        list(shiny_bootstrap_lib(theme))
+      } else {
+        bootstrap_dependencies(theme)
+      }
+      format_deps <- append(format_deps, htmltools::resolveDependencies(bootstrap_deps))
```
cderv, line 101: @cpsievert just curious why do we need `htmltools::resolveDependencies` exactly ?
It is for my understanding.

cderv: Is this in case someone modifies the theme from inside the document ?

cderv (approved):
LGTM ! 

Thanks for your patience on this one @cpsievert and thanks for the work on **bslib** in **rmarkdown** !!

R/html_document_base.R
cpsievert: It's because `shiny::bootstrapLib()` returns a `htmltools::tagFunction()` which needs `htmltools::resolveDependencies()` to be called on it in order to generate the dependency. This is really just be being paranoid about a `dependency_resolver` being specified that doesn't call `htmltools::resolveDependencies()`

cderv: OK great thanks! still learning all this details. 👍

R/shiny.R
```diff
@@ -134,16 +134,21 @@ run <- function(file = "index.Rmd", dir = dirname(file), default_file = NULL,
   yaml_front <- if (length(target_file)) yaml_front_matter(target_file)
   runtime <- yaml_front$runtime
   theme <- render_args$output_options$theme
```
cderv, line 136: @cpsievert as I am going through shiny prerendered code base to understand better how it works, I am just wondering what this line is used for ? It does not seem to be used now with the PR changes. 

Do you recall the aim of catching a special handling when `theme` would be passed inside `rmarkdown::run(renderArgs=)` ?

R/shiny.R
```diff
@@ -134,16 +134,21 @@ run <- function(file = "index.Rmd", dir = dirname(file), default_file = NULL,
...
+  # Let shiny::getCurrentTheme() know about the yaml's theme, so
+  # things like `bslib::bs_themer()` can work with prerendered documents.
+  # Also note that we add the actual shiny::bootstrapLib() dependency
+  # inside the document's pre-processing hook so the 'last' version of
+  # the theme wins out
   if (length(target_file)) {
     format <- output_format_from_yaml_front_matter(read_utf8(target_file))
```
cderv, line 143: @cpsievert Regarding my question in https://github.com/rstudio/rmarkdown/pull/2049/files#r586418699, I wonder if `render_args$output_options` should not be passed to this function

I believe this should be 
```suggestion
format <- output_format_from_yaml_front_matter(read_utf8(target_file), render_args$output_options)
```

This way if theme provided in `render_args$output_options` in `rmarkdown::run` is will correctly be merged with any other option set in YAML header.

I missed that at the first review but it is my understanding now. 

I want to confirm the intention in https://github.com/rstudio/rmarkdown/pull/2049/files#r586418699 before doing any modification. Thanks.

cpsievert: Ahh, yes, thanks for catching that! If you happen to include a fix in another PR, feel free to tag me on it

cderv: Ok. So I'll push a fix for that then. I believe we can remove the line above also.

cderv: See #2065

PR #2050, opened 2021-02-18
Fix for knit_print.data.frame

### LINKED ISSUE #2047: knit_print.data.frame shoud not pass ... to print()
This is the result of an investigation with @DavisVaughan following a CRAN check error with tibbletime https://cran.r-project.org/web/checks/check_results_tibbletime.html and 12 more packages

CRAN email is : 
```
CRAN packages with unused arguments error
That is
     DriveML SmartEDA anomalize bayestestR broom effectsize heemod
     isoreader mlrCPO panelr parameters tibbletime
These show very similar errors in vignette re-building on Solaris and
macOS (which do not have pandoc so test your package without): you may
wish to discuss this with the knitr maintainer.
```

We manage to reproduce the error using 

```r
library(rmarkdown)
knitr::opts_chunk$set(error = FALSE)
knitr::knit("vignettes/TT-01-time-based-filtering.Rmd")
```

which has this traceback

<details>
<summary>traceback</summary>

```r
> traceback()
21: print.tbl_time(x, ...)
20: print(x, ...)
19: knit_print.data.frame(x, ...)
18: knit_print(x, ...)
17: withVisible(knit_print(x, ...))
16: fun(x, options = options)
15: value_fun(ev$value, ev$visible)
14: withVisible(value_fun(ev$value, ev$visible))
13: withCallingHandlers(withVisible(value_fun(ev$value, ev$visible)), 
        warning = wHandler, error = eHandler, message = mHandler)
12: handle(pv <- withCallingHandlers(withVisible(value_fun(ev$value, 
        ev$visible)), warning = wHandler, error = eHandler, message = mHandler))
11: evaluate_call(expr, parsed$src[[i]], envir = envir, enclos = enclos, 
        debug = debug, last = i == length(out), use_try = stop_on_error != 
            2L, keep_warning = keep_warning, keep_message = keep_message, 
        output_handler = output_handler, include_timing = include_timing)
10: evaluate::evaluate(...)
9: evaluate(code, envir = env, new_device = FALSE, keep_warning = !isFALSE(options$warning), 
       keep_message = !isFALSE(options$message), stop_on_error = if (options$error && 
           options$include) 0L else 2L, output_handler = knit_handlers(options$render, 
           options))
8: in_dir(input_dir(), evaluate(code, envir = env, new_device = FALSE, 
       keep_warning = !isFALSE(options$warning), keep_message = !isFALSE(options$message), 
       stop_on_error = if (options$error && options$include) 0L else 2L, 
       output_handler = knit_handlers(options$render, options)))
7: block_exec(params)
6: call_block(x)
5: process_group.block(group)
4: process_group(group)
3: withCallingHandlers(if (tangle) process_tangle(group) else process_group(group), 
       error = function(e) {
           setwd(wd)
           cat(res, sep = "\n", file = output %n% "")
           message("Quitting from lines ", paste(current_lines(i), 
               collapse = "-"), " (", knit_concord$get("infile"), 
               ") ")
       })
2: process_file(text, output)
1: knitr::knit("vignettes/TT-01-time-based-filtering.Rmd")
```

</details>

Without loading **rmarkdown** first, there is no error.

We believe this happens because `rmarkdown:::knit_print.data.frame` is called and pass `...` to `print()`
https://github.com/rstudio/rmarkdown/blob/1707d0b5189e3f12caf16a6cd53391f5be2b2f08/R/knit_print.R#L1-L32

These `...` contains `knit_print` arguments `inline` and `options`
This will then trigger `print.tbl_time` passing those two arguments that throws the unused arguments error. 

This error was hidden for now because it only happens if **rmarkdown** is loaded and we think it is now because of https://github.com/yihui/knitr/commit/fb2f427903814d75ed939705d20aefb25d8ccc08#diff-d0d9f9e52333dc5483c4882b0923ed3e0d31de7534bb8601d3b992998698abc5
* Before pandoc-citeproc was check, and was not there so FALSE was return with `rmarkdown::pandoc_available` being called. Now `knitr::pandoc_available` will load **rmarkdown** and the method will be registered. 

One fix would be to not pass `...` to `print` from `knit_print` so that `inline` and `options` are not passed to unknown `print` method that don't know about these arguments. 

Unless I missed something and this is expected.

@yihui pinging you as this cause some CRAN check failure for others packages.

cderv:
This should fix #2047 

it will insure to not pass `...` to `print()` methods. `knit_print()` accepts `...` but only for `options` and `inline` that can be used by `knit_print` method. 

This two arguments should not be passed to underlying `print()` method. Some of them won't deal with `...` and an `unused argument` error will be thrown.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -23,6 +23,8 @@ rmarkdown 2.7
 
 - `github_document()` will produce a working TOC even if some headers start with number (#2039)
 
+- Fix and issue with `knit_print.data.frame`. The `...` are no more passed to `print()` to avoid passing `knit_print()` arguments `options` and `encoding` to custom `print()` methods. (#2047) 
+
 rmarkdown 2.6
 ================================================================================
 
--- a/R/knit_print.R
+++ b/R/knit_print.R
@@ -27,7 +27,7 @@ knit_print.data.frame <- function(x, ...) {
       context$df_print(x)
     }
   } else {
-    print(x, ...)
+    print(x)
   }
 }
 
--- a/tests/testthat/test-knit_print.R
+++ b/tests/testthat/test-knit_print.R
@@ -1,0 +1,8 @@
+test_that("knit_print.data.frame default to print without error", {
+  df <- data.frame(a = 10)
+  class(df) <- c("df", class(df))
+  print.df <- function(x) {cat("Nb row:", nrow(x)) }
+  expect_output(knit_print.data.frame(df), "Nb row:")
+  expect_error(knit_print.data.frame(df, inline = TRUE), NA)
+  expect_output(knit_print.data.frame(df, inline = TRUE), "Nb row:")
+})
--- a/tests/testthat/test-knit_print.R
+++ b/tests/testthat/test-knit_print.R
@@ -1,8 +1,4 @@
 test_that("knit_print.data.frame default to print without error", {
   df <- data.frame(a = 10)
-  class(df) <- c("df", class(df))
-  print.df <- function(x) {cat("Nb row:", nrow(x)) }
-  expect_output(knit_print.data.frame(df), "Nb row:")
   expect_error(knit_print.data.frame(df, inline = TRUE), NA)
-  expect_output(knit_print.data.frame(df, inline = TRUE), "Nb row:")
 })


yihui (approved):
The fix looks good to me. Feel free to merge by yourself after it passes the tests. Thanks!

PR #2054, opened 2021-02-24
Rework tabset-dropdown to use regular Bootstrap dropdowns

### LINKED ISSUE #1697: .tabset-dropdown's menu right icon doesn't render properly with self_contained: false
Screen Shot 2019-11-18 at 10 46 48 AM

For a minimal example, put [this minimal example](https://github.com/rstudio/rmarkdown/pull/1405) in a rmarkdown website

sdiebolt:
This PR reworks the `.tabset-dropdown` class to work similarly to regular tabsets. It should also close #1697.

---

I wasn't satisfied with the current `.tabset-dropdown` class and wanted to use regular [Boostrap dropdowns included in navs](https://getbootstrap.com/docs/3.3/components/#nav-dropdowns). This lead me to remove the `.tabset-dropdown` CSS and improve the `buildTabsets` JS function to include the `.tabset-dropdown` class.

The `.tabset-dropdown` class can now be added to any sub-header of a header with class `.tabset` or `.tabset-pills`. The `tabset-dropdown` header will determine the dropdown's name and all of its sub-headers will be added as dropdown tabs.

## Example

```markdown
---
title: "RMarkdown Dropdown"
output: 
  html_document:
    self_contained: false
    theme:
      version: 4
---

# First header

Lorem Ipsum

# Dropdown {.tabset .tabset-fade}

## Dropdown {.tabset-dropdown}

### Choice 1

Dropdown 1

### Choice 2

Dropdown 2

### Choice 3

Dropdown 3

## Tab 1

Tab 1

## Tab 2

Tab 2

## Tab 3

Tab 3

# Second header

Lorem Ipsum
```

Screenshot 2021-02-24 at 10 24 37

### DIFF
--- a/inst/rmd/h/navigation-1.1/tabsets.js
+++ b/inst/rmd/h/navigation-1.1/tabsets.js
@@ -1,5 +1,3 @@
-
-
 /**
  * jQuery Plugin: Sticky Tabs
  *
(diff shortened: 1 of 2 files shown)


yihui (approved):
I love this idea (and I'm glad that it could fix #1697), but I didn't review the implementation carefully. I'll let @cderv review and test it later. Thanks!

cderv:
For reference while reviewing this PR, **bslib** is in the process of coming up with new UI function for creating navs & tabs : https://github.com/rstudio/bslib/issues/296 
It is partly already there: https://rstudio.github.io/bslib/reference/index.html#section-create-navs-and-navbars 

I believe this will be possible to use it in R Markdown document using Bootstrap also.

inst/rmd/h/navigation-1.1/tabsets.js
```diff
@@ -75,58 +73,155 @@ window.buildTabsets = function(tocID) {
...
+
+          // get the heading element within it, grab it's text, then remove it
+          var dropdownHeading = dropdownTab
+            .find('h' + dropdownTabLevel + ':first');
+          var dropdownHeadingText = dropdownHeading.html();
+          dropdownHeading.remove();
+
+          // build and append the dropdown tab list item
+          var dropdownLi = $('<li role="presentation"></li>');
+          dropdownMenu.append(dropdownLi);
```
cderv, line 163: Could we make a JS function as this is the same process as for other tabs above but with different parameter ? Refactoring would help later maintaining

sdiebolt:
Thanks for pointing out the new features from **bslib**, they will help me a lot!

Please note that I am not a regular JS user: feel free to be as critical with my implementation as needed.

inst/rmd/h/navigation-1.1/tabsets.js
```diff
@@ -75,58 +73,155 @@ window.buildTabsets = function(tocID) {
...
+          dropdownTab.addClass('tabbed-pane');
+          if (fade)
+            dropdownTab.addClass('fade');
+
+          // move it into the tab content div
+          dropdownTab.detach().appendTo(tabContent);
+        });
+      } else {
+        // build and append the tab link
+        var a = $('<a role="tab" data-toggle="tab">' + tabHeadingText + '</a>');
```
cderv, line 191:
```suggestion
        // build and append the tab link
        var a = $('<a role="tab" data-toggle="tab" role="button" aria-expanded="false">' + tabHeadingText + '</a>');
```

Same as above

cderv (commented):
Thanks for this PR. 

I made very few comments on the code. Overall, it seems to work fine. 
I test with **bslib** and without. 

What I am not sure about is what the previous behavior was supposed to be. 
I tested your example without the PR (using dev rmarkdown) and I got a very different result - it puzzles me because I did not know about this hidden feature of rmarkdown 😅  
But being different, this would mean than any user already using `.tabset-dropdown` would have its document broke with this PR probably. However, the result of this PR seems closer to Bootstrap dropdown's in tabsets.

Also, we are now in a mix of Boostrap 3 and Bootstrap 4 support, with different version : brought by **rmarkdown** itself or by **bslib**. That is a lot of things to test and the HTML come written by the JS script is a mix of BS3 and BS4 syntax to me. 

I asked @cpsievert to have a look and give its thoughts on this one. We need to think this through correctly on what it is supposed to be. 

@yihui this tabsets feature is something we may consider doing using Lua filter maybe instead of using only JS to post process the HTML directly. Changing to a Fenced div syntax is also an option. 

For the record, while reviewing this PR, I can now confirm that the tabset feature is only working currently with `--section-divs` activated for Pandoc with HTML 4 output. (`section_divs = TRUE` in `html_document()`).

cpsievert:
> I asked @cpsievert to have a look and give its thoughts on this one. We need to think this through correctly on what it is supposed to be.

I would be more conservative here and not intentionally break the existing `.tabset-dropdown` functionality and instead add a new `.dropdown` (or similar) class to implement this new behavior.

On a larger note, if the plan is to require Bootstrap 4+ sometime in the near future, maybe it's time to use the [same lua filters as quarto](https://github.com/quarto-dev/quarto-cli/blob/main/src/resources/filters/quarto-pre/tabset.lua) (or bslib via [remarker](https://github.com/rstudio/remarker)) to implement tabsets. The bslib way would be able to support Bootstrap 3 out-of-the-box, but the quarto way would probably require Bootstrap 4+ (unless, at run-time, you have some way of knowing what version in relevant).

cderv:
> I would be more conservative here and not intentionally break the existing .tabset-dropdown functionality and instead add a new .dropdown (or similar) class to implement this new behavior.

We could indeed try to support both using a new class. 🤔 I am really wondering how many people use this feature that was never documented. It see a few though : https://github.com/search?q=.tabset-dropdown+extension%3ARmd&type=Code&ref=advsearch&l=&l=

> On a larger note, if the plan is to require Bootstrap 4+ sometime in the near future, 

Using **bslib** by default is a plan yes. Default to 4 for `html_document()` would be another topic - We are talking of being conservative for tabsets, so I guess we must be for Bootstrap version too. We are starting thinking about it and we'll track this in a new issue.

> maybe it's time to use the same lua filters as quarto

The syntax for tabsets in Quarto is different (using fenced div) so this would break the current usage. And Quarto does not support all the option for tabset that are supported in R Markdown (likes pills or fade). 
So it would require to adapt the filter to be conservative on feature and syntax. 
But yes, as in my first comment, Lua filter is clearly a good way to implement such feature.  

> bslib via remarker. The bslib way should be able to support Bootstrap 3 out-of-the-box, but the quarto way would probably require Bootstrap 4+ (unless, at run-time, you have some way of knowing what version in relevant).

I'll ask Winston about this project to know more. Being able in some way to sure **bslib** directly without re-doing any bootstrap feature directly would be great. 

Thanks for your feedback @cpsievert

yihui:
> @yihui this tabsets feature is something we may consider doing using Lua filter maybe instead of using only JS to post process the HTML directly. Changing to a Fenced div syntax is also an option.

@cderv I don't care about the implementation. Either JS or Lua is fine to me. I'm also okay with changing the syntax. This is an undocumented feature and doesn't seem to be widely used. If a breaking change can make it easier to maintain in the future (and more compatible with quarto), I'm okay with making a breaking change.

Issue #2082, opened 2021-03-27
Improve or drop support for old Pandoc


atusy:
I just happened to know that I fail to build vignette with Pandoc 2.0.1 because `pandoc.utils` reuired by `pagebreak.lua` is available since Pandoc 2.0.6: https://github.com/atusy/ftExtra/pull/59/checks?check_run_id=2208657510

A reprex that fails to be built with Pandoc < 2.0.6:

```yaml
---
title: "Untitled"
output: html_document
---
```

> Error running filter /home/runner/work/_temp/Library/rmarkdown/rmarkdown/lua/pagebreak.lua:
...work/_temp/Library/rmarkdown/rmarkdown/lua/pagebreak.lua:18: module 'pandoc.utils' not found:

I understand we can fallback so to skip `pagebreak.lua` for older Pandoc versions.
However, I think enough time has passed to drop such very old versions, especially Pandoc 1.x.
Dismiss my suggestion if there is any reason that **rmarkdown** still supports Pandoc >= 1.14.

Anyway, I would also suggest that `pkg_file_lua` should be updated so that we can skip filters for old Pandoc versions.
For example, maybe we can update the function so that each filters can require minimal Pandoc version by their names.

```r
pkg_file_lua(c("2.0.6" = "pagebreak.lua"))
```

<details><summary>xfun::session_info('rmarkdown')</summary>

```
R version 4.0.3 (2020-10-10)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 20.04 LTS, RStudio 1.4.1106

Locale:
  LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
  LC_MONETARY=en_US.UTF-8    LC_MESSAGES=C              LC_PAPER=en_US.UTF-8       LC_NAME=C                 
  LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

Package version:
  base64enc_0.1.3   digest_0.6.27     evaluate_0.14     glue_1.4.2        graphics_4.0.3    grDevices_4.0.3   highr_0.8        
  htmltools_0.5.1.1 jsonlite_1.7.2    knitr_1.31        magrittr_2.0.1    markdown_1.1      methods_4.0.3     mime_0.10        
[...]

PR #2088, opened 2021-04-01
Add Pandoc version requirement in Lua Filters

atusy:
This PR requires Lua filters to be run on Pandoc >= 2.1.
The filters are skipped if requirements are unmet.

<details><summary>outdated initial comment</summary>
There are currently 3 filters available in **rmarkdown**

1. `latex-div.lua`
2. `number-sections.lua`
3. `pagebreak.lua`

As far as I understand, 1 and 2 requires Pandoc 2.0.0 and 3 requires Pandoc 2.0.6.
However, there are no way to check the version in Lua filter Pandoc < 2.1.0.
I decided to comment minimal requested version in the filters.
I also tweaked `pagebreak.lua` so that it does nothing when `pandoc.utils` is missing.
</details>

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -31,6 +31,7 @@ jobs:
           - {os: macOS-latest,   pandoc: '2.7.3',   r: 'release'}
           - {os: ubuntu-18.04,   pandoc: '2.7.3',   r: 'oldrel',  rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
           - {os: ubuntu-18.04,   pandoc: '2.7.3',   r: 'devel',   rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest", http-user-agent: "R/4.0.0 (ubuntu-16.04) R (4.0.0 x86_64-pc-linux-gnu x86_64 linux-gnu) on GitHub Actions" }
+          - {os: ubuntu-18.04,   pandoc: '2.0.1',   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
           - {os: ubuntu-18.04,   pandoc: '2.7.3',   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
           - {os: ubuntu-18.04,   pandoc: '2.11.3.1',  r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
           - {os: ubuntu-18.04,   pandoc: 'devel',   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -31,7 +31,8 @@ jobs:
           - {os: macOS-latest,   pandoc: '2.7.3',   r: 'release'}
           - {os: ubuntu-18.04,   pandoc: '2.7.3',   r: 'oldrel',  rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
           - {os: ubuntu-18.04,   pandoc: '2.7.3',   r: 'devel',   rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest", http-user-agent: "R/4.0.0 (ubuntu-18.04) R (4.0.0 x86_64-pc-linux-gnu x86_64 linux-gnu) on GitHub Actions" }
-          - {os: ubuntu-18.04,   pandoc: '2.0.1',   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
+          - {os: ubuntu-18.04,   pandoc: '2.0',   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
+          - {os: ubuntu-18.04,   pandoc: '2.0.0.1',   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
           - {os: ubuntu-18.04,   pandoc: '2.7.3',   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
           - {os: ubuntu-18.04,   pandoc: '2.11.3.1',  r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
           - {os: ubuntu-18.04,   pandoc: 'devel',   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -31,7 +31,6 @@ jobs:
           - {os: macOS-latest,   pandoc: '2.7.3',   r: 'release'}
           - {os: ubuntu-18.04,   pandoc: '2.7.3',   r: 'oldrel',  rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
           - {os: ubuntu-18.04,   pandoc: '2.7.3',   r: 'devel',   rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest", http-user-agent: "R/4.0.0 (ubuntu-18.04) R (4.0.0 x86_64-pc-linux-gnu x86_64 linux-gnu) on GitHub Actions" }
-          - {os: ubuntu-18.04,   pandoc: '2.0',   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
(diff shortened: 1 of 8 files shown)


atusy:
Thanks @cderv for the comment https://github.com/rstudio/rmarkdown/issues/2082#issuecomment-811909826

> Pandoc 2.1 for using Lua filters as I don't think this is very sensible because should be a few users (if none) 2.0 and 2.0.6.

I'm facing some trouble with supporting Pandoc < 2.1, so I will follow your suggestion.

cderv:
Yes I'll review it. @atusy is the PR ready ? I was under the impression you were still working on it. 
Don't hesitate to open as draft and change to ready when you are ready for a review.

atusy:
@yihui @cderv Thanks. I'm still working on this. I turned the PR to draft, and will be back soon.

atusy:
Okay, I think I'm ready now.

Note thtat R CMD Check on macOS has not finished, and is keep saying "Starting worflow run..."
https://github.com/rstudio/rmarkdown/runs/2246561590

yihui (approved):
Okay, the alternative implementation looks even better to me.

cderv:
I'll look at it as soon as possible and merge it. Thanks !

inst/rmarkdown/lua/latex-div.lua
```diff
@@ -4,6 +4,11 @@
      License: Public domain
 --]]
 
+if (not PANDOC_VERSION) or (PANDOC_VERSION < "2.1") then
+    print("latex-div.lua requires Pandoc > 2.1")
```
cderv, line 8: After some thoughts I would use `io.stderr:write` for these. We may also choose a layout that make it unique and recognizable as a warning / log string

Something like 
```
[WARNING](from latex-div.lua) requires Pandoc > 2.1. Lua filter has been ignored and some features disabled
```
What do you think ? 
```suggestion
    io.stderr:write("[WARNING] (latex-div.lua) requires at least Pandoc 2.1")
```

cderv (changes requested):
Thanks ! I wanted to check for how PANDOC_VERSION works. 
This would be clean but it raises an error which we don't want.
```
PANDOC_VERSION:must_be_at_least '2.1'
```
Ignoring is ok I would just use `stderr` instead of `stdout` after some thoughts.
What do you think ? 

Also, we have a vignette about Lua Filter that we may need to update to add about the version requirement.

.github/workflows/R-CMD-check.yaml
```diff
@@ -31,6 +31,7 @@ jobs:
           - {os: macOS-latest,   pandoc: '2.7.3',   r: 'release'}
           - {os: ubuntu-18.04,   pandoc: '2.7.3',   r: 'oldrel',  rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
           - {os: ubuntu-18.04,   pandoc: '2.7.3',   r: 'devel',   rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest", http-user-agent: "R/4.0.0 (ubuntu-16.04) R (4.0.0 x86_64-pc-linux-gnu x86_64 linux-gnu) on GitHub Actions" }
+          - {os: ubuntu-18.04,   pandoc: '2.0.1',   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
```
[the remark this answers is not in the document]
atusy: because 2.0 provides `pandoc-2.0-3-amd64.deb`, which does not match `pandoc-{{version}}-1-amd64.deb`.
However, I just noticed 2.0.0.1 is the minimal version that matches the above format.
Maybe 2.0.0.1 is the choice.

https://github.com/jgm/pandoc/releases?after=2.1

inst/rmarkdown/lua/latex-div.lua
atusy: I like it

.github/workflows/R-CMD-check.yaml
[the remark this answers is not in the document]
cderv: Oh I see it is a `r-lib/actions/setup-pandoc` limitation. 

It seems Pandoc 2.0.0 has not the same format. Earlier versions seems to have one. We could also look at other os. 

What is the aim of adding this version ? Testing that the tests are skipped ? 
I was under the impression you suggested to add a CI for Pandoc < 2.0 initially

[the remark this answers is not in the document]
atusy: I'm currently testing 2.0 and 2.0.0.1 with GitHub Actions.
Let's employ the minimal install-able version.
61fb7e9

.github/workflows/R-CMD-check.yaml
```diff
@@ -31,6 +31,8 @@ jobs:
           - {os: macOS-latest,   pandoc: '2.7.3',   r: 'release'}
           - {os: ubuntu-18.04,   pandoc: '2.7.3',   r: 'oldrel',  rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
           - {os: ubuntu-18.04,   pandoc: '2.7.3',   r: 'devel',   rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest", http-user-agent: "R/4.0.0 (ubuntu-18.04) R (4.0.0 x86_64-pc-linux-gnu x86_64 linux-gnu) on GitHub Actions" }
+          - {os: ubuntu-18.04,   pandoc: '2.0',   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
```
cderv, line 34: But this won't work as you explained ? File is not in the good format for the action to download. 
https://github.com/jgm/pandoc/releases/tag/2.0

Also I don't want us to test all Pandoc version. We are running CI regularly and it becomes a lot of running times (and resources). 

Why is our aim with that ? 

We could also add a Weekly or Monthly run dedicated to Pandoc version on only one version of R. This would reduce the load for each commit or PR.

inst/rmarkdown/lua/latex-div.lua
cderv: To add to that, I am wondering if it would be useful to add more info on the message. like 

````
io.stderr:write("[WARNING] (latex-div.lua) requires at least Pandoc 2.1. Lua Filter has been ignored and some features hence disabled."
```

Too much ?

.github/workflows/R-CMD-check.yaml
atusy: I feel it is reasonable to test minimal 2.x version.
At the same time, I agree it is better to avoid testing too many versions for every commits and PRs.
The addition of 2.0 is the temporary trial, and I will remove it.
I am okay to remove 2.0.0.1 as well.

.github/workflows/R-CMD-check.yaml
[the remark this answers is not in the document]
atusy: Sorry, I missed your latest comment.
About Pandoc < 2.0, I've tested it but raising error which is not related to lua filter.
I think we should fix it and add to test in the different PR.

> Error: Error: pandoc document conversion failed with error 83
> https://github.com/atusy/rmarkdown/pull/2/checks?check_run_id=2219792064

inst/rmarkdown/lua/latex-div.lua
atusy: I agree previous warning message is not kind enough because it does not explain what happens.
I think we can simply say "skipping" the filer, which implies disabling some features.

```
io.stderr:write("[WARNING] Skipping (latex-div.lua) which requires at least Pandoc 2.1."
```

.github/workflows/R-CMD-check.yaml
cderv: yeah we can add one. I'll open a new issue to rediscuss which version we want to test. 

I did 2.7.3 mainly because it was the default in the RStudio IDE but it has now upgraded.

inst/rmarkdown/lua/latex-div.lua
cderv: Like the requirement first which is the cause, then the effect it has. 
Maybe 
```
io.stderr:write("[WARNING] (latex-div.lua) requires at least Pandoc 2.1. Lua Filter skipped")
```

.github/workflows/R-CMD-check.yaml
atusy: great. thanks for your neat comments.

inst/rmarkdown/lua/latex-div.lua
atusy: got it, and done it by 2105f6d.

cderv:
Thanks! 

As a sum up: 

* `pagebreak.lua` requires Pandoc >= 2.1 because it uses `pandoc.utils`
* Other lua filter have no specific requirement **but** we made a requirement for Pandoc >= 2.1  because we use `PANDOC_VERSION` in every filter ?

Is that correct ? 

I'll merge the change after some last tweaks and we'll see if we have feedback from people stuck between 2.0 and 2.1. 

After looking it is between 2017-10-29 and 2018-01-07 - so it is 2 months only. It should be fine. Nobody reported issue as Pandoc should have error with the pagebreak filter.

cderv (approved):

atusy:
Thank you, too, for the nice reviews!
And yes, your sum up is correct!!

jeroen:
I think this PR is the cause of vignette errors: https://github.com/rstudio/rmarkdown/issues/2099

PR #2092, opened 2021-04-02
Add an option to opt-out the special message for RStudio preview

cderv:
This will allow to deactivate printing of this message from other packages before calling rmarkdown::render().

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: rmarkdown
 Type: Package
 Title: Dynamic Documents for R
-Version: 2.7.5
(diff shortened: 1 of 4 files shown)


cderv:
I called it `rmarkdown.rstudio.preview`, do you have a better idea ? 

Other canditates were: 
* `rmarkdown.render.message`
* `rmarkdown.output.message`
* `rmarkdown.render.output_message`

yihui (approved):
I don't have a strong opinion on the name. All three candidates sound good. Perhaps `rmarkdown.render.message`?

PR #2093, opened 2021-04-02
Fix broken links by number-sections.lua on markdown flavors

### LINKED ISSUE #2091: Bug in table of contents links (markdown file)
---

Dear,

I noticed a bug when generating a README.md file. I noticed that when using the `rmarkdown::github_document` output format and selecting the `number_sections: true` and` toc: true` options in the Preamble YAML, the table of contents is generated but the links do not point to the sections and subsections of the text. I don't know if this can be resolved with some minor changes to the functions [pandoc.R](https://github.com/rstudio/rmarkdown/blob/master/R/pandoc.R), [md_document.R](https://github.com/rstudio/rmarkdown/blob/master/R/md_document.R) or in the codes [Lua](https://github.com/rstudio/rmarkdown/tree/master/inst/rmarkdown/lua).

Below is an Rmd file in which you can see that the output md does not generate links that point to the respective sections and subsections:

```r
---
title: "Title"
author: | 
  |
  | **Summary**
output: 
  rmarkdown::github_document:
    number_sections: true
    toc: true
toccolor: 'blue'
editor_options: 
  markdown: 
    wrap: 72
---

# Any section

`r lorem::ipsum(paragraphs = 2)`

`r lorem::ipsum(paragraphs = 2)`

## Any subsection

`r lorem::ipsum(paragraphs = 2)`

# Any section

`r lorem::ipsum(paragraphs = 2)`

## Any subsection

`r lorem::ipsum(paragraphs = 2)`

`r lorem::ipsum(paragraphs = 2)`
# Any section

## Any subsection
```

Using Pandoc directly on the Linux terminal, I can perfectly generate README.md by doing: `pandoc -f gfm --toc -s README.md -o output.md`. Consider README.md to be:

```r
# Any section

`r lorem::ipsum(paragraphs = 2)`

`r lorem::ipsum(paragraphs = 2)`

## Any subsection

`r lorem::ipsum(paragraphs = 2)`

# Any section

`r lorem::ipsum(paragraphs = 2)`

## Any subsection

`r lorem::ipsum(paragraphs = 2)`

`r lorem::ipsum(paragraphs = 2)`
# Any section

## Any subsection
```
The `output.md` file will be generated with the correct links.

Other information:

```r
> xfun::session_info('rmarkdown')
R version 4.0.5 (2021-03-31)
Running under: Manjaro Linux, RStudio 1.4.1103

Locale:
  LC_CTYPE=pt_BR.UTF-8       LC_NUMERIC=C               LC_TIME=pt_BR.UTF-8       
  LC_COLLATE=pt_BR.UTF-8     LC_MONETARY=pt_BR.UTF-8    LC_MESSAGES=pt_BR.UTF-8   
  LC_PAPER=pt_BR.UTF-8       LC_NAME=C                  LC_ADDRESS=C              
  LC_TELEPHONE=C             LC_MEASUREMENT=pt_BR.UTF-8 LC_IDENTIFICATION=C       

Package version:
  base64enc_0.1.3   digest_0.6.27     evaluate_0.14     glue_1.4.2        graphics_4.0.5   
  grDevices_4.0.5   highr_0.8         htmltools_0.5.1.1 jsonlite_1.7.2    knitr_1.31       
  magrittr_2.0.1    markdown_1.1      methods_4.0.5     mime_0.10         rlang_0.4.10     
  rmarkdown_2.7.4   stats_4.0.5       stringi_1.5.3     stringr_1.4.0     tinytex_0.31     
  tools_4.0.5       utils_4.0.5       xfun_0.22         yaml_2.2.1       

Pandoc version: 2.11.4
```

atusy:
This PR attempts the fix as titled by

1. preprocess number_sections.lua, and record original id of headers as Divs
2. fix broken links to *numbered* headers pandoc process, and remove the Divs added by preprocess


yihui (approved):
Seems to be okay to me, although I'm not quite sure if I fully understand it---my understanding is that the section numbers have to be added *before* the TOC is generated; is that correct?

I'll let @cderv review and merge. Thanks!

yihui:
@atusy No worries, and no hurry! :)

atusy:
Thanks. I forgot to answer you question

> my understanding is that the section numbers have to be added before the TOC is generated; is that correct?

Yes, exactly.

atusy:
I also find `number_sections` break Pandoc's `implicit_header_references` extension both in the master branch and in this PR.
Maybe I will open a new sparate issue and work on it later with additional Lua filter.

````md
# foo

[foo]
````

becomes

````md
# 1 foo

[foo](#foo)
````

but it should actually be

````md
# 1 foo

[foo](#1-foo)
````

**Comment by cderv** [comment]
@atusy I am waiting for this PR to be undraft before looking closely. 

For reference, it seems the difficulty is linked to what I discovered in https://github.com/rstudio/rmarkdown/issues/2039: `gfm_auto_identifiers` needs to be set on the input as identifiers are created at read time. It is why when setting a toc, Pandoc will use identifiers read and not consider title modified in AST during a Lua filter. It is a shame that it works this way for us... 
It now seems to involve several processing and filter just to add numbers to heading...

Maybe a `Make section numbered` addin would be simpler to add numbers to heading in the source file directly 😅 😉

**Comment by atusy** [comment]
Thanks for the comments. Let me think about tests before I undraft this PR.

> It is a shame that it works this way for us...

True...

> Maybe a Make section numbered addin would be simpler to add numbers to heading in the source file directly

Do you mean we do the things on R side instead of Lua filter?
I do not want to dive in to the hell of regex...

**Comment by atusy** [comment]
I integrated pre- and post-processing filters to `number-sections.lua`.
In this way, render prints as if running the filter ordinarily for users.

```
/usr/lib/rstudio-server/bin/pandoc/pandoc +RTS -K512m -RTS example.utf8.md --to gfm --from markdown+autolink_bare_uris+tex_math_single_backslash+gfm_auto_identifiers --output example.md --lua-filter /home/rstudio/R/pkg/rmarkdown/inst/rmarkdown/lua/number-sections.lua --standalone --table-of-contents --toc-depth 3 --template /home/rstudio/R/pkg/rmarkdown/inst/rmarkdown/templates/github_document/resources/default.md
```

I'll work on tests tomorrow.

**Comment by atusy** [comment]
If I run tests locally, I face a following error with a traceback.
It seems like the Lua filter in pre_processor fails to find `pandocAvailable`...
Do you have any ideas to fix?

```
Error running filter /home/rstudio/R/pkg/rmarkdown/inst/rmarkdown/lua/number-sections.lua:
...o/R/pkg/rmarkdown/inst/rmarkdown/lua/number-sections.lua:24: attempt to call a nil value (global 'pandocAvailable')
stack traceback:
	...o/R/pkg/rmarkdown/inst/rmarkdown/lua/number-sections.lua:24: in main chunk
── Error (Line 5): number_sections Lua filter works ────────────────────────────────────────────
Error: pandoc document conversion failed with error 83
Backtrace:
 1. global::.generate_md_and_convert(rmd, md_document(number_sections = TRUE))
 2. .render_and_read(input_file, output_format = output_format)
 3. rmarkdown::render(...) tests/testthat/helpers.R:10:2
 4. output_format$pre_processor(...) R/render.R:834:6
 5. rmarkdown::pandoc_convert(...) R/md_document.R:89:6
```

**Comment by atusy** [comment]
@cderv 
If I run tests locally, I face a following error with a traceback.
It seems like the Lua filter in pre_processor fails to find `pandocAvailable`...
Do you have any ideas to fix?

```
Error running filter /home/rstudio/R/pkg/rmarkdown/inst/rmarkdown/lua/number-sections.lua:
...o/R/pkg/rmarkdown/inst/rmarkdown/lua/number-sections.lua:24: attempt to call a nil value (global 'pandocAvailable')
stack traceback:
	...o/R/pkg/rmarkdown/inst/rmarkdown/lua/number-sections.lua:24: in main chunk
── Error (Line 5): number_sections Lua filter works ────────────────────────────────────────────
Error: pandoc document conversion failed with error 83
Backtrace:
 1. global::.generate_md_and_convert(rmd, md_document(number_sections = TRUE))
 2. .render_and_read(input_file, output_format = output_format)
 3. rmarkdown::render(...) tests/testthat/helpers.R:10:2
 4. output_format$pre_processor(...) R/render.R:834:6
 5. rmarkdown::pandoc_convert(...) R/md_document.R:89:6
```

**Comment by cderv** [comment]
This means that the env var `RMARKDOWN_LUA_SHARED` does not contain the expected path to the shared script. I don't know yet which change in this PR makes it fail whereas test passes on Master. Previous test with pagebreak is running ok. That is odd - there must be a difference.

File: R/md_document.R [hunk 1]
```diff
@@ -80,11 +80,22 @@ md_document <- function(variant = "markdown_strict",
...
+      pandoc_convert(
+        input_file, to = "markdown", output = input_file,
+        options = c(
+          "--standalone",
+          "--lua-filter", pkg_file_lua("number-sections.lua"),
+          "--metadata", "preprocess_number_sections=true"
+        )
+      )
+      return(character(0L))
+    }},
```
cderv, line 98: Yes this is because of this.
We set the expected env var in `rmarkdown::render()` - so when runnning the filter in a preprocessor step with pandoc convert, it is not set. 

Maybe it would have been better to set it in `pkg_file_lua()` directly. See long discussion and different choices in https://github.com/rstudio/rmarkdown/pull/2102

Right now, this is missing this to work: 
https://github.com/rstudio/rmarkdown/blob/9036f864d989d06b966c99494521e521b479c700/R/render.R#L884-L888

As documented in the Lua Filters now, this env var needs to be set in order to used the `.lua` files that uses needs to use `shared.lua`. 

We can revisit for sure.

cderv (changes requested):
I think I have found why: `.lua` filters in rmarkdown are expected to be used as part of a output format processed with `render()` which set the correct env var. 

In case of this PR, you use it in a preprocessor step - variables won't be set.

R/md_document.R
atusy: Thank you for your insights!!

Well, as @yihui's https://github.com/rstudio/rmarkdown/pull/2102#discussion_r610737639, we need to unset the env var after the filter is run. If we `set_envvar()` in `pkg_file_lua()`, we cannot unset in `pkg_file_lua()`, but in the higher level. This makes codes complicated...

I think I can work around the problem by using `render` instead of `pandoc_convert`. I'll try it.

cderv: I did not have a close look at your code here yet so I guess you are force to do this two processing. 
One specific case is that you can also set the env in your preprocessor for this specific needs as you are using a filter that needs it. 

Maybe some wrappers a-la **withr** can come handy here: 
```r
with_shared_lua(pandoc_convert(...))
```
would run the command after setting the env and would remove it after. 

We can just also set the env var earlier in `render()` so that it is available when executing the pre processor function. That would be the easiest I think. It would not be set only if there is a lua filter to use though (this is why it is at its place right now) and the env var would be set anytime as soon as `run_pandoc = TRUE`. I think it could fix the PR, isn't it ?

atusy: Yeah, that is true, and using `render` sounds overkill. Ill do that way later

R/md_document.R
```diff
@@ -60,6 +60,19 @@ md_document <- function(variant = "markdown_strict",
...
+  # add pre_processor to fix link IDs when using numbering sections
+  pre_processor <- if (number_sections) {function(metadata, input_file, ...) {
+    markdown_format <- md_document(
+      variant = "markdown",
+      preserve_yaml = TRUE,
+      pandoc_args = c("--metadata", "preprocess_number_sections=true")
+    )
+    markdown_format$pandoc$lua_filters <- pkg_file_lua("number-sections.lua")
+    render(input_file, markdown_format, quiet = TRUE)
+    warning(paste(xfun::read_utf8(input_file), collapse="\n"))
```
yihui, line 72: What's this `warning()` for?

atusy: Oops, sorry that comes from my debugging

yihui: No worries. You are not alone: https://twitter.com/xieyihui/status/1382100305588932608

Next time you could try something shorter like this :)

```r
warning(xfun::file_string(input_file))
```

R/md_document.R
```diff
@@ -60,16 +60,37 @@ md_document <- function(variant = "markdown_strict",
...
+  pre_processor <- if (
+    number_sections
+    && grepl("^(commonmark|gfm|markdown)", variant)
+    && grepl("+gfm_auto_identifiers", md_extensions, fixed = TRUE)
+  ) {
+    function(metadata, input_file, ...) {
+      lua_env_vars <- xfun::set_envvar(
+        c(RMARKDOWN_LUA_SHARED = pkg_file_lua("shared.lua"))
+      )
+      on.exit(xfun::set_envvar(lua_env_vars), add = TRUE)
```
cderv, line 73: Just as a reminder to maybe reconsider before merging: 

the other solution is to set the variable earlier in render so that it is set during pre-processing for cases like this. This would avoid duplication of these lines in the code and I don't think it would be harmful to have this env var set during the rendering even if not used.

From https://github.com/rstudio/rmarkdown/pull/2093#discussion_r614779967

```diff
diff --git a/R/render.R b/R/render.R
index dea53b19..be5aba06 100644
--- a/R/render.R
+++ b/R/render.R
@@ -827,6 +827,10 @@ render <- function(input,

   if (run_pandoc) {

+    # set env vars required during Pandoc processing
+    lua_env_vars <- xfun::set_envvar(c(RMARKDOWN_LUA_SHARED = pkg_file_lua("shared.lua")))
+    on.exit(xfun::set_envvar(lua_env_vars), add = TRUE)
+
     perf_timer_start("pre-processor")

     # call any pre_processor
@@ -882,8 +886,6 @@ render <- function(input,

       # if Lua filters are provided, add the command line switch
       if (!is.null(lua_filters <- output_format$pandoc$lua_filters)) {
-        lua_env_vars <- xfun::set_envvar(c(RMARKDOWN_LUA_SHARED = pkg_file_lua("shared.lua")))
-        on.exit(xfun::set_envvar(lua_env_vars), add = TRUE)
         lua_filters <- pandoc_lua_filter_args(lua_filters)
       }
       pandoc_args <- c(lua_filters, pandoc_args)
```

R/md_document.R
atusy: Wpw, thanks for your tip!

atusy:
OK, I think I'm finally ready for the review.
This work was further tough than my initial expectation.

R/md_document.R
atusy: Yeah, I like the idea. If it is fine for @yihui, I'll include the change in this PR.

yihui: That's fine to me. Thanks for asking!

R/md_document.R
yihui: Or if warnings are suppressed somehow, you may try `cat(x, file = stderr())`. No one can suppress our debugging messages!!

R/md_document.R
atusy: Done via db49c58

cderv (approved):
Going back to this, it looks good. 

i'll do a few tweaks and merge ! thanks!

cderv:
In fact @atusy I have one proposition for you. I was going to merge but I felt something was not correct for me. It took me time to pin point it and find if an alternative is possible.

I don't really like the fact that Div is used to store the attributes as a Div is something modifying the structure IMO. Div is a block structure like headers and this requires wrapping a block in another block to store an attributes for the inner block. This wrapping of header in Div can conflict with the way Pandoc uses div to wrap headers AND content when using section divs. It may not have impact here for markdown conversion but it is does seem hacky to me to use a div to wrap a header to store an attributes.

The best solution would be to set the temp attributes on the header directly, but I understand you did not do that because it would then also set the identifiers in the resulting markdown file with no way to know which one would have been user provided or not. 

So my question and suggestion is : what not use Spans instead of Divs as Spans are among Inlines. We could then store the attributes within the Header structure as Blocks can containes Inlines. 

Basically my suggestion is similar to this markdown
````markdown
# [Header 1]{data-rmarkdown-temporarily-recorded-id="header-1"}
````
instead of 
````markdown
::: {data-rmarkdown-temporarily-recorded-id="header-1"}
# Header 1
:::
```` 

I believe doing this means this modification 

````diff
diff --git a/inst/rmarkdown/lua/number-sections.lua b/inst/rmarkdown/lua/number-sections.lua
index 727286e24..0f5650774 100644
--- a/inst/rmarkdown/lua/number-sections.lua
+++ b/inst/rmarkdown/lua/number-sections.lua
@@ -48,17 +48,6 @@ function is_preprocess(meta)
 end


-function pairwise_link_id(div)
-  local old_id = div.attributes['data-rmarkdown-temporarily-recorded-id']
-
-  if not preprocess and old_id and #div.content and (div.content[1].t == 'Header') then
-    needs_number_sections = false
-    identifiers['#' .. old_id] = '#' .. div.content[1].identifier
-    return div.content
-  end
-end
-
-
 function fix_link_id(link)
   if not preprocess and identifiers[link.target] then
     link.target = identifiers[link.target]
@@ -68,6 +57,29 @@ end


 function number_sections(header)
+
+  -- This will only run on second conversion to help fix the links in the TOC
+  -- by retrieving the recorded attributes
+  if not preprocess and #header.content and (header.content[1].t == 'Span') then
+    needs_number_sections = false
+    -- retrieve the attributes
+    local span_content = header.content[1]
+    local old_id = span_content.attributes['data-rmarkdown-temporarily-recorded-id']
+    if old_id then
+      identifiers['#' .. old_id] = '#' .. header.identifier
+    end
+    -- remove span
+    header.content:remove(1)
+    -- prepend span Inlines content to the header content
+    header.content = span_content.content:__concat(header.content)
+    -- return content without the span
+    return header
+  end
+
+  -- The following part will run during the pre-processing step only to
+  --  * add numbers on header
+  --  * record the old identifier for next processing to fix links
+
   -- Unnumbered
   if not needs_number_sections or header.classes:find("unnumbered") then
     return nil
@@ -97,9 +109,13 @@ function number_sections(header)

   --- Return Div wrapping Header and recording Header's current ID
   if preprocess then
-    div = pandoc.Div({header})
-    div.attributes['data-rmarkdown-temporarily-recorded-id'] = header.identifier
-    return div
+    local attr = {}
+    attr['data-rmarkdown-temporarily-recorded-id'] = header.identifier
+    header.content = pandoc.Span(
+      header.content,
+      pandoc.Attr('', '', attr)
+    )
+    return header
   end

   return header
@@ -108,7 +124,6 @@ end

 return {
   {Meta = is_preprocess},
-  {Div = pairwise_link_id},
+  {Header = number_sections},
   {Link = fix_link_id},
-  {Header = number_sections}
 }
````

I would like your thoughts on it ? I know you spent time on this, and this may sounds unnecessary to you but I would really prefer to not use divs and I wonder what you think about this change.
https://github.com/atusy/rmarkdown/compare/preprocess-number-sections...cderv:lua/use-spans-number-sections

**Comment by atusy** [comment]
@cderv Thanks, and yes, using `Span` sounds cleaner.

File: inst/rmarkdown/lua/number-sections.lua [hunk 4]
```diff
@@ -32,37 +36,112 @@ local section_number_table = {0, 0, 0, 0, 0, 0, 0, 0, 0}
...
+  -- to add recorded attributes set by us as Span attribute in first processing.
+  if not preprocess and #header.content and (header.content[1].t == 'Span') then
+      -- Check attributes
+      local span_content = header.content[1]
+      local old_id = span_content.attributes['data-rmarkdown-temporarily-recorded-id']
+      -- If found, this is our Span added in pre processing step
+      if old_id then
+        identifiers['#' .. old_id] = '#' .. header.identifier
+        -- remove span
+        header.content:remove(1)
```
atusy, line 81: @cderv 
I think `remove` is added since 2.9.2 with the following commit:
https://github.com/jgm/pandoc/commit/4ede11a75f37bb35d384fc9ce95231671e18acf3

If you want to support older versions, you need some workaround.

```lua
table_copy = {}
for i, v in ipairs(table_original) do
  if i > 1 then
    table.insert(table_copy, v)
  end
end
```

cderv: I have started here https://gist.github.com/cderv/de33e6171ff181d7cce7e7630a8b9935 but should probably move in Wiki to offer contribution to others.

cderv (approved):
Ok this is now fine. Working with Lua filters in **rmarkdown** is really painful as long as we need to support several versions in the past

I added and rewrote some tests and did some refactoring to make things clearer for future us. 

Going to merge now. I think this should be fine.

cderv:
And thanks @atusy for digging into this in the first place!

PR #2095, opened 2021-04-06
Add css argument to html_document_base

cpsievert:
By moving css -> pandoc logic from `html_document` into `html_document_base`, this PR allows output formats that build upon `html_document_base()`, such as `flexdashboard::flex_dashboard` to simply pass `css` (or `...`) and let the base format handle the Sass -> CSS -> pandoc transformation


yihui (approved):
Looks good to me. Thanks!

R/html_document_base.R
```diff
@@ -65,6 +66,23 @@ html_document_base <- function(theme = NULL,
     if (is_bs_theme(theme)) bslib::bs_global_set(old_theme)
   }
 
+  # additional sass/css
+  for (f in css) {
+    if (is_bs_theme(theme)) {
+      theme <- bslib::bs_add_rules(theme, sass::sass_file(f))
+      next
```
cderv, line 73: Trying to understand how this will be different than the current behavior: 

`sass_file()` will take any CSS and SCSS file to be processed by **bslib** and no more passed to Pandoc is that right ? not just sass file ?

It won't be as simple as before, like changing header to red  

CSS in `style.css`
```css
h1 {
  color: red;
}
```
With standard theme it works ok
````markdown
---
output: 
  html_document:
    self_contained: false
    css: style.css
---

# A title that I want red

Some text
```` 
With bootstrap 4, I need to tweak more because CSS won't apply as before ? 
````markdown
---
output: 
  html_document:
    self_contained: false
    css: style.css
    theme:
      version: 4
---

# A title that I want red

Some text
````

Is that expected ?

File: R/html_document_base.R [hunk 2]
```diff
@@ -65,6 +66,23 @@ html_document_base <- function(theme = NULL,
...
+    }
+    is_sass <- grepl("\\.s[ac]ss$", f)
+    if (is_sass) {
+      f <- sass::sass(
```
cderv, line 77: We missed that in previous PR but should we check for sass availability and throw a nice error early if **sass** is not installed yet ? 
```suggestion
      if (!xfun::loadable("sass")) {
        stop("Using `.sass` or `.scss` file in `css` argument requires the sass package.", call. = FALSE)
      }
      f <- sass::sass(
```

R/html_document_base.R
```diff
@@ -65,6 +66,23 @@ html_document_base <- function(theme = NULL,
...
+        sass::sass_file(f),
+        output = sub("\\.s[ac]ss$", ".css", f),
```
cderv, line 79: Just wondering: should the built CSS be consider as a temp file or like here as a file to keep locally once the document is rendered ?

I am fine with this - just wondering if we should think of making it a temp file.

cderv (changes requested):
The change in the doc will affect any formats including ioslides or slidy. See the documented output Rd file. 

We may need to fix the documentation by non inheriting the params for html_document for formats that do not support that do not support the CSS with bslib and sass. 

* `slidy` has its own css argument that will only be passed to Pandoc
* `ioslide` also
* `html_vignette` shave theme NULL and bslib can't be used 
* `html_notebook` should work as `html_document` so ok

Our docs is not already the clearest. We don't want to get more confusing by mentioning sass or bslib when this is not expected.

R/html_document_base.R
cpsievert: > With bootstrap 4, I need to tweak more because CSS won't apply as before ? Is that expected ?

I'm not sure what you mean. The difference is that, in the 2nd example, the CSS files are essentially appended to the Bootstrap stylesheet (instead of being processed by Pandoc), but you should still get the same end result. If you're worried about the implications for CSS precedence, we could process CSS files with Pandoc instead, but scss/sass files definitely want to be added via `bslib::bs_add_rules(theme, sass::sass_file(f))`

R/html_document_base.R
cpsievert: Yea, I can't remember if I had a good justification for not using a temp file the 1st time -- I'll go ahead and make that change too.

R/html_document_base.R
cderv: I take the opportunity of this PR to test a little how all this works. 

> Your examples give the same end result

I don't get the same end result - h1 title is not red for me in the second example. Note the `self_contained: false`. If I set to TRUE then I get the red title. Not sure why this difference - CSS does not seem to be included correctly in not self contained mode ? Is this expected or a bug ?

> If you're worried about what this difference means for CSS precedence

I am noticing indeed that the precedence is different: 

This won't change the title in red
```yaml
output: 
  html_document:
    css: style.css
    theme:
      bootswatch: cerulean
```

This will 
```yaml
output: 
  html_document:
    css: style.css
    theme: cerulean
```

Is that because the bootstap rules are not overridden by a custom CSS as it was before ?

We have different behavior that I am trying to understand. Maybe it is expected but in this case it creates confusion some stuff that already works with Rmarkdown won't work the same with new bootstrap 4 support. This means documentation load for sure in this case. cc @apreshill

cpsievert: Yikes, `sass::sass_file()` wasn't working the way I thought it did. Switching from `sass::sass_file()` to `xfun::read_utf8()` (in 7515b16f86d9f8180fac8c2dd7f149ef610bc874) should fix all the issues your seeing.

cderv: Yes now all examples are working

R/html_document_base.R
cderv: Ok in fact you may not want to do that (or not this way) because it will create a file in temp directory on the environment. This will lead (for a `self_contained: false` document) to creating this in HTML 
```html
<link rel="stylesheet" href="C:\Users\chris\AppData\Local\Temp\RtmpKqPzuP/rmarkdownd9b35e5ea2507976/rmarkdown.min.css" type="text/css">
```

which I believe is not what we want - or at least it is not working. 

I believe this is because you created the temp file in the R temp directory. When using RStudio, the rendering will happen in a new process and R temp folder will be removed at the end of the rendering. 
That is one reason why we don't use `tempdir()` so much in rmarkdown (like for plot file for figures - they are in an intermediary folder). If a dependency file needs to be created for HTML to use, it should be in the `lib_dir` folder (like html dependencies ?) or another intermediates directory. 

Html dependencies have different treatment depending on self contained or not.

So We need to think of that if we want

* the user to provide a `.scss` file
* a `.min.css` is created for Pandoc to use
* no trace of this file is in the working directory of the user. 

If we are ok for the `.css` to be right next to the `.scss`, then we could revert back to your first change. I am not sure leaving a file of a run on user directory is what we want though. Hence my first comment.

cpsievert: Ahhh, I guess I was assuming `--css` always inlined the contents because that's all `css` did before I made these changes (the `css` parameter has never worked properly with `self_contained: false`)? I guess I can fix that as well if you'd like

R/html_document_base.R
```diff
@@ -120,6 +143,16 @@ html_document_base <- function(theme = NULL,
 
     preserved_chunks <<- extract_preserve_chunks(input_file)
 
+    # At this point lib_dir should exist
+    # (fs will error it copying fails)
```
cpsievert, line 147:
```suggestion
    # (fs will error if copying fails)
```

R/html_document_base.R
cderv: > the css parameter has never worked properly with self_contained: false

usually you have your CSS file locally in a relative path to the Rmd file, and so it will stay relative to the html file by default as the output dir is the same directory. Here it will be created in a temp dir which will always be absolute path. 

However you're right, `self_contained: false` + `render(..., output_dir = "out/")` will lead to a css file not copied to the output dir. 

There is a mechanism in `html_document_base` to find absolute references of resource and replace them with relative ones though, but yes it is not working properly currently in the case described the line above. 🤔

cpsievert: @cderv I believe 65b0c9d generally addresses the issue?

cderv: Yes I saw that. I was taken by another issue this morning. I'll have a look now.

cpsievert: I just noticed these changes introduce real unit test failures -- would you prefer to go back to the old behavior (i.e., just write the css file to disk? Maybe it'd be possible/safe to remove the css file when `self_contained = TRUE`?)

apreshill:
Can confirm- using rmarkdown from this branch does indeed lose the external CSS file in xaringan 😓 

```
output file: index.knit.md

Error: [ENOENT] Failed to copy '/Users/alison/rtalks/talks/ares-kind-tools/assets/css/my-theme.css' to 'libs/assets/css/my-theme.css': no such file or directory
Execution halted
```

R/html_document_base.R
cderv: I tried to handle this in ad3f64f

cderv:
@cpsievert I took the time to look into **sass** and come up with ad3f64f. 
Basically we need to write the generated CSS file with the other libs and other resources - this folder will get deleted when `self_contained = FALSE`. It will not if `self_contained = TRUE`. 
To keep the cache awareness, I rewrote a special version of `sass::output_template()` that allows to change the temp directory. I'll make a PR into sass as I think it could be better to have this directly in the package. 

I believe this changed should make all this works. 

* By default, `.css` is used as Pandoc argument as before
* when `.scss` or `.sass`, `sass::sass()` is used and it creates a temp file cache aware in the lib dir
* When a bslib theme used, everything is done by **bslib** directly - no css argument is passed to pandoc as the rules of the css / scss / sass files are integrated into `bootstrap.min.css`

How does it look to you ? 
Can you check that this would work as you need for **flexdashboard** ? 

@apreshill can you retry your xaringan example ? I don't see why it would fail considering the changes now.

Thank you.

apreshill:
Hi @cderv- Can confirm that installing from this PR *now* works, at least with my xaringan deck!

R/html_document_base.R
```diff
@@ -83,6 +90,24 @@ html_document_base <- function(theme = NULL,
...
+    for (f in css) {
+      is_sass <- grepl("\\.s[ac]ss$", f)
+      if (is_sass) {
+        if (!xfun::loadable("sass")) {
+          stop("Using `.sass` or `.scss` file in `css` argument requires the sass package.", call. = FALSE)
+        }
+        f <- sass::sass(
+          sass::sass_file(f),
+          # we use a custom fonction to write temp file in lib_dir
+          output = sass_output_template(tmpdir = lib_dir),
```
cpsievert, line 103:
```suggestion
          # write output file to `lib_dir/sass-{sass:::sass_hash()}{[basename(f)}`
          output = sass_output_template(
            basename = tools::file_path_sans_ext(basename(f)),
            tmpdir = lib_dir
          ),
```

cpsievert:
Thanks @cderv, your changes look great (just the one minor suggestion)!

> We could also not use a cache at all and just rewrite fresh the css file output of sass::sass()

Calls to `sass::sass()` get pretty expensive pretty quickly, so it's great to provide users the option of caching (you can opt-out with `options(sass.cache = FALSE)` and/or `shiny::devmode()`

> Will there really a again for this cache when used with rmarkdown ?

Not sure I understand the question -- could you clarify?

> To keep the cache awareness, I rewrote a special version of sass::output_template() that allows to change the temp directory.

Instead of adding this to (the already complex) `output_template()`, maybe we should just export `sass:::sass_hash()`. Then rmarkdown's code could just be `output = file.path(lib_dir, sass::sass_hash(), sub("\\.scss$", ".min.css", basename(f)))`

R/html_document_base.R
cderv: Oh yeah thanks for catching that!

cderv:
> Not sure I understand the question -- could you clarify?

I mistype my question sorry. You explained the gain already to it answered it.

> Instead of adding this to (the already complex) output_template(), maybe we should just export sass:::sass_hash(). Then rmarkdown's code could just be output = file.path(lib_dir, sass::sass_hash(), sub("\\.scss$", ".min.css", basename(f)))

Just adding a 4th argument to `output_template()` to just change a default did not seem to make it more complex. In fact it was quite clear to me that `output_template()` was the function to use to benefit caching without really knowing how it works 😅. However if you prefer to export a new function in **sass** and this does the job why not. Not sure exporting a new function is easier than adding an argument on the long run.

Also, it seems `sass_hash()` should take arguments, isn't it ? 
And having a dirname that is not just a key in lib dirs seems better

So for rmarkdown, it would be something like this maybe ?
```r
output = file.path(lib_dir, paste0("sass-",sass::sass_hash(f)), xfun::with_ext(basename(f)))
```

Does not seems simpler to me to redo the logic IMO. 

in **sass** the all cache key is : 
```r
    sass_hash(list(
      sass_input, options, cache_key_extra,
      # Detect if any attachments have changed
      if (is_sass_layer(layer) && !is.null(layer$file_attachments)) get_file_mtimes(layer$file_attachments)
    ))
```
would that not be missing if we call ourself `sass_hash` ?

Just some feedbacks so that you can decide what you want for **sass** and tell us.

cpsievert:
Ok yea, I don't feel that strongly about it. Feel free to send a PR for `sass::output_template()` if you'd like. Regardless, this PR LGTM.

cderv:
Thanks. 

BTW, PR has already been opened https://github.com/rstudio/sass/pull/76

PR #2096, opened 2021-04-07
Fix "Invalid cross-device link" with `render()`

gorgitko:
When `tempdir()` is used for `intermediates_dir`, this error can occur on Linux systems.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 rmarkdown 2.8
 ================================================================================
 
+- Fix an `Invalid cross-device link` error when `tempdir()` is used for `intermediates_dir` in `render()` (thanks, @gorgitko, PR #2096).
+
 - Fix a regression in HTML default template with floating toc incorrectly placed on small size window (thanks, @grimbough, #2071)
 
 - Provided a `runtime: shiny` fix for output formats that pass a modified `bslib::bs_theme()` object to `html_document_base()`'s `theme` (thanks, @cpsievert, #2049).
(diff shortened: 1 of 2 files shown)


cderv (changes requested):
Thanks! 

We have a `move_dir` function in `util.R` that I think could be useful here
https://github.com/rstudio/rmarkdown/blob/145046175c721af185aa6ba3ecc9262f12dc7369/R/util.R#L263-L268

* It will create the destination directory if not already 
* It will try `file.rename` and 
* use `file.copy` + `unlink` if the rename failed

We could use that right ? 

Also can you add a bullet to NEWS about fixing a bug with render when using a temp directory as intermediates. (This is about that right ? Is there an issue to link to ?)

gorgitko:
Yes, we can certainly use `move_dir()`, and so I have made one more commit to my PR.

There is no issue for this PR, so I just reference it as "PR #2096" in NEWS.md :slightly_smiling_face:

NEWS.md
```diff
@@ -1,6 +1,8 @@
 rmarkdown 2.8
 ================================================================================
 
+- Fix an `Invalid cross-device link` error when `tempdir()` is used for `intermediates_dir` in `render()` (thanks, @gorgitko, PR #2096).
```
cderv, line 4:
```suggestion
- Fix an `Invalid cross-device link` error when `tempdir()` is used for `intermediates_dir` in `render()` (thanks, @gorgitko, #2096).
```

cderv (approved):
That is perfect. Thanks a lot! 

I just tweak NEWS - no need to add PR, just the Github Number (issue or PR) is good

gorgitko:
OK, thanks a lot and have a nice day!

PR #2098, opened 2021-04-08
tweaks intermediates

### LINKED ISSUE #2086: Drop the intermediate files *.knit.md and *.utf8.md
In **rmarkdown**, we used to generate the intermediate file `*.knit.md` from `knitr::knit()`, and convert it to another intermediate file `*.utf8.md`. Then pass the utf8 file to Pandoc (because Pandoc only accepts UTF-8 input).

Since [**knitr** 1.24](https://github.com/yihui/knitr/releases/tag/v1.24), `knitr::knit()` only accepts UTF-8 input. Since [**rmarkdown** 2.0](https://github.com/rstudio/rmarkdown/releases/tag/v2.0), functions in **rmarkdown** also only accept UTF-8. It makes little sense to generate these two intermediate files now (e.g. d8fbf4c4641f9c3d3df13f77d7cb34fe01d3daab makes an unnecessary file copy), and I suggest that we stop generating them in a future version.

Additional benefits:

1. **blogdown** users no longer needs to ignore these files: https://github.com/rstudio/blogdown/blob/a6c92734/R/check.R#L51 It may also avoid the unnecessary refreshing of Hugo server.
2. **bookdown** users won't see these intermediate files, either: https://github.com/rstudio/bookdown/issues/986

cderv:
knitr 1.24+ and rmarkdown 2.0+ only accepts UTF-8 so  `*.utf8.md` is not needed anymore

closes #2086

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: rmarkdown
 Type: Package
 Title: Dynamic Documents for R
-Version: 2.7.7
+Version: 2.7.8
 Authors@R: c(
   person("JJ", "Allaire", role = "aut", email = "jj@rstudio.com"),
   person("Yihui", "Xie", role = c("aut", "cre"), email = "xie@yihui.name", comment = c(ORCID = "0000-0003-0645-5666")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,10 @@
 rmarkdown 2.8
 ================================================================================
 
+- A new internal option `rmarkdown.knit.ext` has been added to control the extension of the intermediary knit output during a rendering. It defaults to `md` to produce `*.knit.md`. Only useful for very advanced usage (#2098).
+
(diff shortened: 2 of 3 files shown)


yihui (approved):
Looks good. Thanks! Let's see if we would discover any problems when checking the reverse dependencies the next time we make the CRAN release.

cderv:
I added the simplest for now: an option `rmarkdown.knit.ext` to customize `*.knit.<ext>`

This would allow to generate `.knit.md~` in **blogdown**, which should be enough. 

Do you think it should be better to customize the all extension: 
* default: `*.knit.md`
* customized : `*.<ext>`


When ready to merge I'll add NEWS bullet

yihui (approved):
The current method you provided is enough (it doesn't matter if it is `*.knit.md~` or `*.md~`). No need to do further work. Thanks!

PR #2101, opened 2021-04-09
Improve Pandoc versions handling in Lua Filters

cderv:
fixes #2099 and will close #2100 

This modifies the CI to check one more pandoc version. it also now test 2.11.4 (last RStudio 1.4.1106 shipped version) and organize the testing matrix. We test 2.7.3, 2.5 and 2.0.0.1 for now.

PR #2102, opened 2021-04-09
[LUA] Add a shared.lua for sharing common functions

cderv:
Following comment on https://github.com/rstudio/rmarkdown/pull/2101#discussion_r610573288

This should be better and aligned with the function pagedown. https://github.com/rstudio/pagedown/blob/66f760efc824ae17d97bc0898c5a7e3322fc8738/inst/resources/lua/uri-to-fn.lua#L5

@atusy is this better to you ? 

I just forgot to add this greater than check 🤦


cderv:
This would also close #2103 following @tarleb idea.

cderv:
@yihui Following ideas in https://github.com/rstudio/rmarkdown/issues/2103#issuecomment-816709371 I added the env var setting in `pandoc_lua_filter_args()` where we decide to include or not Lua filters (using Pandoc 2+ or not)

This means that anyone using this function could use in theory the `shared.lua` file in its filter. 

It could also be done in `pkg_file_lua()` or any other place. 

This one seems a good one but does that seem ok to you ?

R/pandoc.R
```diff
@@ -749,7 +749,12 @@ pandoc_citeproc <- function() {
 #' @export
 pandoc_lua_filter_args <- function(lua_files) {
   # Lua filters was introduced in pandoc 2.0
-  if (pandoc2.0()) c(rbind("--lua-filter", pandoc_path_arg(lua_files)))
+  if (pandoc2.0()) {
+    if (!nzchar(Sys.getenv("RMARKDOWN_LUA_SHARED"))) {
```
yihui, line 753: I often test the emptiness of env vars by this:

```suggestion
    if (is.na(Sys.getenv("RMARKDOWN_LUA_SHARED", NA))) {
```

It's slightly stricter, and also more readable to me ([I often have to think twice](https://yihui.org/en/2018/08/double-negatives/) what `!nzchar()` means, i.e., translate "is not non-zero chars" to "is empty string").

R/pandoc.R
```diff
@@ -749,7 +749,12 @@ pandoc_citeproc <- function() {
...
+      Sys.setenv(RMARKDOWN_LUA_SHARED = pkg_file_lua("shared.lua"))
```
yihui, line 754: When setting options or env vars, usually we need to unset them on exit.

```suggestion
      vars <- xfun::set_envvar(c(RMARKDOWN_LUA_SHARED = pkg_file_lua("shared.lua")))
      on.exit(xfun::set_envvar(vars), add = TRUE)
```

But this needs to be in another higher-level function such as `render()` or `pandoc_convert()`.

yihui (changes requested):
Sounds like the right direction to go. Thanks!

R/pandoc.R
cderv: You have a blog post for everything. I'll remember that. This is indeed clearer!

R/pandoc.R
cderv: 🤔 I see. 

Does this mean it is better to set this env var elsewhere ? 

I find `pandoc_convert()` to low level because you can use it without any **rmarkdown** specific feature, especially Lua filters. 

Maybe I should set that in `render()` then. 

Otherwise, what about the `on_exit=` argument of our `output_format` when our Lua filters are used ?

cderv: I am still thinking about the best way to do it. 
I experimented with a few places and I am not yet satisfied. `pandoc_lua_filter_args` maybe not be the right place as it is not tied to the filters that requires this file.

cderv: I decide that `render()` may be the right place as it concerns setting environment during a rendering of a document, and because this is where lua argument is passed to Pandoc. 

This allows for setting the on.exit() more easily too.

R/render.R
```diff
@@ -882,6 +882,10 @@ render <- function(input,
 
       # if Lua filters are provided, add the command line switch
       if (!is.null(lua_filters <- output_format$pandoc$lua_filters)) {
+        if (is.na(Sys.getenv("RMARKDOWN_LUA_SHARED", unset = NA))) {
+          lua_env_vars <- xfun::set_envvar(c(RMARKDOWN_LUA_SHARED = pkg_file_lua("shared.lua")))
+          on.exit(xfun::set_envvar(lua_env_vars), add = TRUE)
+        }
```
cderv, line 888: Maybe it is not anymore needed to check that this variable has been set. 
Can we decide that it is reserved for rmarkdown used ? 

The only benefit to keep that this way would be that we could set this environment variable to another value inside another package and `render()` would not change it but this would make our filter fail ?

I am still hesitant to do that in `render()` because Lua filters are mainly tied to an output format directly and this env var is required only our filter. There is no harm to make it available to every filter though - it is more lua helpers available. 

So as this is mandatory for our filter to work, it would not be good if this variable would be set to another value.

```suggestion
        lua_env_vars <- xfun::set_envvar(c(RMARKDOWN_LUA_SHARED = pkg_file_lua("shared.lua")))
        on.exit(xfun::set_envvar(lua_env_vars), add = TRUE)
```

This is why I am taking time on this...

cderv: We could also imagine something more generic like
```suggestion
        shared <- attr(lua_filters, "shared")
        if (!is.null(shared)) {
          lua_env_vars <- xfun::set_envvar(c(RMARKDOWN_LUA_SHARED = shared))
          on.exit(xfun::set_envvar(lua_env_vars), add = TRUE)
        }
```
We could add the shared lua filter as an attributes when passing the Lua filters in `lua_filters=` argument in `output_formats`. 
But this would mean make sure to keep the attributes in `merge_pandoc_options()` - what to do if both base and overlay formats defines lua filters with an attributes in that case. 🤔

cderv:
Setting in `render()` solution has the benefit of just setting this once inside **rmarkdown** once and for all. 
So it seems the easiest and maybe the simplest for our purpose.

Just sharing my test to set it in an output format - this is just for `html_document_base` and this would need to be set in each format using the filters

````diff
diff --git a/R/html_document_base.R b/R/html_document_base.R
index 2009c166..619a5888 100644
--- a/R/html_document_base.R
+++ b/R/html_document_base.R
@@ -51,6 +51,10 @@ html_document_base <- function(theme = NULL,

   theme <- resolve_theme(theme)

+  # Lua filters
+  lua_filters <- pkg_file_lua(c("pagebreak.lua", "latex-div.lua"))
+  lua_env_vars <- set_shared_lua_var()
+
   # In the bs_theme() case, we set the theme globally so that knitting code may
   # alter it before we ultimately compile it into an HTML dependency.
   old_theme <- NULL
@@ -63,6 +67,8 @@ html_document_base <- function(theme = NULL,
   on_exit <- function() {
     # In this case, we know we've altered global state, so restore the old theme
     if (is_bs_theme(theme)) bslib::bs_global_set(old_theme)
+    # reset modified environment variables
+    xfun::set_envvar(lua_env_vars)
   }

   # pre_processor
@@ -188,7 +194,7 @@ html_document_base <- function(theme = NULL,
     knitr = NULL,
     pandoc = pandoc_options(
       to = "html", from = NULL, args = args,
-      lua_filters = pkg_file_lua(c("pagebreak.lua", "latex-div.lua"))
+      lua_filters = lua_filters
     ),
     keep_md = FALSE,
     clean_supporting = FALSE,
diff --git a/R/render.R b/R/render.R
index 9bc50c9b..a6896848 100644
--- a/R/render.R
+++ b/R/render.R
@@ -882,10 +882,6 @@ render <- function(input,

       # if Lua filters are provided, add the command line switch
       if (!is.null(lua_filters <- output_format$pandoc$lua_filters)) {
-        if (is.na(Sys.getenv("RMARKDOWN_LUA_SHARED", unset = NA))) {
-          lua_env_vars <- xfun::set_envvar(c(RMARKDOWN_LUA_SHARED = pkg_file_lua("shared.lua")))
-          on.exit(xfun::set_envvar(lua_env_vars), add = TRUE)
-        }
         lua_filters <- pandoc_lua_filter_args(lua_filters)
       }
diff --git a/R/util.R b/R/util.R
index 5e759b33..77b94de9 100644
--- a/R/util.R
+++ b/R/util.R
@@ -89,6 +89,10 @@ pkg_file_lua <- function(filters = NULL, package = "rmarkdown") {
   pandoc_path_arg(files)
 }

+set_shared_lua_var <- function(shared = pkg_file_lua("shared.lua")) {
+  xfun::set_envvar(c(RMARKDOWN_LUA_SHARED = shared))
+}
+
 #' @rdname rmarkdown_format
 #' @export
 from_rmarkdown <- function(implicit_figures = TRUE, extensions = NULL) {
````

So afterall, setting once in render without possibility to customize the script is maybe the simpler - this would allow any filters in **rmarkdown** to do `dofile(os.getenv 'RMARKDOWN_LUA_SHARED')`, and also any filter in any other custom formats.

R/render.R
yihui: > Maybe it is not anymore needed to check that this variable has been set.
> Can we decide that it is reserved for rmarkdown used ?

Yes we can.

yihui:
> Setting in `render()` solution has the benefit of just setting this once inside **rmarkdown** once and for all.
> So it seems the easiest and maybe the simplest for our purpose.

I'm okay with that.

cderv:
Then we'll go with that. 

* `RMARKDOWN_LUA_SHARED` is set by `render()` to the absolute path of `shared.lua`
* This is used in Lua script by calling `dofile(os.getenv 'RMARKDOWN_LUA_SHARED')`
* We do that mainly to support Pandoc versions previous 2.1.2 as `PANDOC_SCRIPT_FILE` is not available to load relatively to a script. 
* This mechanism is for internal use only (at least for now) even if this shared script will be available to any filter used with `render()` as soon as the `dofile()` command is used.

Thanks for the review.

PR #2108, opened 2021-04-14
Add an article about accessibility features

cderv:
* Take write up from the next blog post https://github.com/rstudio/blog/pull/436 and make it an article for the website
* Add some links
* Plan is to add more content later in this post

Note: I used the visual editor to format the Rmd post.

Issue: I don't know why yet but the toc is off with pkgdown... 🤔

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,4 @@
 inst/doc
 doc
 Meta
-docs
+reference/
--- a/vignettes/articles/accessibility.Rmd
+++ b/vignettes/articles/accessibility.Rmd
@@ -1,0 +1,134 @@
+---
+title: "About accessibility for HTML document"
+---
+
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>"
+)
+```
+
+When producing any content for the Web, it is really important to consider **Accessibility.** W3C, the [World Wide Web Consortium](https://en.wikipedia.org/wiki/World_Wide_Web_Consortium "Wikipedia W3C page"), has some standard for Web Design and Application regarding Accessibility
+
+> The power of the Web is in its universality.\
+> Access by everyone regardless of disability is an essential aspect.\
+> --- Tim Berners-Lee, W3C Director and inventor of the World Wide Web
+
+**rmarkdown** has to make things easier to produce such content and this document aims to present the existing features.
+
+## Adding alternative text for figures
+
+Until **knitr** 1.31, [alternate text](https://www.w3schools.com/tags/att_img_alt.asp) was added to figures created in chunks by using the `fig.cap` chunk options. There was no way for knitr users to create alt text and figure captions separately. This missing [feature](https://github.com/rstudio/rmarkdown/issues/1867) was originally requested by [Dr. Mine Dogucu](https://mdogucu.ics.uci.edu/).
+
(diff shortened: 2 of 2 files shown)


yihui (commented):
I feel this article should go to the R Markdown Cookbook---perhaps a section after this https://bookdown.org/yihui/rmarkdown-cookbook/html-accessibility.html (or add a subsection in it). I tend not to add more vignettes or articles into the package, but keep all documentation in the Definitive Guide or Cookbook. Same thing applies to bookdown and blogdown, since they all have companion books (unlike other packages). We can always link to a specific section in the books and promote it when necessary.

What do you think?

cderv:
This is a good question. I asked myself the same thing because I was worried of duplication and several places for documentations. So I understand your comment. 

> I feel this article should go to the R Markdown Cookbook

The R Markdown Cookbook could certainly be a good place for this as there is already the chapters on accessibility. 

> I tend not to add more vignettes or articles into the package, but keep all documentation in the Definitive Guide or Cookbook. Same thing applies to bookdown and blogdown, since they all have companion books (unlike other packages).

However, I feel package websites are a great way to document some specific topics and could be easier for "discoverability" than in the middle of a book. Unfortunately, we don't have a lot of vignette and so documentation inside package is sparsed. Even if vignettes could be a constraint (bc CRAN), 'articles' feature from **pkgdown** is nice to avoid that and still share documentations on websites.

So we should really discuss that and decide where new topics should be added. Current other ideas are: 
* Customization of HTML documents using CSS and SCSS/SASS
* Bootstrap 4 supports though **bslib** (somthing like flexdahsboard : https://flexdashboard-pkg.netlify.app/articles/theme.html#using-bootstrap-1)

All the other packages have useful documentations in their website - I feel we should have some too because it is expected by users and easily findable. Other packages don't have book though and we should also take advantage of that. Now that the books are promoted in front page, there is no way to miss it.

For this specific articles, I agree that a subsection in https://bookdown.org/yihui/rmarkdown-cookbook/html-accessibility.html maybe a better place for it. All the more because it shows how to use `fig.alt` which is in **knitr** really and this is not directly related to **rmarkdown** (except the `fig_caption: FALSE)`

yihui:
I need to go back to the pkgdown site discussion from a few months ago and think more about it. Basically, we have already got three resources for the documentation: the R help (Rd), the official website rmarkdown.rstudio.com, and the two 350-page books. I feel these were enough, and don't really want to invest too much into a fourth place for documentation (vignettes/articles). If there is anything new to add, I'd prefer adding to the existing resources. (BTW, the Lua filters vignette should perhaps go to Part IV of the Definitive Guide.)

cderv:
Yes we need to think this through - I agree.

> BTW, the Lua filters vignette should perhaps go to Part IV of the Definitive Guide.

Yes I think so too. We mention them in the R Markdown Cookbook already too now : https://bookdown.org/yihui/rmarkdown-cookbook/lua-filters.html 
This was after the vignette creation - at first it was mainly as a technical documentation about Lua filters but it should now get part of the book as there is interesting feature tied to it.

I need to make an habit of updating the books as new feature goes in.

apreshill:
Let's remove articles then? Distill does not have any either: https://pkgs.rstudio.com/distill/

yihui:
That sounds good to me (after we move the Lua filters article into a book chapter or section).

apreshill:
It may be worth transitioning those book websites over to `bs4_book` so that online search works better, if that is the sole source of documentation.

PR #2137, opened 2021-05-12
Use a predictable file name within a temp folder

schloerke:
## Issue

When rendering `runtime: shiny` documents through `shinytest`, any plot figures are put into a `*_files` folder that is random every time a test is run.

This is problematic as the `html` field of the dynamic UI produced in https://github.com/rstudio/rmarkdown/blob/c88635f10ae672115baf6697bc235ce3c41012f0/R/shiny.R#L304-L318

If a plot is produced, `value$html` will contain something similar to 
```html
<p><img src="file847d61950236_files/figure-html/unnamed-chunk-12-1.png" width="960" data-figure-id=fig1 /></p>
```

The problem comes from the random name of `file847d61950236_files`, as it is random every time the rmarkdown document is rendered.
https://github.com/rstudio/rmarkdown/blob/c88635f10ae672115baf6697bc235ce3c41012f0/R/shiny.R#L452-L454

`tempfile()`s do not listen to the `set.seed()` value.

```r
> set.seed(1234); tempfile()
#> [1] "/var/folders/0k/bxg5lhr92sq74mb1d446ql540000gp/T//RtmpTi0WqW/filee5b3219c1a8f"
> set.seed(1234); tempfile()
#> [1] "/var/folders/0k/bxg5lhr92sq74mb1d446ql540000gp/T//RtmpTi0WqW/filee5b3f93b386"
```

Given the random output file name, the `*_files` directory is created from it.
https://github.com/rstudio/rmarkdown/blob/c88635f10ae672115baf6697bc235ce3c41012f0/R/shiny.R#L240-L241

-------------------------

# Fix 

If rmarkdown rendered inside a temp folder within the R session temp folder, then names could be predictable and would not need to worry about clashing.

Ex: Looking at the same `` tag as above when running `rmarkdown::run("index.Rmd")` :
```html
<p><img src="index_files/figure-html/unnamed-chunk-12-1.png" width="960" data-figure-id=fig1></p>
```

`index_files` is a consistent name, but because there was another temp folder used, then there is no clashing.

--------------------

Cautions:

* The temp folder is not deleted until the R session is stopped. (A slow memory leak.)

### DIFF
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -450,8 +450,10 @@ rmd_cached_output <- function(input) {
     }
   } else {
     # It's not cacheable, and should be rendered to a session-specific temporary
-    # file
-    output_dest <- tempfile(fileext = ".html")
+    # directory, but with a predictable file name.
+    tmp_dir <- tempfile()
+    output_dest_name <- paste0(xfun::sans_ext(basename(input)), ".html")
+    output_dest <- file.path(tmp_dir, output_dest_name)
   }
   list(
     cacheable = cacheable,
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -452,7 +452,7 @@ rmd_cached_output <- function(input) {
     # It's not cacheable, and should be rendered to a session-specific temporary
     # directory, but with a predictable file name.
     tmp_dir <- tempfile()
-    output_dest_name <- paste0(xfun::sans_ext(basename(input)), ".html")
+    output_dest_name <- xfun::with_ext(basename(input), ".html")
     output_dest <- file.path(tmp_dir, output_dest_name)
   }
   list(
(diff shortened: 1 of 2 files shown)


R/shiny.R
```diff
@@ -450,8 +450,10 @@ rmd_cached_output <- function(input) {
     }
   } else {
     # It's not cacheable, and should be rendered to a session-specific temporary
-    # file
-    output_dest <- tempfile(fileext = ".html")
+    # directory, but with a predictable file name.
+    tmp_dir <- tempfile()
+    output_dest_name <- paste0(xfun::sans_ext(basename(input)), ".html")
```
yihui, line 455: You can use `with_ext()` to change the extension:

```suggestion
    output_dest_name <- xfun::with_ext(basename(input), ".html")
```

yihui (approved):
Thanks!

@cderv Feel free to merge if it looks good to you, too.

cderv (approved):
Yes it looks good to me too. 

I don't know a lot about **shinytest** so if it fixed the issue, then it is good. Thanks @schloerke !

PR #2150, opened 2021-05-28
Loop correctly on splitted file by file_scope function

### LINKED ISSUE #2149: Lua filters broken since v2.8
A bug got introduced in v2.8:
https://github.com/rstudio/rmarkdown/blob/e340d757943cc9ec956aac41e61b0e42ffb64f58/R/render.R#L903

As `input` is a named list, `dirname(input)` fails.

cderv:
This will fix #2149 (at least the error mentioned but I don't see the relation to Lua filters 🤷 )

It resolve a name conflict after change in 92b437a571a0b2f5574b465c3d121eefa6c39de4 (#2098) that caused an issue with file_scope for loop. 

I took the time to add some test so that we can detect this next time (if any). 
To do a proper unit test (and not a big rendering test), I refactored by moving out the file scope logic into a function. Also, I used snapshot tests on the dummy example as it is simple and works well in this case.

Hope this is fine.

### DIFF
--- a/R/render.R
+++ b/R/render.R
@@ -891,7 +891,7 @@ render <- function(input,
       # in case the output format turns on the --file-scope flag, run its
       # file_scope function to split the input into multiple files
       input_files <- input
-      if (!is.null(output_format$file_scope) && is.function(output_format$file_scope)) {
+      if (is.function(output_format$file_scope)) {
         input_files <- file_scope_split(input, output_format$file_scope)
         # ignore if input_files has not really been splitted
         if (length(input_files) > 1) {
@@ -1186,8 +1186,8 @@ resolve_df_print <- function(df_print) {
 #' @export
 output_metadata = knitr:::new_defaults()
 
-file_scope_split <- function(input, file_scope_fun) {
-  inputs <- file_scope_fun(input)
+file_scope_split <- function(input, fun) {
+  inputs <- fun(input)
 
   # file_scope_fun should split the input file in several
   # do nothing if not and return input file unsplited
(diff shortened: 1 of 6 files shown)


R/render.R
```diff
@@ -891,22 +891,15 @@ render <- function(input,
...
-        input_files <- unlist(lapply(inputs, function(input) {
-          file <- file_with_meta_ext(input$name, "split", "md")
-          file <- file.path(dirname(input), file)
-          write_utf8(input$content, file)
-          file
-        }))
-
-        # cleanup the split files after render
-        on.exit(unlink(input_files), add = TRUE)
+      if (!is.null(output_format$file_scope) && is.function(output_format$file_scope)) {
```
yihui, line 894: Since `is.function(NULL)` returns `FALSE`, the test `is.null()` should be unnecessary:

```suggestion
      if (is.function(output_format$file_scope)) {
```

R/render.R
```diff
@@ -1192,3 +1185,21 @@ resolve_df_print <- function(df_print) {
 #' @keywords NULL
 #' @export
 output_metadata = knitr:::new_defaults()
+
+file_scope_split <- function(input, file_scope_fun) {
+  inputs <- file_scope_fun(input)
```
yihui, line 1190: Cosmetic (I feel it is cleaner not to repeat the prefix `file_scope_` in the argument):

```suggestion
file_scope_split <- function(input, fun) {
  inputs <- fun(input)
```

yihui (approved):
> To do a proper unit test (and not a big rendering test), I refactored by moving out the file scope logic into a function.

That's great! To be honest, `render()` is a too big function and I'm all for splitting it into smaller ones, which can be good for both testing and avoiding inadvertent name conflicts.

R/render.R
cderv: Oh right! Nicely seen !

R/render.R
cderv: I agree. Thanks!

PR #2151, opened 2021-06-01
Support 'href' dependencies 

cderv:
This PR is a duplicate of #1948 where I can't push. It also superseed #1805. 

This will fix #2021, fix #2147, fix #794 and fix part of #1842

This sum up the work of @matthewstrasiotto and @crazycapivara with addition on my side based on review element. 

Tests were also added to increase the coverage and be sure we did not miss something. 

With this PR, it is now possible to use HTML dependency that have only a `href` in their `src`

<details>
<summary>Like example in #2147</summary>

````markdown
---
title: "Untitled"
output: html_document
---

```{r setup, include=FALSE}
knitr::opts_chunk$set(echo = TRUE,error = T)
```

## R Markdown

```{r}
library(htmlwidgets)
blink = function(text, interval = 1) {
  htmlwidgets::createWidget(
    'blink', 
    list(text = text, interval = interval),
    dependencies = htmltools::htmlDependency(
      'blink', '0.1', src = c(href = ''), head = '
<script>
HTMLWidgets.widget({
  name: "blink",
  type: "output",
  factory: function(el, width, height) {
    return {
      renderValue: function(x) {
        setInterval(function() {
          el.innerText = el.innerText == "" ? x.text : "";
        }, x.interval * 1000);
      },
      resize: function(width, height) {}
    };
  }
});
</script>'
    ),
  )
}

blink('Hello htmlwidgets!', .5)
```
````

</details>

<details>
<summary>Or like in #1842</summary>

````markdown
---
output: html_document
---

```{r echo=FALSE}
library(htmltools)
```

```{r}
tagList(htmlDependency(
  name = "google-font-exo-2",
  version = "1.0.0",
  src = c(href = "https://fonts.googleapis.com/"),
  stylesheet = "css2?family=Exo+2:ital,wght@0,100;0,200;0,300;0,400;0,500;0,600;0,700;0,800;0,900;1,100;1,200;1,300;1,400;1,500;1,600;1,700;1,800;1,900&display=swap"
))
tags$style("body { font-family: 'Exo 2'; }")
```

Some content

````
</details>

PR #2152, opened 2021-06-01
use `stop2()` from knitr internal instead of `stop(..., call. = FALSE)`

cderv:
This is from https://github.com/rstudio/rmarkdown/pull/2151#discussion_r643156102

I open this PR only to review that I did correctly modified the calls in each files and run the test (in case I missed something)


yihui (approved):
Looks good!

PR #2160, opened 2021-06-04
Restore previous shiny theme when run() exits

### LINKED ISSUE #2159: run() leaks shiny's theme state
`run()`-ing a `runtime: shiny` document with a {bslib} `theme` currently leaks `shiny::getCurrentTheme()` state. This means, you'll see behavior like:  

```r
shiny::getCurrentTheme()
#> NULL
rmarkdown::run("Untitled.Rmd")
class(shiny::getCurrentTheme())
#> [1] "bs_version_4" "bs_theme"     "sass_bundle" 
```

When `Untitled.Rmd` has something like

````
---
title: "Untitled"
runtime: shiny
output: 
  html_document:
    theme:
      version: 4
      bg: black
      fg: white
---

```{r}
selectInput("state", "states", state.abb)
```
````

### PR DESCRIPTION
**Pull request by cpsievert**
Closes #2159

### DIFF
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -141,6 +141,8 @@ run <- function(file = "index.Rmd", dir = dirname(file), default_file = NULL,
   # the theme wins out
   if (length(target_file)) {
     format <- output_format_from_yaml_front_matter(read_utf8(target_file))
+    old_theme <- shiny::getCurrentTheme()
+    on.exit(set_current_theme(old_theme), add = TRUE)
     set_current_theme(resolve_theme(format$options$theme))
   }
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: R/shiny.R [hunk 1]
```diff
@@ -141,6 +141,8 @@ run <- function(file = "index.Rmd", dir = dirname(file), default_file = NULL,
   # the theme wins out
   if (length(target_file)) {
     format <- output_format_from_yaml_front_matter(read_utf8(target_file))
+    old_theme <- shiny::getCurrentTheme()
+    on.exit(set_current_theme(old_theme), add = TRUE)
     set_current_theme(resolve_theme(format$options$theme))
```
cpsievert, line 146: Pretty sure when I originally wrote the `set_current_theme()` code I was thinking that since `set_current_theme()` uses `shiny::shinyOptions()`, the previous value would be automatically restored, but it actually doesn't because this call happens outside of an app/session scope (in that case, `shiny::shinyOptions()` is basically like setting `options()`).

cderv (approved):
Looks good to me. It makes sense to reset this. 

Can you add a NEWS bullet though before we merge ?

cderv:
Thanks !

PR #2164, opened 2021-06-07
do not normalize web paths in css argument

### LINKED ISSUE #2163: Using url in CSS argument throws an error now
This is a regression we have 

Reproducible example : 

````r
dir.create(tmp_dir <- tempfile())
owd <- setwd(tmp_dir)
xfun::write_utf8(c(
  "---", 
  "title: test",
  "---", 
  "",
  "# test"), "test.Rmd")

rmarkdown::render("test.Rmd", 
                  rmarkdown::html_document(
                    css = "http://netdna.bootstrapcdn.com/font-awesome/4.0.3/css/font-awesome.css"
                  )
)

setwd(owd)
unlink(tmp_dir, TRUE)
````

````
Could not fetch http:/netdna.bootstrapcdn.com/font-awesome/4.0.3/css/font-awesome.css
InvalidUrlException "http:/netdna.bootstrapcdn.com/font-awesome/4.0.3/css/font-awesome.css" "URL must be absolute"
Error: pandoc document conversion failed with error 61
Run `rlang::last_error()` to see where the error occurred.
````

It must come from the new sass support in css arg I believe. 

This was first reported in https://github.com/yihui/knitr/issues/1864#issuecomment-855301382

### PR DESCRIPTION
**Pull request by cderv**
This fix #2163 

It was a regression after #2095 and we did not had test for R Markdown document using a url in `css` argument it seems. 

I took the opportunity to add some tests for this, including with **bslib** theme activated.

### CODE REVIEW

File: R/html_document_base.R [hunk 1]
```diff
@@ -106,7 +106,8 @@ html_document_base <- function(theme = NULL,
           options = sass::sass_options(output_style = "compressed")
         )
       }
-      f <- normalized_relative_to(output_dir, f)
+      # do not normalize web path
+      if (!xfun::is_web_path(f)) f <- normalized_relative_to(output_dir, f)
```
yihui, line 110: I'm just wonder if it would make more sense to test the existence of the file, but I guess it would be a more risky fix (i.e., there might be cases in which the file doesn't exist given the current working directory---in particular, I'm not sure about Shiny documents):

```suggestion
      if (file.exists(f)) f <- normalized_relative_to(output_dir, f)
```

yihui (approved):
I think it's okay to merge this PR. Just shared one thought, but feel free to discard it. Thanks!

R/html_document_base.R
cderv: I don't know about this either. This would require some testing to be sure. 
This transformation happens in Pandoc pre processor so right before Pandoc is ran so it should exists. 
Unless there is an edge case we missed (like order of pre processor ?) :think

yihui: Okay. Let's just be conservative here and merge the PR. Thanks!

PR #2190, opened 2021-07-20
use yaml_metadata_block for preserve_yaml feature when possible

### LINKED ISSUE #2118: Pandoc 2.13 adds YAML block support for commonmark ( and gfm)
This was first reported in https://github.com/tidyverse/reprex/issues/375 and needs to be handled in **rmarkdown** directly. 

Pandoc 2.13 added support for `yaml_metadata_block` to `commonmark` in addition to `markdown`: https://github.com/jgm/pandoc/commit/c389211e2f678d9327cd2c008d54a1e438f07a07

This impact `gfm` and `commonmark_x` for which it is activated by default. The change for `gfm` is a breaking change as you need to use `gfm-yaml_metadata_block` to get the previous behavior.

Basic `+yaml_metadata_block` is by default, and if `--standalone` is passed in command line, the YAML block will be added in the output `.md` document. Which was not the case before this new extension. 

We allow that only for `markdown` format I think when `preserve_yaml = TRUE` only. So it should probably work the same for those format too. 
I think we need to update `md_document()` and `github_document()` to take that into account regarding `preserve_yaml = FALSE`. 

Note that `github_document()` is not affected currently I think because we don't use the default Pandoc template for `commonmark` and we don't have `titleblock` variable in ours. 
Default template is
```
❯ pandoc --print-default-template=gfm
$if(titleblock)$
$titleblock$

$endif$
$for(header-includes)$
$header-includes$

$endfor$
$for(include-before)$
$include-before$

$endfor$
$if(toc)$
$table-of-contents$

$endif$
$body$
$for(include-after)$

$include-after$
$endfor$
```

cderv:
This will fix #2118 and modify a bit how `md_document()` is working internally

* We use a standalone (`--standalone`) document if `preserve_yaml = TRUE`
* We try to add / remove the extension `yaml_metadata_block` depending on the value of `variant` and `preserve_yaml`. This is because some formats have the extension set or unset by default, and also depending of the pandoc version. 
* We add some mechanism to handle parsing and setting those extensions. They could be externalize at some point to be used in other functions. 

Doing this should handle correctly most of the case. Let's add that `--standalone` flag will produce this header, and that some format will have special header if not yaml: 
* `markdown` will have a [`pandoc_title_block`](https://pandoc.org/MANUAL.html#extension-pandoc_title_block) if not a YAML block
* `markdown_mmd` will have `mmd_title_block` if YAML block is unset. 

For now, this is the behavior we apply for all variants: 
* `md_document(preserve_yaml = TRUE)` will always produce a YAML block. 
* `md_document(preserve_yaml = FALSE)` will always produce no YAML block.

And this is currently not working which is the main issue of #2118

Should we do something simpler or does it seems like a right thing to do ?

### DIFF
--- a/R/md_document.R
+++ b/R/md_document.R
@@ -118,7 +118,7 @@ adapt_md_variant <- function(variant, preserve_yaml) {
       } else {
         variant_extensions
       }
-    }
+    },
     markdown =,
     markdown_phpextra =,
     markdown_github =,
--- a/R/md_document.R
+++ b/R/md_document.R
@@ -50,7 +50,7 @@ md_document <- function(variant = "markdown_strict",
 
 
   # base pandoc options for all markdown output
-  args <- c("--standalone")
+  args <- c(if (preserve_yaml) "--standalone")
 
   # table of contents
   args <- c(args, pandoc_toc_args(toc, toc_depth))
--- a/R/md_document.R
+++ b/R/md_document.R
@@ -65,7 +65,7 @@ md_document <- function(variant = "markdown_strict",
   variant <- adapt_md_variant(variant, preserve_yaml)
 
   # add post_processor for yaml preservation if not supported by pandoc
-  post_processor <- if (preserve_yaml && grepl('yaml_metadata_block', variant, fixed = TRUE)) {
+  post_processor <- if (preserve_yaml && !grepl('yaml_metadata_block', variant, fixed = TRUE)) {
     function(metadata, input_file, output_file, clean, verbose) {
       input_lines <- read_utf8(input_file)
       partitioned <- partition_yaml_front_matter(input_lines)
@@ -103,17 +103,28 @@ adapt_md_variant <- function(variant, preserve_yaml) {
     paste0(format, ext, collapse = "")
   }
 
+  add_yaml_block_ext <- function(extensions, preserve_yaml) {
+    set_extension(variant_extensions, "yaml_metadata_block", preserve_yaml)
+  }
+
   # yaml_metadata_block extension
   variant_extensions <- switch(
     variant_base,
-    gfm=,
-    commonmark=,
-    commonmark_x= if (pandoc_available(2.13)) set_extension(variant_extensions, "yaml_metadata_block", preserve_yaml),
+    gfm =,
+    commonmark =,
+    commonmark_x = {
+      if (pandoc_available(2.13)) {
+        add_yaml_block_ext(variant_extensions, preserve_yaml)
+      } else {
+        variant_extensions
+      }
+    }
     markdown =,
-    markdown_phpextra=,
-    markdown_github=,
-    markdown_mmd=,
-    markdown_strict= set_extension(variant_extensions, "yaml_metadata_block", preserve_yaml),
+    markdown_phpextra =,
+    markdown_github =,
+    markdown_mmd =,
+    markdown_strict = add_yaml_block_ext(variant_extensions, preserve_yaml),
+    # do not modified for unknown (yet) md variant
     variant_extensions
   )
 
--- a/R/md_document.R
+++ b/R/md_document.R
@@ -48,8 +48,9 @@ md_document <- function(variant = "markdown_strict",
                         pandoc_args = NULL,
                         ext = ".md") {
 
+
   # base pandoc options for all markdown output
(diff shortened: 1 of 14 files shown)


R/md_document.R
```diff
@@ -88,3 +92,41 @@ md_document <- function(variant = "markdown_strict",
...
+      }
+    },
+    markdown =,
+    markdown_phpextra =,
+    markdown_github =,
+    markdown_mmd =,
+    markdown_strict = add_yaml_block_ext(variant_extensions, preserve_yaml),
+    # do not modified for unknown (yet) md variant
+    variant_extensions
+  )
```
yihui, line 129: Below is how I would write the code (replacing `switch` with `if` to avoid the repetition of `add_yaml_block_ext(variant_extensions, preserve_yaml)`, and listing two conditions instead of enumerating all possible formats that could be left to be considered in the future):

```suggestion
  set_extension <- function(format, add = TRUE, ext = "yaml_metadata_block") {
    ext <- paste0(ifelse(add, "+", "-"), ext)
    if (grepl(ext, format, fixed = TRUE)) return(format)
    paste0(format, ext, collapse = "")
  }

  if ((variant_base == 'markdown_strict') ||
      (variant_base == 'commonmark_x' && pandoc_available(2.13))) {
    variant_extensions = set_extension(variant_extensions, preserve_yaml)
  }
```

It's just a suggestion. If you prefer `switch`, that's okay to me.

yihui (approved):
Looks good. Thanks!

R/md_document.R
cderv: Let me clarify something. We need 2 things: 

1. `add_yaml_block_ext(variant_extensions, preserve_yaml)` should be apply for each of those formats `markdown`,`markdown_phpextra`,`markdown_github`,`markdown_mmd`,`markdown_strict`. These are all the formats that supports the extension since a long time.
1. `if (pandoc_available(2.13)) add_yaml_block_ext(variant_extensions, preserve_yaml) else  variant_extensions}` should be apply for each of those formats:  `gfm`,`commonmark`,`commonmark_x`. That are the new format that supports the extensions since recently. 

This is why I used `switch()` to avoid a `if()` with long condition, as it offers a special syntax to list several cases and apply the same value

``` r
switch ('a',
    a = ,
    b = "I am a and b"
)
#> [1] "I am a and b"
```

I find it clearer to read to list all the formats also. 

Otherwise it would be 

```suggestion
  set_extension <- function(format, ext, add = TRUE) {
    ext <- paste0(ifelse(add, "+", "-"), ext)
    if (grepl(ext, format, fixed = TRUE)) return(format)
    paste0(format, ext, collapse = "")
  }

  add_yaml_block_ext <- function(extensions, preserve_yaml) {
    set_extension(variant_extensions, "yaml_metadata_block", preserve_yaml)
  }

  # yaml_metadata_block extension
  if ((variant_base %in% c('markdown', 'markdown_phpextra', 'markdown_github', 'markdown_mmd', 'markdown_strict')) || 
      (variant_base %in% c('gfm', 'commonmark', 'commonmark_x') && pandoc_available('2.13'))) {
    variant_extensions = add_yaml_block_ext(variant_extensions, preserve_yaml)
  }
```

This would be longer lines but less lines, and less easier to read than the switch that allows nice reading of supported format with the indentation. However, it repeat the function `add_yaml_block_ext()` indeed. 

In your suggestion, if `variant_base == "gfm"` then, `variant_extensions = set_extension(variant_extensions, preserve_yaml)` would not be apply but we want it to be applied. am I missing something ? 

Also, I would like to keep `set_extension()` with no default, as I think this could be useful in other places.

yihui: I see. I've never used `switch(x, a =, b = value)` before, so I misunderstood it. Sorry about that.

yihui (approved):
Please feel free to merge after adding a news item. Thanks!

R/md_document.R
cderv: No worries. I am glad you can also learn from me 😉 😄

cderv:
Do we want to merge this for the next release mid-august planned for the IDE ? 
We discussed last time to only release current change.

I am hesitant as there would not be much time for this to live in dev version in case I missed something. 
How do you feel about that ?

yihui:
I don't feel this is a risky change that would affect users widely, so I'm okay with merging it into the next release.

PR #2193, opened 2021-07-26
Add a forced mode to use bslib for bootstrap

cderv:
Use bslib package for boostrap HTML dependency instead of rmarkdown internal hard HTML dependency

This PR adds an opt-in mode to force the use of **bslib** for bootstrap dependency

- Either opt-in by using setting option `rmarkdown.bslib_forced_mode` to TRUE
- Or either when `rmd/h/bootstrap` folder is not more present in installed directory in local library

When doing that, this will force the use of Boostrap using **bslib**. Currently version 3 will be used, and any string theme will be used as `boostwatch` theme. This means that when forced mode is ON this 

````yaml
output:
  html_document:
    theme: cerulean
````

will use `bslib::bs_theme(version = 3, theme = "cerulean")`

Without forced mode, it would have been 

```yaml
output:
  html_document:
    theme: 
    version: 3
      bootswatch: cerulean
```

This PR can be merged safely I believe for next version as it will only have effet when activated. 

This PR is the first step to try using **bslib** as a default for **rmarkdown** 

cc @apreshill @yihui

### DIFF
--- a/R/html_dependencies.R
+++ b/R/html_dependencies.R
@@ -41,6 +41,8 @@ html_dependency_jqueryui <- function() {
 #' @export
 html_dependency_bootstrap <- function(theme) {
   theme <- resolve_theme(theme)
+
+  # Bootstrap with bslib package
   if (is_bs_theme(theme)) {
     # TODO: would it make sense for these additional rules to come as a part of
     # bslib::bs_theme_dependencies() (for consistency sake)?
@@ -53,6 +55,8 @@ html_dependency_bootstrap <- function(theme) {
     )
     return(bslib::bs_theme_dependencies(theme))
   }
+
+  # Rmarkdown own BS3 dependency
   htmlDependency(
     name = "bootstrap",
     version = "3.3.5",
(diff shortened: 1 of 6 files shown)


R/html_dependencies.R
```diff
@@ -88,16 +92,35 @@ bootstrap_dependencies <- function(theme) {
...
+    if (identical(theme, "default")) {
+      theme <- "bootstrap"
+    }
+
+    # special handling triggered when bootstrap folder removed from installation folder
+    # In this case, bslib is used as default.
+    bslib_forced_mode <- getOption("rmarkdown.bslib_forced_mode", FALSE)
+    if (bslib_forced_mode || !dir.exists(pkg_file("rmd/h/bootstrap"))) {
+      # we are in special bslib default mode, theme is tweaked to be a list
+      if (!is_available("bslib")) {
```
yihui, line 110: In both cases, I think we can assume that users know what they are doing, so I wonder if this explicit error message is necessary, i.e., if they set the option to `TRUE` or delete the bootstrap folder, they must already know that they will need to install **bslib**.

Another factor to consider is that **bslib** will be a hard dependency in the future. By that time, this check won't be necessary.

yihui (approved):
Looks good to me. Thanks!

R/html_dependencies.R
cderv: I'll leave as is for now. and remove later when bslib will be hard dependency and dependening on feedback we have. 

Thanks!

cderv:
Oh and also @yihui I did not document that for now in NEWS so that is stays a bit hidden, and because the option will surely disappear once we use bslib as default. 

I can add a bullet to NEWS if it is better.

yihui:
No, no need to add a news bullet. Thanks!

Issue #1859, opened 2020-07-08
Enhancement: support for non-standard RMarkdown directory structures.


jonathan-g:
I have the same issue as was raised in #146 (and I originally posted this issue as a comment there).

I organize my reveal.js slides for a semester in a directory structure:

    semester:
        lecture_lib (common material: css, js, etc.)
            assets (custom css themes, custom js, images, etc. that are shared across the whole project)
            revealjs
            font-awesome
        slides
            class_01
                assets (per-class assets)
                    images (images to include)
                    videos (video files)
                    fig (figures generated by RMarkdown)
            class_02
                assets (per-class assets)
                    images (images to include)
                    videos (video files)
                    fig (figures generated by RMarkdown)
            c;lass_03
                assets (per-class assets)
                    images (images to include)
                    videos (video files)
                    fig (figures generated by RMarkdown)
            ...

This lets me keep a consistent structure for each lecture.

If I use pandoc 2.7.x, RMarkdown works fine with my directory structures, but if I use any later version of Pandoc rendering the .Rmd file fails with an error message about `relativeTo()` because `lib_dir` is not a descendant of `output_dir`.

This is not a bug per-se since [by design](https://github.com/rstudio/rmarkdown/issues/146#issuecomment-48591200), RMarkdown does not support this kind of directory structure. But it is something that a number of people have expressed interest in (#146, #472, https://github.com/jayhesselberth/widgetdown/issues/4, https://github.com/jdblischak/workflowr/issues/95).

My non-standard (and unsupported) directory structure worked for many years, using a custom `reveal.js` HTML template for Pandoc to handle a more flexible directory structure but it broke several months ago when Pandoc moved to version 2.9 and rmarkdown introduced `html_dependency_header_attrs` to fix the duplication of HTML class attributes from `<section>` to `<h[1-9]>` tags that Pandoc introduced with version 2.8 or 2.9. I was able to work around this by continuing to work with an older Pandoc version (2.7).

I raise this in part because I think I could address this with a relatively minor PR (I have this working on a fork of rmarkdown at https://github.com/jonathan-g/rmarkdown/tree/minimal-tree-fix, and it changes about 14 lines of code, all of which are in the `html_dependencies_as_string()` function). I need to test this patch more before submitting it as a PR, but it should allow this non-standard directory structure through the YAML `lib_dir` parameter without disturbing any use cases except where `lib_dir` begins with `..`. 

---

By filing an issue to this repo, I promise that

[...]

Issue #146, opened 2014-06-28
Multiple Rmd files in subdirectories and common lib_dir


gaborcsardi:
I have a bunch of Rmd files, spread in some subdirectories, e.g. 

```
aaa/bbb/one.Rmd
aaa/ccc/two.Rmd
...
```

I want to convert them to HTML, and use a common `lib_dir` for all of them. The problem is, that if I do

```
render("aaa/bbb/one.Rmd", NULL, 
       output_options=list(self_contained=FALSE, lib_dir="lib"))
```

then the `lib` directory (and more importantly the the references in the generated HTML accordingly) end up in `aaa/bbb/lib`, which is obviously not what I want.

If I try

```
setwd("aaa/bbb/")
render("one.Rmd", NULL, output_options=list(self_contained=FALSE, lib_dir="../../lib"))
```

then I get

```
Error in relativeTo(basepath, dir) : 
  The path /private/tmp/lib/jquery-1.11.0 does not appear to be a descendant of /private/tmp/aaa/bbb/
```

Is there a way to create a common `lib_dir` for these files, and still get the links in the HTML files right?

sanjmeh:
### This is a strange restriction. 
If I have understood it correct, the `lib_dir` is expected to be in the same directory as `output_dir` argument of `rmarkdown::render()`. This is a big limitation as it forces all the output html files to be saved in the same parent directory where the `.Rmd` is present. If we want to use the same `.Rmd` to generate a separate bunch of html files (based on the argument `output_dir`), we cannot share the same `lib_dir`. I have 50+ html files generated in every pass, and cannot abandon the subdirectory structure. Do we have a solution to this?

jonathan-g:
I have the same issue.

I organize my reveal.js slides for a semester in a directory structure:

    semester:
        lecture_lib (common material: css, js, etc.)
            assets (custom css themes, custom js, etc.)
            revealjs
            font-awesome
        slides
            class_01
                assets (per-class assets)
                    images (images to include)
                    videos (video files)
                    fig (figures generated by RMarkdown)
            class_02
                assets (per-class assets)
                    images (images to include)
                    videos (video files)
                    fig (figures generated by RMarkdown)
            c;lass_03
                assets (per-class assets)
                    images (images to include)
                    videos (video files)
                    fig (figures generated by RMarkdown)
            ...

This lets me keep a consistent structure for each lecture.

[...]

PR #2199, opened 2021-08-07
Allow `lib_dir` not to be a descendant of the base directory for documents derived from `html_document_base`

j-magnolia:
This PR addresses the issues raised in #146 and #1859.

It's a somewhat big PR, but a lot of the changes are documentation for the new functionality and a couple of testthat tests 
to make sure it works properly.

Changes introduced in this PR:
* Introduce a new argument `allow_uptree_lib_dir` to `html_document_base()` that turns on new functionality in handling 
  HTML dependencies, so `lib_dir` does not need to be a descendant of the base directory.
  * This can be a function argument or a parameter in the `output` block of the YAML header in a `.md` or `.Rmd` file.
  * This should work for any HTML format that uses `html_document_base`.
* Add functions to `html_dependencies.R` to implement the new functionality.
* Add documentation of the new functionality
  * The principal documentation is in a new section on directory structures in `html_document`.
* Add `testthat ` tests for new functionality.

The reason for requiring a new argument/parameter for enabling this functionality is because I'm a bit paranoid about 
accidentally breaking anything that depends on the old behavior, which would throw an error if `lib_dir` was not a descendant
of the base direcctory.

Anyone who wants to install the rmarkdown package with this PR applied can do so by installing from my fork: `remotes::install_github("jonathan-g/rmarkdown", ref="jg-tree-fix")`.  I try to keep my PR fork up to date with the main branch in this repository.


ma-riviere:
Is there any ETA for this (awesome) update ?

jonathan-g:
> Is there any ETA for this (awesome) update ?

We must be patient, because it takes time and effort to review a PR.

This one ended up larger than I had expected: The basic code is small, but I added documentation and unit tests, and those ended up much larger than the new code. I think it's worth the extra lines to document the changes and to add tests to make sure they work and don't break other parts of  `rmarkdown`. But this makes the PR larger, and adds to the time and effort it will take to review it, so we must be patient.

In the meantime, while we are waiting for this to be reviewed and a decision to be made about whether to accept the PR, you can always install a version of rmarkdown that includes this change via `remotes::install_github("jonathan-g/rmarkdown", ref="jg-tree-fix")`. I am actively keeping the PR branch of my code up to date with new commits on the main branch of `rmarkdown`, so it should always be approximately up to date with the latest version of `rmarkdown`, but with the added functionality.

ma-riviere:
Thank you for your reply, and for your work on this issue !

I'm eagerly waiting for this update to better structure / factorize my slides & reports' code. In the meantime, I'll give your branch a try :smile:

yihui:
Thank you for this PR and for your patience over the years — the underlying need is real and well-motivated, and the issues you linked (#146, #1859) have been asking for it for a long time.

After reviewing, I've opened #2632 as a smaller alternative that achieves the same goal. The key difference: rather than gating the behavior behind a new `allow_uptree_lib_dir` argument, it auto-detects when `lib_dir` is not a descendant of the output directory and emits an up-tree relative reference (`../lib/...`) in that case. Because an up-tree `lib_dir` previously only ever errored, enabling it automatically cannot regress any currently-successful render, so no new argument or opt-in is needed. It reuses `xfun::relative_path()` (already a dependency) instead of forking `htmltools::copyDependencyToDir()`, which keeps the change to ~16 lines of core logic. It also covers the reveal.js/slides case from #1859 for free, since those formats route through `html_document_base()`.

Your PR was invaluable in scoping the problem, the directory-structure documentation, and the test cases — that groundwork is what made the smaller version straightforward to write and verify. I'll make sure your contribution is credited.

Please let me know if #2632 solves your problem. If not, I'm happy to continue working on it. Thanks!

PR #2204, opened 2021-08-24
Rm pirates

apreshill:
Hi @yihui -

I got a message from @sarahemlin that there was some strange pirate talk in the HTML of one of our webpages, and I confirmed it was in there! This PR removes pirate talk 🏴‍☠️ I cannot build the site locally unfortunately- so please check before merging.

Possibly the weirdest PR I'm ever opened- would love to know the history here!
Alison


yihui (approved):
I had no idea about these pirate talk... I just checked the history and it seems they were from Robby, e.g., 17d48cb787f8c6417c9d966c5181d7a6d742574c. Since these elements are hidden by CSS, I have never noticed them.

I'll rebuild the site. Thanks!

PR #2215, opened 2021-09-14
Minimal example of server-side math in rmarkdown

jeroen:
This is mostly to test right now (it uses a feature in the katex package that is not yet on CRAN).

Here is a minimal example of what server-side math rendering could look like. I used this example [math.Rmd](https://gist.github.com/jeroen/32b8d15c7a0d373ab29714c4e5b479fe#file-math-rmd). In the front-matter you can flip `katex: true` to compare the mathjax output vs katex output.

I'd be interested to hear if this seems to work as expected.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -101,6 +101,7 @@ Suggests:
     tibble,
     fs,
     rsconnect,
+    katex (>= 1.3.0),
     withr (>= 2.4.2),
     bslib (>= 0.2.5.1),
     sass (>= 0.4.0)
--- a/R/html_document.R
+++ b/R/html_document.R
@@ -63,6 +63,8 @@
 #'  MathJax CDN. The "local" option uses a local version of MathJax (which is
 #'  copied into the output directory). You can pass an alternate URL or pass
 #'  \code{NULL} to exclude MathJax entirely.
+#'@param katex if \code{TRUE} math is rendered server-side using the katex package,
+#' such that mathjax is not needed in the html. This implies \code{mathjax = FALSE}.
 #'@param section_divs Wrap sections in \code{<div>} tags, and attach identifiers to the
 #'  enclosing \code{<div>} rather than the header itself.
 #'@param template Pandoc template to use for rendering. Pass "default" to use
@@ -231,6 +233,7 @@ html_document <- function(toc = FALSE,
                           highlight = "default",
                           mathjax = "default",
                           template = "default",
+                          katex = FALSE,
                           extra_dependencies = NULL,
                           css = NULL,
                           includes = NULL,
@@ -468,6 +471,13 @@ html_document <- function(toc = FALSE,
     args
   }
 
+  post_processor <- if(isTRUE(katex)){
+    mathjax <- FALSE
+    function(metadata, input_file, output_file, clean, verbose) {
+      katex::render_math_in_html(output_file, output = output_file)
+    }
+  }
+
   # return format
   output_format(
     knitr = knitr_options_html(fig_width, fig_height, fig_retina, keep_md, dev),
@@ -480,6 +490,7 @@ html_document <- function(toc = FALSE,
     pre_knit = pre_knit,
     post_knit = post_knit,
     pre_processor = pre_processor,
+    post_processor = post_processor,
     on_exit = on_exit,
     base_format = html_document_base(theme = theme,
                                      self_contained = self_contained,
--- a/man/html_document.Rd
+++ b/man/html_document.Rd
@@ -24,6 +24,7 @@ html_document(
   highlight = "default",
   mathjax = "default",
   template = "default",
+  katex = FALSE,
   extra_dependencies = NULL,
   css = NULL,
   includes = NULL,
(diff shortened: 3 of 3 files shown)


yihui (approved):
Thanks! This looks good to me! Let's see if other people have any thoughts.

jeroen:
The new version of katex is now on CRAN.

cderv:
Thanks @jeroen ! 

@yihui I believe we should consider this while also looking at the old PR #1940 about support more math engine. 

This current PR here suggest to add a `katex` argument in the output format function. The other (#1940) suggests a more generic `math` argument to configure various math engine supported by Pandoc. And I believe more option from Pandoc should be supported. It could easily support also support server-side math using **katex** R package. 

Unless we decide that current `mathjax` is enough while adding server-side rendering using **katex**, and for more options one could opt-out and use `pandoc_args` directly. 

What do you think ? 

cc @atusy who worked on #1940

yihui:
@cderv I agree with you. I'll revisit this PR when reviewing #1940. Thanks for the reminder!

jeroen:
Thanks. I could help refactoring if you like. I had made this pr minimal to
make it easy to review, and avoid too many changes at once.


Op ma 20 sep. 2021 14:21 schreef Yihui Xie ***@***.***>:

cderv:
@jeroen I am currently working on #1940 for broader math support. 

Idea is to have YAML like this 
````yaml
output:
  html_document: 
    math:
      engine: katex
      url: https://cdn.jsdelivr.net/npm/katex@0.11.1/dist/
````

or 

````yaml
output:
  html_document: 
    math: mathml
````


I would bring support for server side using **katex** R package within this new logic
Either using a new engine name 
````yaml
output:
  html_document: 
    math: r-katex
````

or leveraging the url part
````yaml
output:
  html_document: 
    math: 
      engine: katex
      url: local
````

Maybe you have other idea for this ? 

We could also have something more related to `self_contained = TRUE`: 
If `self_contained = TRUE` and `math = "katex"` and **katex** R package is installed, then we activate post processing of HTML using the R package.

**Comment by cderv** [comment]
@jeroen Global math support has been merged now. This can be provided 
```yaml
output:
  html_document:
    math_method: 
      engine: katex
      url: https://cdn.jsdelivr.net/npm/katex@0.11.1/dist/
```

We should leverage this yaml structure for supporting **katex** R package for server side rendering. 

I can do the update on this PR. Do you have suggestion for the way to call the method ? Thanks

cderv:
So I have looked closer to how to support this with our new logic. 

From what I understand, **katex** server side processing with the R package suppose two things: 

* Pandoc has processed the HTML file with a math flag (ideally `--katex` or `--mathjax`
* Pandoc does not include any library in the head of the HTML document. 

In the PR it was working `mathjax <- FALSE` still had the effect of adding `--mathjax` to the CLI. 

Anyway, I'll probably do another PR instead of modifying this one if you don't mind.

jeroen:
The [render_math_in_html](https://github.com/cran/katex/blob/master/inst/js/bindings.js) function queries the HTML document for nodes with `class="math inline"` and `class="math display"` and substitutes the corresponding math with html. 

Inside `math inline` nodes it looks for patterns with `$...$` or `\(...\)`  and in `math display` nodes it looks for a patterns with `$$...$$` or `\[...\]`. So it should work for pandoc output, either with `--mathjax` or without.

cderv:
Thanks for the precision. 

> So it should work for pandoc output, either with --mathjax or without.

Without setting a math engine would have been great but Pandoc will warn every time it converts to HTML in the case no math processing is done.
````
[WARNING] Could not convert TeX math 
  \frac{x}{y}
  , rendering as TeX
[WARNING] Could not convert TeX math \frac{x}{y}, rendering as TeX
````

So ideally we need Pandoc to process Math but have no JS / CSS added in HTML head.

cderv:
For the record, I don't think `render_math_in_html` works for an HTML processed by `pandoc --katex`. 
I got these warning for the package 
````
Warning: Did not find '$...$' or '\(...\)' equation in:
\frac{x}{y}

Warning: Did not find '$$...$$' or '\[...\]' equation in:
x + y

Warning: Did not find '$$...$$' or '\[...\]' equation in:
\frac{x}{y}
````

In fact, Pandoc does not wrap equation in special chars when `--katex` is set 
````
❯ pandoc -t html --katex
display
$$
\frac{x}{y}
$$

inline $\frac{x}{y}$
^Z
<p>display <span class="math display">
\frac{x}{y}
</span></p>
<p>inline <span class="math inline">\frac{x}{y}</span></p>
````

This would be the cleaner way IMO to apply the post processing for any document
* Set `--katex` so that pandoc process Math and create the correct HTML
* Remove `$math$` part in Pandoc's template if any so that no online JS / CSS are added.


For it to work with current **rmarkdown**: 
* Pandoc needs to be tricked by setting `--mathjax` flag (even if we won't use Mathjax)
* So one need to be sure no Mathjax resources are added into `<head>` by Pandoc or **rmarkdown**. Currently done with our default template where we have already removed this part. But for other Pandoc template, it would require some processing (which will be the case for our other HTML formats). 

This means that with non default template, setting 
````
html_document:
	math_method: r-katex
````
will add the CSS and JSS for Mathjax. Not ideal. 

Am I right to believe that it would make sense for **katex** R package to work on the result of `pandoc --katex` ? 
Is this something that could be handled in **katex** or it is tied to Mathjax like syntax ? I can open an issue

Anyway, I have opened #2304 to implement this based on current **rmarkdown** main instead of rebasing this PR. Hope it is fine.

cderv:
Closing this as #2304 has been merged. Thanks!

Issue #2196, opened 2021-07-28
Provide a syntax definition file with updated support for R new syntax 


cderv:
First issue has been reported in rstudio/bookdown#1157.  See discussion there for details. In a nutshell: 
* Pandoc's highlighting uses KDE XML syntax file 
* the KDE XML syntax file for R language does not correctly handles the new R syntax. 

Current fix in **rmarkdown** is for `html_document_base()` only as a post processing - see https://github.com/rstudio/rmarkdown/commit/99df4e7db5ba0108c51557639e661df05554bf1b . This fixes the issue rstudio/bookdown#1157

Proper fix should be: 

* Push upstream correct support for Pandoc for later Pandoc's version.
* Use the XML definition file directly in **rmarkdown** for previous pandoc version that won't have the updated support, using `--syntax-definition` flag.

cderv:
From a report in https://github.com/tidyverse/reprex/issues/389, let's acknowledge that the current fix is not working for `github_document()` HTML preview

````markdown
---
title: "Minimal Example"
output: 
  github_document: default
  html_document:
    highlight: pygments
---

```{r}
library(magrittr)
1.23 %>% round()
1.23 |> round()
```
````

**Comment by cderv** [comment]
Waiting for this to be completely fixed, I published a gist with a patched xml file one could use: https://gist.github.com/cderv/079b78caefbe45da6d1c03d737970b55

It can be downloaded then used by adding `pandoc_args` in YAML header. e.g with `pdf_document()`

````
output: 
  pdf_document:
    pandoc_args: "--syntax-definition=r.xml"
````

See comment here for more detail and example on how to set a font for having the ligature in PDF

https://github.com/rstudio/bookdown/issues/1157#issuecomment-933366422

**Comment by yihui** [comment]
Could you submit a PR to KDE? I guess this is the file to edit: https://invent.kde.org/frameworks/syntax-highlighting/-/blob/master/data/syntax/r.xml The earlier we submit a PR, the better, since it will take time for the upstream change (if merged) to propagate to Pandoc.

**Comment by cderv** [comment]
I wanted to try add support for new syntax to define function too
``` r
(\(x) x + 1)(2)
#> [1] 3
```

But I still haven't done the KDE file update for this. 

You're right I should do it wit the current addition.

cderv:
[...]

PR #2228, opened 2021-10-04
Fix the syntax highlighting of |> in LaTeX/PDF output

yihui:
See https://github.com/rstudio/bookdown/issues/1157#issuecomment-932776299

Again, this is just a temporary fix, and may be removed in the future after #2196 is fixed.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: rmarkdown
 Type: Package
 Title: Dynamic Documents for R
-Version: 2.11.2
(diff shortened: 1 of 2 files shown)


cderv (approved):
yes you're right - We could do the same kind of patch. 

Maybe it is also better than using our own XML definition file, which I was thinking about. However, this would allow the issue to be fixed as new syntax definition supported in all pandoc related format using syntax highlighting. 

Anyway, this will be for later so this is fine for now I guess and it is safe.

PR #2271, opened 2022-01-05
Add dir variable to HTML templates

### LINKED ISSUE #2063: Missing `dir` attributes in HTML template
Pandoc HTML template has a `dir` attributes: 
https://github.com/jgm/pandoc-templates/blob/master/default.html4#L2
```html
<html xmlns="http://www.w3.org/1999/xhtml"$if(lang)$ lang="$lang$" xml:lang="$lang$"$endif$$if(dir)$ dir="$dir$"$endif$>
```
which allow to select the language direction for the whole document. https://pandoc.org/MANUAL.html#language-variables

I think we should add its support too.

cderv:
This closes #2063 by adding the variable to the different templates using a if clause. 

doc about this attribute: https://developer.mozilla.org/fr/docs/Web/HTML/Global_attributes/dir

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: rmarkdown
 Title: Dynamic Documents for R
-Version: 2.11.5
+Version: 2.11.6
 Authors@R: c(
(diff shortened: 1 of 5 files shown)


yihui (approved):
Looks good!

PR #2275, opened 2022-01-10
Add a LICENSE.md file in repo

### LINKED ISSUE #1766: Add separate licensing information
Not much of an issue but it would be informative to add a separate license file for this repo rather than relying on the `DESCRIPTION` file.

cderv:
closes #1766

Is this addition ok with you @yihui ? 
Main advantages are that this file will get picked up by Github to show the license, and also by pkgdown to make it available on the website as a HTML page in the website. (works currently and point to https://www.r-project.org/Licenses/GPL-3) 

Currently, it is just a copy of the full license as I understand this is good practive to have in a project. I wonder if we should do that in all our repo. 

Also, there may be other things to do to get ok considering we include file with other licenses. 🤔 

Anyway, not that important but seems to be good practice to have a copy of the license in the repo directly.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -17,3 +17,4 @@
 ^reference$
 ^vignettes/articles$
 ^CITATION\.cff$
+^LICENSE\.md$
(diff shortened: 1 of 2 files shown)


yihui (approved):
I don't have an opinion. Please feel free to merge. Thanks!

PR #2281, opened 2022-01-12
add insert image rstudio addins

yufree:
#2280 The original idea and code is copied and modified from `blogdown` package's rstudio [addins](https://github.com/rstudio/blogdown/blob/6ed98d75017f095c9273532de18d712b60af424c/inst/rstudio/addins.dcf). It will introduce two extra suggest packages (`rstudioapi` and `miniUI`).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -81,7 +81,9 @@ Suggests:
     tibble,
     tufte,
     vctrs,
-    withr (>= 2.4.2)
+    withr (>= 2.4.2),
+    rstudioapi,
+    miniUI
 VignetteBuilder: knitr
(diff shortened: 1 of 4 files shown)


yihui (commented):
If we add this addin to **rmarkdown**, I tend to retire the addin in **blogdown**, otherwise we have to maintain two similar addins.

That means this addin in **rmarkdown** has to also support **blogdown** when it detects a **blogdown** site (see if `blogdown:::site_root()` throws an error).

yihui:
On second thought, I feel it makes more sense to let this addin stay in **blogdown**. I have just made it work with any Rmd documents, i.e., it also works outside site projects now. Thanks!

PR #2285, opened 2022-01-14
Support custom highlight theme in other formats

### LINKED ISSUE #2035: Allow to load external highlighting theme for Pandoc HTML and PDF
I have a customized `pandoc` syntax highlighting theme that I would like to use for my `Rmd` file: `my.theme`. For PDF output this can be achieved easily by using the following YAML header:

    ---
    title: "Test"
    author: "Me"
    date: "2/8/2021"
    output:
      pdf_document:
        pandoc_args: "--highlight-style=my.theme"
    ---

Unfortunately the same wouldn't work for `html_document`. While the console output shows `--highlight-style=my.theme` was successfully passed to `pandoc`, the `pandoc_html_highlight_args()` utility attached an argument `--no-highlight` to the end which then overtakes the previous argument. Stating `highlight: my.theme` is not an option either since `pandoc_html_highlight_args()` will do a `match.arg` check. The only work around I can think of is to use `trace()` to intercept the `pandoc_args` inside `render()` before it was passed to `pandoc_convert()`, and this is obviously a terrible idea.

A fix shouldn't be too complicated though. Maybe an option `customized` can be added to the output of `html_highlighters()`? In which case `pandoc_html_highlight_args()` will stop attaching further arguments so that `--highlight-style` can be processed correctly.

---

cderv:
This PR follows #1941 by adding support for custom highlight `.theme` file in other format like PDF 

this will close #2035 

It uses the same internal function as previous PR to support HTML but added a logic for non HTML format too. 

I also brought back backward compatibility for partial matching of Pandoc's style name are `match.arg` was used before and also took the opportunity to add informative error message if unknown value is provided in order to error from R and not from Pandoc

### DIFF
--- a/R/pandoc.R
+++ b/R/pandoc.R
@@ -606,13 +606,9 @@ is_highlightjs <- function(highlight) {
 }
 
 resolve_highlight <- function(highlight, supported = highlighters()) {
-  is_supported <- TRUE
   # for backward compatibility, partial match still need to work
-  highlight <- tryCatch(
-    error = function(e) { is_supported <<- FALSE; highlight },
-    match.arg(highlight, supported)
-  )
-  if (is_supported) return(highlight)
+  i <- pmatch(highlight, supported, nomatch = 0L)
+  if (i > 0L) return(supported[i])
 
   # Otherwise it could be a custom (built-in) .theme file
   if (!pandoc2.0()) {
--- a/R/pandoc.R
+++ b/R/pandoc.R
@@ -628,7 +628,17 @@ resolve_highlight <- function(highlight, supported = highlighters()) {
     rstudio = pkg_file_highlight("rstudio.theme")
   )
   # if not an alias use the provided custom path
-  custom[[highlight]] %||% highlight
+  highlight <- custom[[highlight]] %||% highlight
+  # Check for extension or give informative error otherwise
+  if (!identical(xfun::file_ext(highlight), "theme")) {
+    msg <- c(
+      sprintf("`highlight` argument must be one of %s",
+              knitr::combine_words(c(supported, names(custom)), and = " or ", before = "`")),
+      " or a file with extension `.theme`."
+    )
+    stop(msg, call. = FALSE)
+  }
+  highlight
 }
 
 #' Find the \command{pandoc} executable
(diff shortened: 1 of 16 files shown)


R/pandoc.R
```diff
@@ -624,6 +605,42 @@ is_highlightjs <- function(highlight) {
...
 }
 
+resolve_highlight <- function(highlight, supported = highlighters()) {
+  is_supported <- TRUE
+  # for backward compatibility, partial match still need to work
+  highlight <- tryCatch(
+    error = function(e) { is_supported <<- FALSE; highlight },
+    match.arg(highlight, supported)
+  )
+  if (is_supported) return(highlight)
```
cderv, line 615: I did this change to bring back the partial matching feature that was there when `match.arg()` was used. Meaning
this would work. 
````yaml
output:
  html_document: 
    highlight: breez
````
It was not following my other PR.

File: R/pandoc.R [hunk 2]
```diff
@@ -624,6 +605,42 @@ is_highlightjs <- function(highlight) {
...
+  # Check for extension or give informative error otherwise
+  if (!identical(xfun::file_ext(highlight), "theme")) {
+    msg <- c(
+      sprintf("`highlight` argument must be one of %s",
+              knitr::combine_words(c(supported, names(custom)), and = " or ", before = "`")),
+      " or a file with extension `.theme`."
+    )
+    stop(msg, call. = FALSE)
+  }
+  highlight
```
cderv, line 641: I then added that to throw early a useful message if the `highlight` value is not one of the expected (either one supported theme or a file path to a `.theme` file). 

This makes code less simple than in the other PR but it brings back useful message.

R/pandoc.R
```diff
@@ -624,6 +605,42 @@ is_highlightjs <- function(highlight) {
   !is.null(highlight) && (highlight %in% c("default", "textmate"))
 }
 
+resolve_highlight <- function(highlight, supported = highlighters()) {
+  is_supported <- TRUE
+  # for backward compatibility, partial match still need to work
+  highlight <- tryCatch(
+    error = function(e) { is_supported <<- FALSE; highlight },
+    match.arg(highlight, supported)
```
yihui, line 613: Instead of `tryCatch(match.arg())`, I think you can just use `pmatch()` for partial matching.

yihui (approved):
Looks good to me. Thanks!

R/pandoc.R
cderv: Indeed you're right. I simplified in fca7a75. 
Thanks!

PR #2286, opened 2022-01-17
Support incremental for Powerpoint presentation

### LINKED ISSUE #2212: [FR] Add incremental option for PowerPoint presentations
Dependent on this PR being merged in (and thus a specific Pandoc version): https://github.com/jgm/pandoc/pull/7559

To help our users take advantage of this new incremental feature, we'd like for them to be able to specify the option in the YAML without resorting to Pandoc args:

```
output:
  powerpoint_presentation:
    incremental: true
```

From @cderv:
> This would be simple: Check for the pandoc version and add the flag. Warn that the value can't be changed (or error) if pandoc is too old.

cderv:
Closes #2212 

This PR adds support for `incremental = TRUE` in `powerpoint_presentation()` based on support in Pandoc 2.15

A warning will be issued when using Pandoc before 2.15

I added some quick test to check error and the argument. I think we should add more simple test like that just to be sure we don't break anything in the future. Seems like dummy tests but I guess that simple test are often the best.

I also took the opportunity to convert some other doc to Markdown syntax. 

I'll prepare a PR for the R Markdown book that we could merge following the next release

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: rmarkdown
 Title: Dynamic Documents for R
-Version: 2.11.8
+Version: 2.11.9
 Authors@R: c(
     person("JJ", "Allaire", , "jj@rstudio.com", role = "aut"),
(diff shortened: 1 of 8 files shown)


yihui (approved):
Looks good! Thanks!

Issue #2003, opened 2021-01-07
Math rendered with github_document changed between 2.7 and 2.11


cderv:
This is following observation on https://github.com/rstudio/rmarkdown/issues/1886 and it is related. 
Opening the issue just to track the change than happened between Pandoc version 

```sh
# pandoc 2.7.3
$ ./pandoc -t gfm <<< '$\frac{e^{i \pi}}{10}$'
\(\frac{e^{i \pi}}{10}\)
$ ./pandoc -t gfm+tex_math_dollars <<< '$\frac{e^{i \pi}}{10}$'
\(\frac{e^{i \pi}}{10}\)
# pandoc 2.11.3.2
$ pandoc -t gfm <<< '$\frac{e^{i \pi}}{10}$'
[WARNING] Could not convert TeX math '\frac{e^{i \pi}}{10}', rendering as TeX
$\\frac{e^{i \\pi}}{10}$
$ pandoc -t gfm+tex_math_dollars <<< '$\frac{e^{i \pi}}{10}$'
$\frac{e^{i \pi}}{10}$
$ pandoc -t gfm+tex_math_single_backslash <<< '$\frac{e^{i \pi}}{10}$'
The extension tex_math_single_backslash is not supported for gfm
```

How does it look with `github_document` ? 

``` r
dir.create(tmp_dir <- tempfile())
owd <- setwd(tmp_dir)
xfun::write_utf8(c(
  "---",
  "title: test",
  "---",
  "",
  "$e^{i \\pi}$",
  "",
  "$\\frac{e^{i \\pi}}{10}$",
  ""
), "test.Rmd")

# With Pandoc 2.7 (RStudio 1.3)
pandoc_2.7 <- fs::path_home("Logiciel/pandoc/pandoc-2.7.3")
rmarkdown::find_pandoc(cache = FALSE, dir = pandoc_2.7)
#> $version
#> [1] '2.7.3'
#> 
#> $dir
#> [1] "C:/Users/chris/Logiciel/pandoc/pandoc-2.7.3"
res <- rmarkdown::render("test.Rmd", "github_document")
#> processing file: test.Rmd
#> output file: test.knit.md
[...]

PR #2301, opened 2022-02-03
Support webtex for Github documents format 

### LINKED ISSUE #1886: Improved math rendering in github documents
Hello,

I recently stumble upon this solution for having math formulas in github readme documents which provide much better quality images than the current system: https://gist.github.com/a-rodin/fef3f543412d6e1ec5b6cf55bf197d7b

Could it be a good improvement?

cderv:
This should resolve #1886 and #2003

This adds a `math_method` argument to `github_document()` so that equation can be rendered as an image in the resulting md document

````markdown
---
output:
  github_document:
    math_method: "webtex"
---

Some content

$$
\int \frac{1}{x} dx = \ln \left| x \right| + C
$$
````

is rendered as 

----

Some content

\\int \\frac{1}{x} dx = \\ln \\left\| x \\right\| + C

----

We ended up activating by default in github document where there is math. Like in Quarto. 

We default to using PNG format with a DPI of 110 and using white background because otherwise (using svg for example) the equation would not be seen on dark mode in Github. We can't change the color font with the API. 
However, Quarto defaults to SVG because this gives way cleaner output. Quarto documents how to change the background using PNG. Quarto doc about this: https://quarto.org/docs/output-formats/gfm.html#webtex-math 

Should we default to the same ? 

One thing is I am not sure about how we should consider CodeCog fair usage policy. https://editor.codecogs.com/docs/3-fair_usage.php, but I guess it is ok.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: rmarkdown
 Title: Dynamic Documents for R
-Version: 2.11.21
+Version: 2.11.22
 Authors@R: c(
     person("JJ", "Allaire", , "jj@rstudio.com", role = "aut"),
     person("Yihui", "Xie", , "xie@yihui.name", role = c("aut", "cre"), comment = c(ORCID = "0000-0003-0645-5666")),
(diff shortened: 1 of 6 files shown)


yihui (approved):
Looks great. Thanks!

PR #2304, opened 2022-02-07
Support R KaTeX package as a new `math-method`

cderv:
This is based of #2215 and replaces it. It is based on new math support in **rmarkdown** to offer support for R package **katex** (https://docs.ropensci.org/katex/) 

````yaml
output:
  html_document:
    math_method: r-katex
````

If R package is not available, then it fallbacks to use `math_method: katex`. 

Regarding the implementation,  Pandoc requires us to set some math method otherwise some warnings will be thrown (`[WARNING] Could not convert TeX math `) by Pandoc. And **katex** R packages only handles the syntax resulting by `--mathjax` flag for Pandoc (and not `--katex`). 

This leads to a limitation: On a template where `$math$` variable is included into Pandoc's template, the Mathjax JS library will be inserted in `<head>`. For our default template, we don't have it, but it could happen with non default template (`template` arg in `html_document_base()`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -76,6 +76,7 @@ Suggests:
     fs,
     rsconnect,
     downlit (>= 0.4.0),
+    katex (>= 1.3.0),
     sass (>= 0.4.0),
     shiny (>= 1.6.0),
     testthat (>= 3.0.3),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: rmarkdown
 Title: Dynamic Documents for R
-Version: 2.11.15
+Version: 2.11.16
 Authors@R: c(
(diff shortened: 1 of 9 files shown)


cderv:
@jeroen here is how I would integrate **katex** to be used with **rmarkdown** document. 

Does it looks good to you ? 

Thanks for the katex update BTW. It is cleaner to use `--katex`. Only drawback still existing is that when used with non default template, Pandoc will still insert some JS and CSS for katex. But our templates does not have that so we're good.

jeroen:
I think so. I would personally prefer to error instead of warn if the user sets `r-katex` but the katex package is not installed. If the document gets rendered in CI, people may miss the warning and end up with something that they did not expect.

cderv:
Good thinking. I'll change to an error.

yihui (approved):
Looks good. Thanks!

Issue #2272, opened 2022-01-09
shiny_prerendered server contexts cannot be empty


jjallaire:
Originally reported by @jcheng5 here (https://github.com/quarto-dev/quarto-cli/issues/88):

I would've expected empty Shiny server contexts to be allowed:

    ```{r}
    #| context: server
    ```

But if the server context is empty (or contains only whitespace), you get this error:

```
Error in shiny_prerendered_app(target_file, render_args = render_args) :
  No server contexts or server.R available for density-contours.qmd
```

Follow up from @cderv:

````markdown
---
title: "Hello Prerendered Shiny"
output: html_document
runtime: shinyrmd
---

```{r, echo=FALSE}
sliderInput("bins", "Number of bins:", min = 1, max = 50, value = 30)
```

```{r, context="server"}
```
````

````r
Backtrace:
    x
 1. \-rmarkdown::run(...)
 2.   \-rmarkdown:::shiny_prerendered_app(target_file, render_args = render_args)
Erreur dans shiny_prerendered_app(target_file, render_args = render_args) : 
  No server contexts or server.R available for C:/Users/chris/Documents/test.Rmd
````

Support for empty server context should probably be added there.

cderv:
I had a quick look into this. 

This come as deep as **knitr** in https://github.com/yihui/knitr/blob/887fc44cadea648b562c2cf7065199b603975dd5/R/block.R#L236
````r
res = if (is_blank(code)) list() else if (isFALSE(ev)) {
    as.source(code)
  } else if (cache.exists && isFALSE(options$cache.rebuild)) {
    fix_evaluate(cache$output(options$hash, 'list'), options$cache == 1)
  } else in_dir(
    input_dir(),
    evaluate(
      code, envir = env, new_device = FALSE,
      keep_warning = !isFALSE(options$warning),
      keep_message = !isFALSE(options$message),
      stop_on_error = if (is.numeric(options$error)) options$error else {
        if (options$error && options$include) 0L else 2L
      },
      output_handler = knit_handlers(options$render, options)
    )
  )
````

If the code in a chunk is empty, then `evaluate` hook won't be called. In the context of a shiny prerendered document, this means that no server context will be registered in the HTML document., even an empty one. So `shiny_prerendered_app()` throw the error because it can't find a server context among the HTML document context node as none have been added for `server` during  knitting. 

To make this work, we would need a way to trick **knitr** somehow. Or maybe just initialize an empty server if none is found 🤔 
[...]

PR #2305, opened 2022-02-11
Allow empty server context in shiny prerendered

cderv:
This closes https://github.com/rstudio/rmarkdown/issues/2272 

After looking into other more complex solution, I think this is the easiest and the more targeted to the issue. We don't change **knitr** behavior but only empty chunk with `context = "server"` by defaulting to a empty server function in that case.

Example provided is 
````markdown
---
title: "Hello Prerendered Shiny"
output: html_document
runtime: shinyrmd
---

```{r, echo=FALSE}
sliderInput("bins", "Number of bins:", min = 1, max = 50, value = 30)
```

```{r, context="server"}
```
````

With current **rmarkdown** version, it will lead to an error because no value is passed to context and no `server.R` file is provided. 
As explained in https://github.com/rstudio/rmarkdown/issues/2272#issuecomment-1034009694, this is because **knitr** won't call evaluate hook on an empty code chunk, and so the shiny prerendered magic does not apply. 

This PR insert a comment in the code chunk if an empty server is provided so that knitr trigger the evaluate hook and the context is registered in the HTML.  `shiny_prerendered_app()` generate this server function, 
````r
  server <- function(input, output, session) {
    eval(xfun::parse_only(.server_context))
  }
````
which will result in an empty server function for the app if `.server_context` is only a comment. 

Only possible breaking change I am seing is that previously, if you had a empty server context but a `server.R` file, the app would render. With this change, the `server.R` file won't be used. I don't think this situation would happen so I think it is fine to change. 

Do you think that solves your issue in the first place ? Do you see any side effect I may have not think of ?

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -112,6 +112,15 @@ is_null_or_string <- function(text) {
   is.null(text) || (is.character(text) && (length(text) == 1))
 }
 
+# TODO: Export knitr:::is_blank() in xfun, and use here
+is_blank <- function(x) {
+  if (length(x)) {
+    all(grepl("^\\s*$", x))
+  } else {
+    TRUE
+  }
+}
+
 read_utf8 <- function(file) {
   if (inherits(file, 'connection')) con <- file else {
     con <- base::file(file, encoding = 'UTF-8'); on.exit(close(con), add = TRUE)
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: rmarkdown
 Title: Dynamic Documents for R
-Version: 2.11.20
+Version: 2.11.21
 Authors@R: c(
     person("JJ", "Allaire", , "jj@rstudio.com", role = "aut"),
     person("Yihui", "Xie", , "xie@yihui.name", role = c("aut", "cre"), comment = c(ORCID = "0000-0003-0645-5666")),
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: R/util.R [hunk 1]
```diff
@@ -112,6 +112,15 @@ is_null_or_string <- function(text) {
   is.null(text) || (is.character(text) && (length(text) == 1))
 }
 
+# TODO: Export knitr:::is_blank() in xfun, and use here
```
yihui, line 115: Done: https://github.com/yihui/xfun/commit/6f4eddd0d74974a12bbc431053849495ac6bd1bc

yihui (approved):
This seems to be a reasonable workaround to me. Thanks!

R/util.R
cderv: Thanks ! 

Should we use it now here ? Or usually, you go through this kind of TODO after a new **xfun** is released ?

yihui: No, not now, but after the next **xfun** release.

Yes, I search for "TODO" in each package before I start thinking about making a CRAN release of that package.

cderv: Perfect. I'll try to remember that ! 👍

yihui: Usually a TODO comment is not urgent to me, and can be done at any time in the future (not necessarily before the next release) if I remember.

cderv:
@jcheng5 I'll merge this. I think it is ok and won't bother your any longer for a review. Please feel free to react if otherwise. Thanks !

PR #2317, opened 2022-02-24
Fix dropdowns and navbar collapsing with Bootstrap 5

cpsievert:
For some context, see https://github.com/rstudio/bslib/issues/414

Also, it turns out navbar's are pretty broken in Bootstrap 5 with the latest bslib, but I'll be fixing that inside bslib

### DIFF
--- a/R/html_document.R
+++ b/R/html_document.R
@@ -769,6 +769,7 @@ navbar_links_tags <- function(links, depth = 0L) {
                 tags$a(
                   href = "#", class = "dropdown-toggle",
                   `data-toggle` = "dropdown", role = "button",
+                  `data-bs-toggle` = "dropdown", # BS5
                   `aria-expanded` = "false", link_text),
                 tags$ul(class = "dropdown-menu", role = "menu", submenuLinks)
         )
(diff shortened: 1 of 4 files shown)


yihui (approved):
Thanks!

guasi:
I wanted to add a new issue, but before doing so I want to make sure that my issue is not a duplicate of this one. 

**Conditions:** output: html_document: theme: version: 5
**problem:** The `<style>` element added via javascript to adjust the navbar is not added for the index page in bootstrap 5. This causes the header to hide under the navbar. Also, the active navbar is never highlighted when page is index. Looking at the [default.html](https://github.com/rstudio/rmarkdown/tree/main/inst/rmd/h/default.html) document, it seems the problem is the javascript section between lines 269-298. There is a problem with `menuAnchor`, but I don't know what it is. I just know that the style element the JS adds is not added when in index.html.

Can you tell me if I should add a new issue or if what ever fixes are done in bslib will take care of this problem?

cderv:
> Can you tell me if I should add a new issue or if what ever fixes are done in bslib will take care of this problem?

It would be helpful to open a new issue, and provided an example we can reproduce so that it is easier to look at it closely.

PR #2322, opened 2022-03-02
Add an `error` argument to `render()`

wurli:
This allows a user to render files which possibly include errors. This is simply passed to `knitr::opts_chunk$set()`, which previously had the value hard-coded.

### DIFF
--- a/R/render.R
+++ b/R/render.R
@@ -103,7 +103,7 @@ NULL
 #'
 #' Note that the \pkg{knitr} \code{error} option is set to \code{FALSE} during
 #' rendering (which is different from the \pkg{knitr} default value of
-#' \code{TRUE}).
+#' \code{TRUE}). However, this can be changed via the \code{error} option.
 #'
 #' For additional details on rendering R scripts see
 #' \link[=compile_notebook]{Compiling R scripts to a notebook}.
@@ -218,6 +218,8 @@ NULL
 #'   "Output created: " message, you can set \code{rmarkdown.render.message} to
 #'   \code{FALSE}
 #' @param encoding Ignored. The encoding is always assumed to be UTF-8.
+#' @param error Can be set to \code{TRUE} to pass over errors in rendered files.
+#'   This is internally passed to \code{rmarkdown::opts$set()}.
 #' @return
 #'   When \code{run_pandoc = TRUE}, the compiled document is written into
 #'   the output file, and the path of the output file is returned. When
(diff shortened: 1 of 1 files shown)


cderv:
Hi, 

Please open an issue before sending a PR so that we can discuss a proposal. 

I don't think any new argument is needed. You can also change the default behavior by setting the knitr option inside the Rmd document as usual, either per chunk, or as a global option

````markdown
---
title: "Untitled"
author: "TEST"
date: "2/28/2022"
output: html_document
---

```{r setup, include=FALSE}
knitr::opts_chunk$set(error = TRUE)
```

```{r}
stop("I am an error")
```
````

If you want to set this outside the document, you could use a parameter inside the document, and use `params` in `render()`

````markdown
---
title: "Untitled"
author: "TEST"
date: "2/28/2022"
output: html_document
params:
  stop_on_error: TRUE
---

```{r setup, include=FALSE}
knitr::opts_chunk$set(error = !params$stop_on_error)
```

```{r}
stop("I am an error")
```
```` 

* `rmarkdown::render("test.Rmd")` will error
* `rmarkdown::render("test.Rmd", params = list(stop_on_error = FALSE))` will render showing the error in the output document


So this is already possible in several way. Does that work for you ?

**Comment by wurli** [comment]
Apologies - in future I'll open an issue first! 

Unfortunately your proposed alternative isn't quite what I'm looking for. In practise I'm using a loop to knit multiple R scripts at a time, taking advantage of the fact that `render()` will first 'spin' an R script into an Rmd document. While various hacky solutions are possible, e.g. adding `knitr::opts_chunk$set(error = TRUE)` to the start of each script before knitting, it seems like simply exposing this control would be a much neater solution, and one which I'm sure others would find themselves reaching for.

**Comment by cderv** [comment]
If I understand correctly you are using `render()` to convert R script to HTML, is that correct ? 

And those R scripts do not have any YAML header or specific chunk option on code that errors - those kind of special comments are part of spinning feature. https://bookdown.org/yihui/rmarkdown-cookbook/spin.html

Would using those feature help ? Or you just want to `render()` the R script to HTML without changing anything to the script making them more compatible for spinning ? 

There is one advanced **knitr** feature you could use: Any **knitr** option can be modified using a R option. So in your case
```r
options(knitr.chunk.error = TRUE)
rmarkdown::render("script.R")
```

This option will be catched during `knitr::knit()` and will replace the default one set by `render()`. 
I let you try this. You could easily create a wrapper, even using `withr::with_options()` to apply only on the render call

I really don't think adding an argument to `render()` for this is the solution. There is already a lot of argument, and modifying a knitr option does not seems like what an argument to `rmarkdown::render()` should do. 

Hope it helps

PR #2325, opened 2022-03-03
`match.arg()` can't determine choices when argument is an internal function

### LINKED ISSUE #2324: Issue with new 2.12 version and vignette building
First reported @hadley as **ggplot2** CI is failing for oldrel-4 (R 3.4.4) during vignette creation. 
https://github.com/tidyverse/ggplot2/runs/5409764795?check_suite_focus=true#step:5:4808

I added 3 more R versions to tests **rmarkdown** (f46a1756c5ce5c0817c47ae92add7dbe03491c15) and I get the same issue: 
https://github.com/rstudio/rmarkdown/runs/5410652398?check_suite_focus=true

It seems there is something that changed between 3.4.4 et 3.5.3 that impact our new internal function usage. 

We need to look into and probably do a patch release - 2.12.1 ideally, but will be 2.13 with current scheme.

cderv:
This fixes #2324 

 in R 3.4.4 and possibly earlier version, `match.arg()` does not play well when argument is an internal function. It needs to be more explicit.


cderv:
@yihui this PR fixes the issue from #2324 but also a few others found along the way of adding 4 R versions in our CI. We should probably do that in other repo to prevent such issue after a release. 

I let you merge (and tweak if necessary). Is it possible to do a patch release to CRAN ?

yihui (approved):
Looks good. Thanks!

I'd like to wait for a couple of more days before making the patch release, in case we miss other problems in the last release that haven't been reported. The affected packages may try to install the dev version of rmarkdown in their GHA at the moment.

yihui:
FYI the submission didn't pass the automated check, but the problem was not from us (https://github.com/glmmTMB/glmmTMB/issues/787#issuecomment-1061455809). Now we have to wait for a CRAN maintainer to let us pass.

PR #2345, opened 2022-04-08
`site_generator()` stops search at RStudio project root

### LINKED ISSUE #2344: Issue with knitting R Markdown documents
I am facing a strange issue in last 2-3 days. I am unable to knit any .Rmd file to any format (HTML/PDF/Word). First, I create an RStudio Project. Within the RStudio Project, from RStudio File Menu>New File > R Markdown, I choose any of the three output formats. I simply save the document without making any changes in the defaults in the Rmd file. Then I press the "knit" button in RStudio. I get the following error while rendering it.
```
processing file: _main.Rmd
Error in parse_block(g[-1], g[1], params.src, markdown_mode) : 
  Duplicate chunk label 'setup', which has been used for the chunk:
knitr::opts_chunk$set(echo = FALSE)
Calls: <Anonymous> ... process_file -> split_file -> lapply -> FUN -> parse_block
Please delete _main.Rmd after you finish debugging the error.
Execution halted
```
Isn't _main.Rmd supposed to be in bookdown? How can I get this error while knitting a single R Markdown file? Moreover, this _main.Rmd file is saved in a different location - it's the default working directory (while not in a project). 
Irrespective of this, if I try to knit it again, I get the error message

```
Error in render_book(input_file, output_format, envir = envir, preview = TRUE) : 
  The file _main.Rmd exists. Please delete it if it was automatically generated. If you are sure it can be safely overwritten or deleted, please set the option 'delete_merged_file' to true in _bookdown.yml.
Calls: <Anonymous> -> <Anonymous> -> <Anonymous> -> render_book
Execution halted
```

So, again it confirms that the underlying function is from bookdown instead of R Markdown. 
I have updated all the relevant software - downloaded knitr, R Markdown and bookdown from github. I have also installed latest daily build of tinytex (using tinytex::install_prebuilt("TinyTeX"))and RStudio. The session info is as follows;

```
xfun::session_info("Rmarkdown")
R version 4.1.3 (2022-03-10)
Running under: Windows 10 x64 (build 19044), RStudio 2022.6.0.270

Locale:
  LC_COLLATE=English_India.1252 
  LC_CTYPE=English_India.1252   
  LC_MONETARY=English_India.1252
  LC_NUMERIC=C                  
  LC_TIME=English_India.1252    
system code page: 65001

Package version:
  Rmarkdown_2.13.3
```
<!--

How can I solve this issue?
Screenshot 2022-04-07 063014

cderv:
This fixes #2344 by making sure that we don't look too far up the folder trees when looking for a project root. 

I added a `stop_when` condition which should be a function that will take the current `dir` analysed and return `TRUE` is stop condition is met. This allows to keep the `proj_root` function quite generic. However, we could maybe internalize the stop condition ? 

`proj_root()` is also used in here 
https://github.com/rstudio/rmarkdown/blob/04a8bebb0a0d1a1489acecd0f3994608feca3fee/R/shiny_prerendered.R#L330
do you think it should also be stopped at `.Rproj` level ? 

I am not sure as the main issue here is to prevent the RStudio IDE to wrongly decide it is a project website to build.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: rmarkdown
 Title: Dynamic Documents for R
-Version: 2.13.3
+Version: 2.13.4
 Authors@R: c(
     person("JJ", "Allaire", , "jj@rstudio.com", role = "aut"),
     person("Yihui", "Xie", , "xie@yihui.name", role = c("aut", "cre"), comment = c(ORCID = "0000-0003-0645-5666")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 rmarkdown 2.14
 ================================================================================
 
+- Fix an issue with `site_generator()` detection of a site project (thanks, @bhattmaulik, #2344).
+
 - Fix an issue with Shiny prerendered documents and Pandoc not correctly rendering last markdown paragraph in HTML (thanks, @gadenbuie, #2336).
 
 - Fixed a bug that `site_generator()` fails to detect the root dir of the site and causes infinite recursion (thanks, @fisher-j, #2339).
--- a/R/render_site.R
+++ b/R/render_site.R
@@ -251,10 +251,12 @@ clean_site <- function(input = ".", preview = TRUE, quiet = FALSE,
 #' @rdname render_site
 #' @export
 site_generator <- function(input = ".", output_format = NULL) {
-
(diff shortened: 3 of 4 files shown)


yihui (approved):
Looks good to me. Thanks!

> do you think it should also be stopped at `.Rproj` level ?

Not sure about this one. Probably harmless to do the same thing there. I'm okay no matter if you want to do it or not.

PR #2362, opened 2022-05-24
Add support for native math by default in Github document

### LINKED ISSUE #2361: [FR] `github_document` no longer needs to create images for math expressions
Pretty excited about the news that GitHub now natively supports latex math expressions when displaying markdown documents. Thought I would flag as it might be nice to add an option to not parse $$ $$ into images for the `github_document` format (save from creating a lot of images in the GitHub repo)

https://github.blog/changelog/2022-05-19-render-mathematical-expressions-in-markdown/

$a^2 + b^2 = c^2$

$$
a^2 + b^2 = c^2
$$

cderv:
This closes #2361

This PR introduce `math_method = "default"`, at use it by default; This will make sure that maths are passed as-is to Markdown document as Github will now handle it using Mathjax for the rendering. 

For github document HTML preview, we use `mathjax` or `webtex` to make the math rendered.

Only uncertainty right now is which is the minimal version that supports its. I'll test.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,9 @@ rmarkdown 2.15
 
 - The `tufte_handout()` function inside **rmarkdown** is defunct now. Its codebase was moved to the **tufte** package in 2016, and this function was marked as deprecated in 2021. Please use `tufte::tufte_handout()` instead of `rmarkdown::tufte_handout()`. The latter will be removed eventually from this package.
 
+- `github_document()` gains `math_method = "default"` and defaults to it. No special processing will be done to inline maths in `$` and block maths 
+in `$$` as now [Github supports it](https://github.blog/changelog/2022-05-19-render-mathematical-expressions-in-markdown/) and will render using Mathjax (thanks, @kylebutts, #2361).
+
 
 rmarkdown 2.14
 ================================================================================
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,8 +3,7 @@ rmarkdown 2.15
 
 - The `tufte_handout()` function inside **rmarkdown** is defunct now. Its codebase was moved to the **tufte** package in 2016, and this function was marked as deprecated in 2021. Please use `tufte::tufte_handout()` instead of `rmarkdown::tufte_handout()`. The latter will be removed eventually from this package.
 
-- `github_document()` gains `math_method = "default"` and defaults to it. No special processing will be done to inline maths in `$` and block maths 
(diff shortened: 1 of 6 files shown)


cderv:
@yihui this is ready to review. Should be quick. 

I discovered that math feature is only available for Pandoc 2.0.4 and above. However, I silently ignore the `math_support` argument in that case. Would it be better to warn that math won't be process in that case ? 

Default is now the native support for github using Mathjax syntax. However, with Pandoc this is only available from Pandoc 2.10.1 I believe as it requires `+tex_maths_dollars` extensions and it seems it is not available before that. (I think `tex_math_double_backslash` was used before.  So for version between 2.0.4 and 2.10.1 I kept webtex as a default. 
I'll add that in the help page.

NEWS.md
```diff
@@ -3,6 +3,9 @@ rmarkdown 2.15
 
 - The `tufte_handout()` function inside **rmarkdown** is defunct now. Its codebase was moved to the **tufte** package in 2016, and this function was marked as deprecated in 2021. Please use `tufte::tufte_handout()` instead of `rmarkdown::tufte_handout()`. The latter will be removed eventually from this package.
 
+- `github_document()` gains `math_method = "default"` and defaults to it. No special processing will be done to inline maths in `$` and block maths 
+in `$$` as now [Github supports it](https://github.blog/changelog/2022-05-19-render-mathematical-expressions-in-markdown/) and will render using Mathjax (thanks, @kylebutts, #2361).
```
yihui, line 7: Removed the line-break:

```suggestion
- `github_document()` gains `math_method = "default"` and defaults to it. No special processing will be done to inline maths in `$` and block maths in `$$` as now [Github supports it](https://github.blog/changelog/2022-05-19-render-mathematical-expressions-in-markdown/) and will render using Mathjax (thanks, @kylebutts, #2361).
```

R/github_document.R
```diff
@@ -33,13 +42,16 @@
 #' ```
 #'
 #' Background or fonts color cannot be changed for now and your equation may not be visible on dark theme.
+#'
+#' **Using `webtex` will be the default with Pandoc 2.0.4 until Pandoc 2.10. Before 2.0.4, Github document output does not support math.**
```
yihui, line 46: Do you want to hard-wrap the line here?

yihui (approved):
Looks good! Thanks!

> I discovered that math feature is only available for Pandoc 2.0.4 and above. However, I silently ignore the `math_support` argument in that case. Would it be better to warn that math won't be process in that case ?

I tend to ignore Pandoc before v2.0 for these non-significant features, so I think it's okay to silently ignore the argument.

PR #2402, opened 2022-08-22
Update GHA to ubuntu-latest to avoid deprecation

cderv:
and test one more Pandoc version


cderv:
@yihui, it seems with ubuntu-latest, we are missing `pandoc-citeproc` in Pandoc installation for old pandoc. 
Do you think it is still worth testing old version before 2.7.3 ? 

I can try install it as a step for those version and report to r-lib/actions that is missing, or... 
we can just comment and not tests those versions anymore.  

They are quite old now to be honest, so I am thinking about the second solution. What is your take ?

yihui:
I agree with your choice (i.e., not testing that old version).

cderv:
Great !

PR #2416, opened 2022-10-04
Vectorise str_replace_all()

hadley:
In the development version of stringr, I'm exploring vectorising the function form of `str_replace_all()` (https://github.com/tidyverse/stringr/pull/462) because it offers considerable performance improvements. This patch quickly hacks rmarkdown to work with this change. It would probably be worthwhile to vectorise the two users of `process_html_res()` as that is likely to give a small, but meaningful, performance boost when there are many matches.

This patch will work with both CRAN and dev stringr.

### DIFF
--- a/R/base64.R
+++ b/R/base64.R
@@ -4,7 +4,6 @@ process_html_res <- function(html, reg, processor) {
   html <- one_string(html)
   process_img_src <- function(img_src) {
     src <- sub(reg, '\\1', img_src)
-
     vapply(
       seq_along(img_src),
       function(i) processor(img_src[[i]], src[[i]]),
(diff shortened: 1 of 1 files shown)


cderv (approved):

cderv:
Thank you. 

Last release of **rmarkdown** was in August so we could probably release a new version very soon if you need that for **stringr** new release. @yihui would that work ? 

When do you schedule the new **stringr** release ?

R/base64.R
```diff
@@ -4,7 +4,12 @@ process_html_res <- function(html, reg, processor) {
   html <- one_string(html)
   process_img_src <- function(img_src) {
     src <- sub(reg, '\\1', img_src)
-    processor(img_src, src)
+
```
yihui, line 7:
```suggestion
```

yihui (approved):
Yes, we can do a CRAN release this week.

hadley:
That would be great, but there's no need to rush on my behalf as this is I'm unlikely to start the stringr release process for at least two weeks.

PR #2422, opened 2022-10-14
Add render() output_file argument repetition

### LINKED ISSUE #2421: [FR] Change file name for all outputs formats, without repeating the output_file argument
Hello, I create this issue for a simple feature I miss when creating a report with multiple formats.

When I create a report in Rmarkdown, I want to output it in different formats, so I use the `rmarkdown::render()` function to change the directory and add the date to the file name.

Here is a simple header of a project:

```
---
title: "Title"
author: "Author"
output:
  html_document: default
  pdf_document: default
  word_document: default
knit: (function(inputFile, encoding) {
  rmarkdown::render(inputFile,
                    encoding = encoding,
                    output_dir = "../docs",
                    output_file = paste0("report_", Sys.Date()),
                    output_format = "all")
                    })
---
```

My issue is that only the first file has its name changed. I looked a bit in the code and this comes from the for loop which looks for the file name i in the vector of file names.
https://github.com/rstudio/rmarkdown/blob/bc3bfb88f5115298f49219e01dc29b7122e949f4/R/render.R#L299 

My current solution is to repeat the `output_file` argument: `output_file = rep(paste0("report_", Sys.Date()), 3)`, but this is not flexible if I want to change the number of outputs to have a quick preview.

The feature I'm looking for is to change the file name for all output formats, while providing only one file name.

This should be feasible if we include in R/render.R line 293:
`if (length(output_file) == 1) output_file <- rep(output_file, length(output_format))`

If this feature sounds good to you, I can add it and submit a pull request.

Best,
Mael Astruc--Le Souder

MaelAstruc:
Closes #2421

If multiple output formats are provided and only one file name is provided, this file name will be used for all formats, instead of only the first one.

Here is a short example :
```
---
title: "Title"
author: "Author"
output:
  html_document: default
  pdf_document: default
  word_document: default
knit: (function(inputFile, encoding) {
  rmarkdown::render(inputFile,
                    encoding = encoding,
                    output_file = paste0("test_", Sys.Date()),
                    output_format = "all")
                    })
---
```

Before :
Before_2022-10-14_182015

After :
After_2022-10-14_183457

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 rmarkdown 2.18
 ================================================================================
 
+- `rmarkdown::render()` argument `output_file` is now used for all output formats, if there is only one name for multiple output formats (thanks, @MaelAstruc, issue #2421).
 
 rmarkdown 2.17
 ================================================================================
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,7 @@
 rmarkdown 2.18
 ================================================================================
 
-- `rmarkdown::render()` argument `output_file` is now used for all output formats, if there is only one name for multiple output formats (thanks, @MaelAstruc, issue #2421).
+- `rmarkdown::render()` argument `output_file` is now used for all output formats, if there is only one name for multiple output formats (thanks, @MaelAstruc, #2421).
 
 rmarkdown 2.17
 ================================================================================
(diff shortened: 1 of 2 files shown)


yihui (approved):
Perfect. Thanks again!

PR #2428, opened 2022-11-24
make sure `ioslides_presentation` is properly self-contained when requested

mnazarov:
Currently if `ioslides_presentation` is used with a custom `logo` image and `self_contained = TRUE`, the logo is not properly embedded in the resulting HTML - the relative path to the logo image is used making the HTML document non-portable... This PR aims to fix that by automatically base64-encoding logo

### DIFF
--- a/R/ioslides_presentation.R
+++ b/R/ioslides_presentation.R
@@ -360,7 +360,7 @@ ioslides_presentation <- function(number_sections = FALSE,
         file.copy(from = logo, to = logo_path)
         logo_path <- normalized_relative_to(output_dir, logo_path)
       } else {
-        logo_path <- xfun::base64_uri(logo_path)
+        logo_path <- if (!grepl("^data:", logo_path)) xfun::base64_uri(logo_path) else logo_path
       }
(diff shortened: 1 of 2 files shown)


yihui (commented):
Thanks! I think this approach can easily hit the command line length limit:

- https://learn.microsoft.com/en-us/troubleshoot/windows-client/shell-experience/command-line-string-limitation
- https://serverfault.com/questions/163371/linux-command-line-character-limit

One alternative implementation is to use a random string as a placeholder for the `logo` variable, and then substitute it with the base64 data later: https://github.com/rstudio/rmarkdown/blob/d979f7d60e34a44ff1dcd676197c68a2523b152b/R/ioslides_presentation.R#L467

mnazarov:
You are right, I didn't think of that (and I was using rather small images in my tests).
I just pushed a better solution using your idea of a placeholder. Note that we can't just put a random text or non-existent filename there, since pandoc would complain - while the data url doesn't raise any questions.

yihui (approved):
Perfect. Thanks a lot!

PR #2429, opened 2022-11-25
make sure to avoid creating invalid paths when copying resources

mnazarov:
This is to address quite a specific use-case:

Running a rmarkdown document with `runtime: shiny` using a custom output format that includes some resources (e.g. logo or css) from an absolute path - which common if they are coming from a `system.file` call.

When `runtime:shiny` is used the `render` is called with explicit [`copy_resources = TRUE`](https://github.com/rstudio/rmarkdown/blob/main/R/shiny.R#L265) which then calls `copy_resources`. Currently it just appends the path to the ..._files folder:
* On *nix absolute path starts with `/` so the final url becomes e.g. `..._files//home/...` which is not too bad, but still not recommended
* On windows the path becomes  `..._files/C:/...` which is invalid


R/html_resource_copy.R
```diff
@@ -25,6 +25,12 @@ copy_html_resources <- function(html_str, lib_dir, output_dir) {
   copy_resources(html_str, lib_dir, output_dir, resource_locator)
 }
 
+make_abs_path_safe <- function(path) {
+  path <- sub("^[/]*", "", path)  # *nix
+  path <- sub("^.:(/|\\\\)*", "", path)  # windows
+  path
```
yihui, line 31:
```suggestion
  if (xfun::is_windows()) {
    sub("^.:(/|\\\\)*", "", path)  # windows
  } else {
    sub("^[/]*", "", path)  # *nix
  }
```

yihui (approved):
Thanks! Could you also add a brief news item to `NEWS.md`?

mnazarov:
thanks! sorry I missed your comment about news

yihui:
No worries!

PR #2448, opened 2023-01-20
Copy CSS files to relevant directory before `render()` happens via `run()`

### LINKED ISSUE #2443: `css` argument of `html_document` doesn't work with `runtime: shiny` + `{bslib}` `theme`
Can reproduce by doing `rmarkdown::run("index.Rmd")` with the following `index.Rmd`

```yml
---
title: "Title"
runtime: shiny
output:
  html_document:
    css: styles.css
    theme:
      version: 5
---

### Another title
```

And `styles.css`:

```css
body {background-color: purple;}
```

Which will generate the following error:

```
Warning in file(con, "r") :
  cannot open file 'styles.css': No such file or directory
```

### LINKED ISSUE #2447: `css` argument of `html_document_base` doesn't work with `scss/sass` files when `runtime: shiny`
Seemingly for the same reason as https://github.com/rstudio/rmarkdown/issues/2443 (that is, when `runtime: shiny`, `html_document_base()` gets invoked in a temporary directory)

````yaml
---
title: "Title"
runtime: shiny
output:
  html_document:
    css: styles.scss
    theme: darkly
---

### Another title

````

```scss
body {background-color: purple;}
```

Which will generate the following error:

```
Error in sass::sass_file: Could not find file: 'styles.scss' in dir: 
  /private/var/folders/t8/jxw4bftj35g15fcb6m4mfbc80000gn/T/Rtmpg4LF0x
```

cpsievert:
Closes #2443
Closes #2447

I'll admit, I'm not confident this is the best way to address these issues, but it's the best one I could think of so far, and would like to know if this seems like a reasonable direction.

I'm also not quite sure why I need to `file.copy()` to `dirname(dirname(output_dest))`  to get this to work (shouldn't it be `dirname(output_dest)`?)

### DIFF
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -259,6 +259,17 @@ rmarkdown_shiny_server <- function(dir, file, auto_reload, render_args) {
         list()
       }
 
+      # The input Rmd may have CSS file paths that are relative to its directory,
+      # so copy them over to the directory that html_document_base() will actually
+      # execute in (since html_document_base() will try to read them before
+      # copy_resources happens)
+      front_matter <- yaml_front_matter(reactive_file())
+      css_files <- unlist(lapply(front_matter$output, `[[`, "css"))
+      file.copy(
+        file.path(dirname(reactive_file()), css_files),
+        dirname(dirname(output_dest))
+      )
+
       # ensure that the document is not rendered to one page
       output_opts <- list(
         self_contained = FALSE,


cderv (commented):
Thanks @cpsievert. This PR is a solution for shiny only, but the issue is more deep in the code base. This is a problem with resource discovery when some intermediate_dir are used. (which is the case of shiny). 

See #2486 for a PR with explanation - I'll close this one if you don't mind

Issue #2347, opened 2022-04-11
[FR] Add a new type of dependency that bundles lua filter and other type of dependencies


cderv:
More general thought: 

I think I have a wish in mind I should start working on at some point: a way to have a _feature extensions bundle_ where we have a  `htmlDependency` with a Lua filter requirements. Something that we could write in a wrapped way, and that **rmarkdown** could unwrapped and use the correct way. We don't have a `lua_dependency` component, and maybe that would make sense to have at some point adding more feature like this 
````r
lua_dependency(
	filters = "code-folding-lua",
    html_deps = htmlDependency(
    	"codefolding-lua",
	    version = "1.1",
    	src = pkg_file("rmd/h/navigation-1.1"),
	    stylesheet = "codefolding-lua.css",
    	all_files = FALSE
  )
}
````

We could then group some feature. Just an idea...

_Originally posted by @cderv in https://github.com/rstudio/rmarkdown/pull/2340#discussion_r841559258_

**Comment by yihui** [comment]
That sounds like a great idea to me!

**Comment by atusy** [comment]
I'm currently working on this.

Some thoughs:

To make things simple yet powerful, I would add `pandoc_dependency` rather  than `lua_dependency`.
Then, we can also control metadata and variables.
I also guess there are some usecases where to add metadata or variables but no lua filters.
In such cases, the name `lua_dependency` is not adequate.

``` r
pandoc_dependency(
  lua_filters,
  metadata,
  variables,
  ... # can be more?
)
```

Of course, we can still add `rmd_dependency` which allows declare variety of dependencies at once.
I think we do not need `_deps` suffix as it is obvious from the function name.

``` r
rmd_dependency(
  pandoc # pandoc_dependency
  html # htmlDependency
  latex # latex_dependency
)
```

cderv:
> I'm currently working on this.

Cool ! Thank you ! 

> To make things simple yet powerful, I would add pandoc_dependency rather than lua_dependency.

Really interesting ! Yes I like that idea. Let's try with that !

Issue #2314, opened 2022-02-18
Allow `output_file` to be a function in `render()` [FR]


mikmart:
It would be useful to have an easy way to reuse metadata included in the YAML front matter of an RMarkdown file for generating output file names when rendering. This originally [came up on StackOverflow](https://stackoverflow.com/q/71173343/4550695). 

At the moment it's simple enough to do with `yaml_front_matter()` if the metadata doesn't include R expressions. Just extract the metadata from the RMarkdown file, and build the output file name.

However, it becomes quite a bit more complicated if the metadata fields contain expressions that need to be rendered first. One way to do that is to keep the intermediate Markdown file, and extract the metadata from there. For example:

``` md
---
title: "Untitled"
author: "Jane Doe"
date: "`r Sys.Date()`"
output: word_document
knit: >
  (function(input_file, encoding) {
    # Render, keeping intermediate files for extracting front matter
    md_dir <- tempdir()
    output_file_temp <- rmarkdown::render(
      input = input_file,
      output_file = tempfile(),
      intermediates_dir = md_dir,
      clean = FALSE
    )
    
    # Get the rendered front matter from the intermediate Markdown file
    md_file <- fs::path_ext_set(fs::path_file(input_file), ".knit.md")
    metadata <- rmarkdown::yaml_front_matter(fs::path(md_dir, md_file))
    
    # Build output file name based on rendered metadata
    output_name <- with(metadata, paste(title, "by", author, "on", date))
    output_ext <- fs::path_ext(output_file_temp)
    output_file <- fs::path_ext_set(output_name, output_ext)
    
    fs::file_move(output_file_temp, output_file)
  })
---

204 No Content
```

Would it be possible to allow the `output_file` argument to `render()` to be a function that gets the rendered metadata as an argument? The above could then become:

``` md
---
title: "Untitled"
author: "Jane Doe"
date: "`r Sys.Date()`"
output: word_document
knit: >
  (function(input_file, encoding) {
    rmarkdown::render(
      input = input_file,
      output_file = function(metadata) {
        with(metadata, paste(title, "by", author, "on", date))
      }
    )
  })
---

204 No Content
```

And the resulting output file would be named `Untitled by Jane Doe on 2022-02-18.docx`.

atusy:
I would suggest define custom format and tweak `post_processor` with `rmarkdown::output_format`.
In this way, you do not have to write `knit` in YAML front matter.

yihui:
[...]

PR #2462, opened 2023-03-09
Add output_format_dependency()

atusy:
This PR is an attempt to solve #2347, but in a more generalized way.

In the issue #2347, I suggested `pandoc_dependency()`, however, this function needs to follow the updates of Pandoc's API changes.

Instead, this PR implements `output_format_dependency()`, which allows defining some of the arguments in `output_format` from within a Rmd file.
Supported arguments are `pre_processor`, `post_processor`, `file_scope`, `on_exit`.
The others are not because dependency can be solved after `post_knit`, which finalizes `knit_meta`.

Fortunately, `pre_processor` can return extra arguments to pass to pandoc, so we can dynamically add lua filters, which covers the motivation of #2347.

```r
output_format_dependency(
  name = "example",
  pre_processor = function(...) {
    pandoc_lua_filter_args(c("example1.lua", "example2.lua"))
  },
  post_processor = function(...) {},
)
```

Also, this PR should be helpful in the situations like #2314.
In #2314, the issue author suggests `knit` filed in the YAML front matter to modify the `render()` function.
This is a powerful solution, but limited to RStudio users...

My proposal is available for R users on any environments.

Note that there is a remaining work to implement `attach_output_format_dependencies()`, like `htmltools::attachDependencies()`, but this can be a future work.


atusy:
An example Rmd

<details>

````
---
title: foo
output:
  html_document:
    keep_md: true
---

```{cat, engine.opts=list(file = "example.lua")}
print("====lua filter====")
```

```{r}
rmarkdown::output_format_dependency(
  name = "foo",
  pre_processor = function(...) {
    cat("====pre_processor====\n")
    pandoc_lua_filter_args("example.lua")
  },
  post_processor = function(metadata, input_file, output_file, ...) {
    cat("====post_processor====\n")
    output_file
  }
)
```
````
</details>

**Comment by cderv** [comment]
Wow this is an interesting idea - I need to look into that closer so it could be not right now as I am currently focus on Quarto 1.3 release. But as soon as I can make time for some rmarkdown work (which is completely due and should happen soon), I'll give this a try. Maybe @yihui will have more time before. 

Anyhow, generalized approach needs more thoughts than simple PR. Your contribution is really appreciated and I wanted to be honest with you with current situation so that you don't think this PR is ignored! Thank you very much for this @atusy !

**Comment by atusy** [comment]
No worries! I'm not hurry.

File: R/output_format.R [hunk 1]
```diff
@@ -815,3 +815,56 @@ citeproc_required <- function(yaml_front_matter,
...
+
+#' @export
+knit_print.output_format_dependency <- function(x, ...) {
+  knitr::asis_output(list(), meta = list(x))
+}
+
+merge_output_format_dependency <- function(fmt, dep) {
+  dep$name <- NULL
+  do.call(output_format,
+          c(dep, list(knitr = NULL, pandoc = list(), base_format = fmt)))
```
cderv, line 858:
```suggestion
          c(dep, list(knitr = list(), pandoc = list(), base_format = fmt)))
```
would this work too ?

cderv (commented):
That is really clever way to generalize that ! Thank you.  

Should we completely make all the `output_format` argument available that way ? it seems something possible. 

At least, I would like to see if we can add a specific handling for some of the pandoc component. `args` and `lua_filters` are among the content one would want to modify 

I wonder if we shouldn't extend this also to the knitr and pandoc component currently set as `knitr = NULL` and `pandoc = list()`

This would mean being possible to do something like 
```r
output_format_dependency <- function(name,
                                     pre_processor = NULL,
                                     post_processor = NULL,
                                     file_scope = NULL,
                                     on_exit = NULL,
                                     knitr = list(),
                                     pandoc = list())
```

so that one can do 
```r
output_format_dependency(pandoc = list(lua_filters = "example.lua"))
```

Did you try making that works ? 

Obviously we need to add some unit test for this so that we are sure it merges as we expect, and we don't break it in the future. 

Otherwise, all good ! Really exciting work ! Thanks again !

atusy:
@cderv

Thank you so much for your review! I am excited by your reactions!

I will try adding the pandoc argument and unit tests in this or next week.

> I wonder if we shouldn't extend this also to the knitr and pandoc component currently set as knitr = NULL and pandoc = list()

About the above comment,

- `knitr` is not supproted because
    - the timing of solving `output_format_dependency` is generally too late because `merge_output_format_dependencies()` is called **after knit()**.
    - users/developers already have full powers on controlling the options within chunks by setters and getters (e.g., `opts_knit$set()`)
    - Calling setters within `output_format_dependency()` can be a solution, but the timing of the effect can be ambious. Specifically, there is no way to controll chunk options of the **current** chunk https://github.com/yihui/knitr/issues/1798 .

- `pandoc` seems possible. Although I wrote it is not supported for some reason, maybe I was wrong here :smile: 

    > Supported arguments are pre_processor, post_processor, file_scope, on_exit.
    > The others are not because dependency can be solved after post_knit, which finalizes knit_meta.

cderv:
> the timing of solving output_format_dependency is generally too late because merge_output_format_dependencies() is called after knit().

Oh indeed. I missed that. I was too exited by the new feature probably ! 😄 

> The others are not because dependency can be solved after post_knit, which finalizes knit_meta.

Interesting 🤔  - initial issue was to have a way to add Lua filter more easily as a dependency. I thought we could somehow add to the `pandoc = list(lua_filters = )` . Maybe this is the only element of Pandoc that we can use 

Otherwise, it is easy enough to do a wrapper of this new function, as lua_filters can be added  in pre_processor as  you did. But I did it another for some reason (I think to add a format lua filter last or something like that). 

I let you look into this and we'll see if we can change things to make it happen.

atusy:
Okay, I confirmed `pandoc = list(lua_filters = ...)` works.
Here's an example Rmd that a filter echoes `====lua filter====` to stdout.

<details><summary>Rmd</summary>

````
---
title: foo
output:
  html_document:
    keep_md: true
---

```{cat, engine.opts=list(file = "example.lua")}
print("====lua filter====")
```

```{r}
rmarkdown::output_format_dependency(
  name = "foo",
  pandoc = list(
    lua_filters = "example.lua"
  ),
  pre_processor = function(...) {
    cat("====pre_processor====\n")
    # pandoc_lua_filter_args("example.lua")
  },
  post_processor = function(metadata, input_file, output_file, ...) {
    cat("====post_processor====\n")
    output_file
  }
)
```
````

</details>

<details><summary>stdout from render()</summary>
```
processing file: example.Rmd
                                                                                                                                                                                                                                                                                                                                                                                            
                            
output file: example.knit.md

====pre_processor====
/usr/bin/pandoc +RTS -K512m -RTS example.knit.md --to html4 --from markdown+autolink_bare_uris+tex_math_single_backslash --output example.html --lua-filter /home/atusy/ghq/github.com/rstudio/rmarkdown/inst/rmarkdown/lua/pagebreak.lua --lua-filter /home/atusy/ghq/github.com/rstudio/rmarkdown/inst/rmarkdown/lua/latex-div.lua --lua-filter example.lua --embed-resources --standalone
 --variable bs3=TRUE --section-divs --template /home/atusy/ghq/github.com/rstudio/rmarkdown/inst/rmd/h/default.html --no-highlight --variable highlightjs=1 --variable theme=bootstrap --mathjax --variable 'mathjax-url=https://mathjax.rstudio.com/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML' --include-in-header /tmp/RtmpxmaQr6/rmarkdown-strb7dc26a6dd7a3.html 
====lua filter====
====post_processor====

Output created: example.html
```

</details>

So the last task is to add tests.

**Comment by atusy** [comment]
To implement tests, I rebased my PR onto #2488 .
I will rebase again to main when #2488 is merged.

File: R/render.R [hunk 2]
```diff
@@ -840,6 +840,8 @@ render <- function(input,
 
     perf_timer_start("pre-processor")
 
+    output_format <- merge_output_format_dependencies(output_format, knit_meta)
+
```
cderv, line 844: I think we should make this conditional. This is safer than just relying on `merge_output_format_dependencies` to do nothing IMO. More clear also from `render.R` file 

```suggestion
    if (has_dependencies(knit_meta, "output_format_dependency")) {
        output_format <- merge_output_format_dependencies(output_format, knit_meta)
    }
```

What do you think ?

cderv (commented):
All good. You can rebase on main and maybe make the suggested change, and all good to merge ! 

Thank you

R/render.R
cderv: Yes I understand why you feel that. But thank you :)

cderv (approved):
Awesome ! 

Need to be AFK but I'll rebase later if you haven't done it and then I'll merge 

Thank you

atusy:
rebased and added news.

yihui:
@atusy Do you have time to write a short blog post on this new feature? One or two examples would be enough. I feel many people will be interested in this feature. Feel free to publish the post in your own blog (https://blog.atusy.net), and we can help promote it. Thanks!

atusy:
Yeah sure. I would post one in the next week!

atusy:
done https://blog.atusy.net/2023/08/28/rmarkdown-output-format-dependency/

Issue #2405, opened 2022-08-25
Update `jquery-ui` to latest version to mitigate vulnerabilities


mahesh2013:
Location : rmarkdown/rmd/h/jqueryui/jquery-ui.js
vulnerabilities:
https://nvd.nist.gov/vuln/detail/CVE-2016-7103 
https://nvd.nist.gov/vuln/detail/CVE-2021-41182  
https://nvd.nist.gov/vuln/detail/CVE-2021-41183  
https://nvd.nist.gov/vuln/detail/CVE-2021-41184   
https://nvd.nist.gov/vuln/detail/CVE-2022-31160  

jquery-ui version fixes these issue: 1.13.2
https://www.npmjs.com/package/jquery-ui 

```
> xfun::session_info('rmarkdown')
R version 4.2.0 (2022-04-22)
Platform: x86_64-conda-linux-gnu (64-bit)
Running under: Red Hat Enterprise Linux 8.6 (Ootpa), RStudio 2022.7.1.554

Locale:
  LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8       
  LC_COLLATE=en_US.UTF-8     LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
  LC_PAPER=en_US.UTF-8       LC_NAME=C                  LC_ADDRESS=C              
  LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

Package version:
  base64enc_0.1.3 digest_0.6.29   evaluate_0.16   fastmap_1.1.0   glue_1.6.2     
  graphics_4.2.0  grDevices_4.2.0 highr_0.9       htmltools_0.5.3 jquerylib_0.1.4
  jsonlite_1.8.0  knitr_1.39      magrittr_2.0.3  methods_4.2.0   rlang_1.0.4    
  rmarkdown_2.11  stats_4.2.0     stringi_1.7.8   stringr_1.4.0   tinytex_0.35   
  tools_4.2.0     utils_4.2.0     xfun_0.32       yaml_2.3.5     

Pandoc version: 2.18
```

cderv:
Thanks. 

We can probably download latest stable version from https://jqueryui.com/download/

@yihui FWIW shiny uses already one version above the one in **rmarkdown** 
https://github.com/rstudio/shiny/tree/main/inst/www/shared/jqueryui

[...]

PR #2477, opened 2023-04-21
fix: bump jquery-ui to v1.13.2 to fix multiple CVEs

daschnerm:
## What

- bump jquery-ui to v1.13.2 to fix CVE-2016-7103
- I've tested the floating ToC, as well as the normal ToC feature still works without any issues.

## Why

https://github.com/rstudio/rmarkdown/issues/2405


inst/rmd/h/jqueryui/jquery-ui.css
```diff
@@ -1,9 +1,13 @@
...
+* Includes: draggable.css, core.css, resizable.css, selectable.css, sortable.css, accordion.css, autocomplete.css, menu.css, button.css, controlgroup.css, checkboxradio.css, dialog.css, progressbar.css, selectmenu.css, slider.css, spinner.css, tabs.css, tooltip.css, theme.css
```
cderv, line 3: Do you think we can just keep the component we had ? 

It seems that this adds controlgroup, checkboxradio . Are they needed ¿

cderv (commented):
Thanks a lot. 

I left a comment. I wonder if we should have exactly the same component. 

Otherwise, I was wondering if we should sync with the version in shiny. What do you think ?  
I did compare with this script executed on your branch 
```r
# This script aims to update and sync jquery-ui dependency with the one in shiny

rmd <- "inst/rmd/h/jqueryui/"

shiny <- "https://github.com/rstudio/shiny/archive/refs/heads/main.zip"

temp_zip <- tempfile(fileext = ".zip")
xfun::download_file(shiny, temp_zip, mode = "wb")

dir.create(temp_shiny <- tempfile("shiny"))

unzip(temp_zip, exdir = temp_shiny)

jqueryui <- file.path(temp_shiny, "shiny-main", "inst", "www", "shared", "jqueryui")

unlink(rmd, recursive = TRUE)

file.copy(jqueryui, dirname(rmd), overwrite = TRUE, recursive = TRUE)

unlink(temp_shiny, recursive = TRUE)
unlink(temp_zip, recursive = TRUE)
````

It seems there are some differences probably due to the  components includes.

File: inst/rmd/h/jqueryui/jquery-ui.css (hunk 1 above)
Comment (daschnerm) [reply inst/rmd/h/jqueryui/jquery-ui.css:3]: I don't think that they are needed, but I didn't check the code. However the jquery-ui from Shiny includes both of them:

```
diff shiny-css.txt rmarkdown-main-css.txt 
4,5d3
< checkboxradio.css
< controlgroup.css
7d4
< datepicker.css
```

If we want to sync with shiny, then I guess we should use the same components ?

daschnerm:
@cderv I've updated the PR and pulled in the`jquery-ui` version from Shiny.

cderv (approved):

PR #2481, opened 2023-04-28
[ioslide] Only do the chcp setting with Pandoc before 2.0

### LINKED ISSUE #134: ioslides_presentation "Invalid UTF-8 stream" error with &nbsp;
I'm using `render` to create an ioslides presentation with a table generated from `pandoc.table` and am running into the following error: 

*pandoc.exe: Cannot decode byte '\xa0': Data.Text.Encoding.Fusion.streamUtf8: Invalid UTF-8 stream*

The error appears to be caused by `pandoc.table` returning `&nbsp;` for the first column header, as removing `&nbsp;` eliminates the error.  Below is an Rmd file that can be used to reproduce the error message.  The same error message occurs when using the *Knit HTML* button in RStudio.

    ---
    title: "Untitled"
    output: ioslides_presentation
    ---
    
    ## Slide
    
    ```{r, echo=FALSE, results='asis'}
    library(pander)
       
    m <- mtcars[1:2, 1:3]
    
    # error
    pandoc.table(m, style = "rmarkdown")
    
    # success
    ptable = function(...) {
      cat(paste0(gsub("&nbsp;", "      ",
                      pander.return(...)), collapse = "\n"))
    }
    # ptable(pandoc.table(m, style = "rmarkdown"))
    ```

### LINKED ISSUE #2480: Error Knit ioslides_presentation (R 4.3)
Hello,

I upgraded to the latest version of R to 4.3 and now when I Knit an Rmarkdown file where the output is ioslides_presentation the following error occurs. I tried using the default ioslides_presentation template and the error also happens.

```
Error in gsub("\\D", "", system2("chcp", stdout = TRUE)) : 
  input string 1 is invalid
Calls: <Anonymous> -> <Anonymous> -> overlay -> gsub
Además: Warning message:
In gsub("\\D", "", system2("chcp", stdout = TRUE)) :
  unable to translate 'P<a0>gina de c<a2>digos activa: 850' to a wide string
Ejecución interrumpida
```

```
R version 4.3.0 (2023-04-21 ucrt)
Running under: Windows 10 x64 (build 19045)

Matrix products: default


locale:
[1] LC_COLLATE=Spanish_Colombia.utf8 
[2] LC_CTYPE=Spanish_Colombia.utf8   
[3] LC_MONETARY=Spanish_Colombia.utf8
[4] LC_NUMERIC=C                     
[5] LC_TIME=Spanish_Colombia.utf8    

time zone: America/Bogota
tzcode source: internal

attached base packages:
[1] stats     graphics  grDevices utils    
[5] datasets  methods   base     

 [1] compiler_4.3.0  fastmap_1.1.1  
 [3] cli_3.6.1       htmltools_0.5.5
 [5] tools_4.3.0     rstudioapi_0.14
 [7] yaml_2.3.7      rmarkdown_2.21 
 [9] knitr_1.42      xfun_0.39      
[11] digest_0.6.31   rlang_1.1.0    
[13] evaluate_0.20  

```

Regards,

cderv:
To fix #2480 

This usage of chcp was introduced in 955ebae74a6797f13953bee5d586b83b84a0dada to fix #134 - quite old code.

It was pre pandoc 2.0 and accompanied with an issue upstream https://github.com/jgm/pandoc/issues/2101

I believe pandoc now handles directly this better and this is not needed anymore. I suggest to not do that anymore when Pandoc 2 or above. 

@yihui any argument against this ? 

I could also try to fix the proper issue, but I believe this is something related to encoding between what system call returns and the OS or R encoding. 

EDIT: I investigated at https://github.com/rstudio/rmarkdown/issues/2480#issuecomment-1527533282 and it seems this is R 4.3 big change on Windows

### DIFF
--- a/R/ioslides_presentation.R
+++ b/R/ioslides_presentation.R
@@ -428,7 +428,7 @@ ioslides_presentation <- function(number_sections = FALSE,
     # duration of the Pandoc command. Without this, Pandoc fails when attempting
     # to hand UTF-8 encoded non-ASCII characters over to the custom Lua writer.
     # See https://github.com/rstudio/rmarkdown/issues/134
-    if (is_windows()) {
+    if (is_windows() && !pandoc2.0()) {
       # 'chcp' returns e.g., "Active code page: 437"; strip characters and parse
       # the number
       codepage <- as.numeric(gsub("\\D", "", system2("chcp", stdout = TRUE)))


yihui (approved):
Sounds good! I'm always for getting rid of old hacks.

PR #2488, opened 2023-05-26
add missing functionality to merge file_scope on creating output_format

atusy:
Currently, new output format ignores the file scope of the base format.
It is totally lost.

In this PR I fix the above behavior and add `merge_file_scope()`.

To achieve merging, `file_scope` function needs two arguments, and the latter is new

- `input_files` which already exists
- `current_file_scope` which is determined by the base `file_scope` function

In this way, overlay can take into account of file scope determined by the base.

This is an API change, so I added some warnings and fallback mechanisms.

### DIFF
--- a/R/output_format.R
+++ b/R/output_format.R
@@ -52,9 +52,11 @@
 #' @param file_scope A function that will split markdown input to pandoc into
 #'   multiple named files. This is useful when the caller has concatenated a set
 #'   of Rmd files together (as \pkg{bookdown} does), and those files may need to
-#'   processed by pandoc using the \code{--file-scope} option. The function
-#'   should return a named list of files w/ \code{name} and \code{content} for
-#'   each file.
+#'   processed by pandoc using the \code{--file-scope} option. The first
+#'   argument is input file paths and the second is \code{NULL} or current file
+#'   scope which is a named list of files w/ \code{name} and \code{content} for
+#'   each file. The return is the new file scope. Also, the arguments should
+#'   include \code{...} for the future extensions.
 #' @param base_format An optional format to extend.
 #' @return An R Markdown output format definition that can be passed to
 #'   \code{\link{render}}.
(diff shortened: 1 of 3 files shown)


atusy:
Note that I noticed this when adding tests on #2462, which should be able to extend file_scope from within chunks.

yihui (approved):
Looks good to me. Thanks!

cderv (commented):
Sorry for the delay on this @atusy. 

it looks good to me. I think it won't break current behavior. Only place that we use it in in **bookdown**. It is even deactivated by default, and could be completely broken in fact ... :( 

I am just curious:  do we have a use case where we think merging `file_scope` requires applying `overlay` on `base` results ?  

To me this would mean that we want to define a format that would for example  extend `gitbook()` but would split again each of the splitted file, instead of providing an overwrite of the split function. Is that right ? 

Just trying to understand correctly. I guess someone can do a function that ignores `current_scope` to completely ignore the base splitted content.

I was thinking maybe the latter would be the desired behavior when extending the base format. So I was curious if this change was driven by a use case. Thanks. 

Other than this question I have, the implementation looks very good ! thanks a lot !

cderv (approved):
In fact it is all good. Both usage can be done. Let's go with this default. Thanks a lot !

atusy:
Actually I'm not sure if there would be any practical usecases...

One exception is test.
We can easily wrap testing output format and test if `file_scope` is working as expected.

```` r
test_that("file_scope replaces base format", {
  out_fmt <- output_format(
    knitr = NULL, pandoc = NULL, keep_md = NULL, clean_supporting = NULL,
    file_scope = function(input_file, file_scope, ...) {
      # do some tests here!
      testthat::expect_snapshot(cat(file_scope))
      return(file_scope)
    },
    base_format = bookdown::git_book
  )
  rmarkdown::render("example.Rmd", output_format = out_fmt)
})
````

cderv:
Good use case ! I had not thought of that 😆

PR #2489, opened 2023-05-26
Change the code-folding button text from "Code" to "Show"

yihui:
The content to show is not necessarily code, e.g., https://github.com/yihui/knitr/issues/2227


cderv (approved):
Yes it makes sense. This is more clear. 

FWIW Quarto mix both : Show Code / Hide Code 😅 
https://quarto.org/docs/output-formats/html-code.html#code-tools

I think this is a safe change, I don't think it will break anything.

PR #2507, opened 2023-08-31
feat: allow package developers to attach output_format_dependency from arbitrary functions

atusy:
This PR implements `attach_output_format_dependency` that internally calls `knitr::knit_meta_add` so that the dependency is attached explicitly compared to the attachment via `knit_print.output_format_dependency`

With this, package authors can implement functions whose return values rely on the output format dependency.

For example, the rmd below implements a section refference.

<details><summary>rmd</summary>

```` rmd
---
title: referring sections
output:
  html_document:
    keep_md: true
---

```{cat, engine.opts=list(file = "example.lua")}
local headers = {}

local function collect_header(header)
  headers["#" .. el.identifier] = header
end

local function link_header(link)
  if #link.content == 0 and headers[link.target] then
    link.content = headers[link.target].content
  end
  return link
end

return {
  { Header = collect_header },
  { Link = link_header },
}
```

```{r}
ref <- function(id) {
  dep <- rmarkdown::output_format_dependency(
    name = "example",
    pandoc = list(lua_filters = "example.lua"),
  )
  rmarkdown:::attach_output_format_dependency(dep)
  return(sprintf("[](#%s)", id))
}
```

# awesome section {#awesome} 

see `r ref("brilliant")`

# brilliant section {#brilliant}

see `r ref("awesome")`
````

</details>


cderv:
Just curious : why not just rely on `knitr::knit_meta_add()` in that case ? 
I believe other packages do that for attaching HTML dependencies (like **learnr** for example)

Is there really an advantage to have a dedicated function in **rmarkdown** ? I don't think there is one for other deps. 

Maybe I am missing something (possibly related to the `label` argument - I don't think I specify the `label` usually as specific chunk does not matter often)

atusy:
hmm, sounds right...
maybe I'm doing too much.

Thanks for the comment. I close the PR.

PR #2530, opened 2023-12-05
Remove `stringr` dependency

etiennebacher:
The aim of this PR is to remove the `stringr` dependency. `stringr` is only used once in the package but adds 4 (sometimes heavy) dependencies: `stringi`, `vctrs`, `magrittr`, and `stringr` itself.

```r
pak::pkg_deps("rmarkdown", dependencies = "hard") |> 
    dplyr::pull(ref) |> 
    unique() |> 
    length()
#> 31
```
After this PR:
```r
pak::local_deps(".", dependencies = "hard") |> 
    dplyr::pull(ref) |> 
    unique() |> 
    length()
#> 27
```
I didn't test on external packages, I only relied on the test suite here.

### DIFF
--- a/R/base64.R
+++ b/R/base64.R
@@ -10,15 +10,9 @@ process_html_res <- function(html, reg, processor) {
       character(1)
     )
   }
+  m <- gregexpr(reg, html, perl = TRUE)
+  regmatches(html, m) <- lapply(regmatches(html, m), process_img_src)
 
-  # base R alternative to stringr::str_replace_all() with a function in
-  # replacement
-  mtchs <- regmatches(html, gregexpr(reg, html, perl = TRUE))[[1]]
-  if (length(mtchs) > 0) {
-    for (i in rev(mtchs)) {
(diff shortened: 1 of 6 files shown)


R/base64.R
```diff
@@ -10,7 +10,15 @@ process_html_res <- function(html, reg, processor) {
...
-  html <- stringr::str_replace_all(html, reg, process_img_src)
+
+  # base R alternative to stringr::str_replace_all() with a function in
+  # replacement
+  mtchs <- regmatches(html, gregexpr(reg, html, perl = TRUE))[[1]]
+  if (length(mtchs) > 0) {
+    for (i in rev(mtchs)) {
+      html <- gsub(i, process_img_src(i), html, fixed = TRUE)
+    }
+  }
```
yihui, line 21: You can use `regmatches<-` here to process the matches (I'll do it by myself in a minute).

yihui (approved):
Great! Thanks!

R/base64.R
etiennebacher: Thanks for the tip!

PR #2532, opened 2023-12-11
Use new version of `has_crop_tools()` from knitr

cderv:
Context: 

* https://github.com/yihui/knitr/issues/2246
* https://github.com/yihui/knitr/commit/4953bec489479bcef6eeb2f52cb9be40daabe47b

We improved the logic in **knitr** but never ported to **rmarkdown**. 

I use conditional to use internal **knitr** function. 

* Should it be exported from **knitr** instead ? 
* Should we inline instead to replace the one in **rmarkdown** completely ? 
* Or is it old enough so that we bump **knitr** min requirement ?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,8 @@ rmarkdown 2.26
 
 - Removed the **stringr** dependency since it is used only once in the package and the equivalent base R code is simple enough (thanks, @etiennebacher, #2530).
 
+- When `fig_crop: auto`, it will now uses the same logic as in knitr to decide if cropping is possible (yihui/knitr#2246). This fix an issue with detecting Ghostscript on Windows. It requires **knitr** 1.43 for this fix to apply.
+
 
 rmarkdown 2.25
 ================================================================================
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,11 +1,13 @@
 rmarkdown 2.26
 ================================================================================
 
+- **rmarkdown** has now a dependency on **knitr** >= 1.43.
+
 - Get rid of the superfluous warning in `find_pandoc()` (thanks, @jszhao, #2527).
 
 - Removed the **stringr** dependency since it is used only once in the package and the equivalent base R code is simple enough (thanks, @etiennebacher, #2530).
 
-- When `fig_crop: auto`, it will now uses the same logic as in knitr to decide if cropping is possible (yihui/knitr#2246). This fix an issue with detecting Ghostscript on Windows. It requires **knitr** 1.43 for this fix to apply.
+- When `fig_crop: auto`, it will now uses the same logic as in knitr to decide if cropping is possible (yihui/knitr#2246).
 
 
 rmarkdown 2.25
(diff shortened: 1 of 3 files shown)


R/util.R
```diff
@@ -316,6 +316,9 @@ find_program <- function(program) {
 }
 
 has_crop_tools <- function(warn = TRUE) {
+  if (packageVersion("knitr")>= "1.43")) return(knitr:::has_crop_tools(warn))
```
yihui, line 319:
```suggestion
  if (packageVersion("knitr")>= "1.43") return(knitr:::has_crop_tools(warn))
```

yihui (commented):
> * Should it be exported from **knitr** instead ?
> * Should we inline instead to replace the one in **rmarkdown** completely ?
> * Or is it old enough so that we bump **knitr** min requirement ?

I think it's fine to use `:::` (CRAN allows for it when the maintainer is the same for the two packages) and not export it from **knitr**.

I'm also okay with requiring `knitr >= 1.43`. In theory, when users install the new version of **rmarkdown**, **knitr** should be updated automatically if we specify this version requirement. In practice, I did see (rare) cases in which the version requirement was bypassed, but I have never understood how it was possible.

cderv:
Ok I went with the dependency upgrade. 

I did not use the function directly. Not sure if any of our others tools is using `rmarkdown:::had_crop_tools()` so I went with just using knitr internal in it. 

We could have re-export but don't think that works with internal function as I would have done
````r
#' @importFrom knitr has_crop_tools
#' @export
knitr::has_crop_tools
````

File: NEWS.md [hunk 2]
```diff
@@ -1,10 +1,14 @@
 rmarkdown 2.26
 ================================================================================
 
+- **rmarkdown** has now a dependency on **knitr** >= 1.43.
```
```suggestion
- **rmarkdown** now requires **knitr** >= 1.43.
```

NEWS.md
```diff
@@ -1,10 +1,14 @@
...
+
 - Get rid of the superfluous warning in `find_pandoc()` (thanks, @jszhao, #2527).
 
 - Removed the **stringr** dependency since it is used only once in the package and the equivalent base R code is simple enough (thanks, @etiennebacher, #2530).
 
+- When `fig_crop: auto`, it will now uses the same logic as in knitr to decide if cropping is possible (yihui/knitr#2246).
```
yihui, line 10:
```suggestion
- For the output format option `fig_crop: auto`, it will now use the same logic as in **knitr** to decide if cropping is possible (yihui/knitr#2246).
```

R/util.R
```diff
@@ -316,18 +316,7 @@ find_program <- function(program) {
...
-  )
-  missing <- tools[tools == ""]
-  if (length(missing) == 0) return(TRUE)
-  x <- paste0(names(missing), collapse = ", ")
-  if (warn) warning(
-    sprintf("\nTool(s) not installed or not in PATH: %s", x),
-    "\n-> As a result, figure cropping will be disabled."
-  )
-  FALSE
+  return(knitr:::has_crop_tools(warn))
```
yihui, line 319:
```suggestion
 knitr:::has_crop_tools(warn)
```

yihui (approved):
Looks good. Thanks!

Issue #2534, opened 2024-01-16
`rmarkdown::render()` corrupts input file


J-Moravec:
## Issue:

Made a typo and instead of `rmarkdown::render("report.rmd")` wrote `rmarkdown::render("report.rds")`. To my surprise, `rmarkdown` finished without error, but the `rds` file was mangled. This shouldn't happen.

### Why this is an issue:

Somewhere in the pipeline, the input file is being modified. This means that potentially, the call of `rmarkdown::render(input)` can corrupt the `input` file. This breaks the assumption of the call being safe.

### MRE:

```r
saveRDS("test", "test.rds")
rmarkdown::render("test.rds") # no error
readRDS("test.rds") # error
```

Reproduced on both the latest cran and the development version of `rmarkdown`.

Output of `xfun::session_info("rmarkdown")`
```
R version 4.1.2 (2021-11-01)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 22.04.3 LTS

Locale:
  LC_CTYPE=en_NZ.UTF-8       LC_NUMERIC=C              
  LC_TIME=en_NZ.UTF-8        LC_COLLATE=en_NZ.UTF-8    
  LC_MONETARY=en_NZ.UTF-8    LC_MESSAGES=en_NZ.UTF-8   
  LC_PAPER=en_NZ.UTF-8       LC_NAME=C                 
  LC_ADDRESS=C               LC_TELEPHONE=C            
  LC_MEASUREMENT=en_NZ.UTF-8 LC_IDENTIFICATION=C       

Package version:
  base64enc_0.1.3   bslib_0.6.1       cachem_1.0.8      cli_3.6.2        
  digest_0.6.34     ellipsis_0.3.2    evaluate_0.23     fastmap_1.1.1    
  fontawesome_0.5.2 fs_1.6.3          glue_1.7.0        graphics_4.1.2   
  grDevices_4.1.2   highr_0.10        htmltools_0.5.7   jquerylib_0.1.4  
  jsonlite_1.8.8    knitr_1.45        lifecycle_1.0.4   memoise_2.0.1    
  methods_4.1.2     mime_0.12         R6_2.5.1          rappdirs_0.3.3   
  rlang_1.1.3       rmarkdown_2.25.2  sass_0.4.8        stats_4.1.2      
  tinytex_0.49      tools_4.1.2       utils_4.1.2       xfun_0.41        
  yaml_2.3.8       

[...]

PR #2536, opened 2024-01-17
unit-test for #2534

J-Moravec:
Unit-test for this particular user-case.

Running `pandoc` is unfortunately necessary.

Tested on CRAN version of `rmarkdown` and a dev version using this fork.

In case of CRAN version, `pandoc` also likes to throw some additional warnings and errors that cannot be suppressed (or at least I don't know how).

If a `withr::with_tempfile()` is required instead of `on.exit()`, I can rework this. Since `on.exit` is already used in this file, I thought that it won't be confusing.

### DIFF
--- a/tests/testthat/test-render.R
+++ b/tests/testthat/test-render.R
@@ -37,6 +37,6 @@ test_that("Input file is not corrupted", {
   input <- tempfile()
   on.exit(unlink(input), add = TRUE, after = FALSE)
   saveRDS("test", input)
-  try(rmarkdown::render(file, quiet = TRUE), silent = TRUE)
+ rmarkdown::render(input, quiet = TRUE)
   expect_equal(readRDS(input), "test")
 })
--- a/tests/testthat/test-render.R
+++ b/tests/testthat/test-render.R
@@ -32,3 +32,11 @@ test_that("file_scope split correctly input file", {
   expect_snapshot_file(splitted[1])
   expect_snapshot_file(splitted[2])
 })
+
+test_that("Input file is not corrupted", {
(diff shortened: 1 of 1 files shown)


tests/testthat/test-render.R
```diff
@@ -32,3 +32,11 @@ test_that("file_scope split correctly input file", {
   expect_snapshot_file(splitted[1])
   expect_snapshot_file(splitted[2])
 })
+
+test_that("Input file is not corrupted", {
+  input <- tempfile()
+  on.exit(unlink(input), add = TRUE, after = FALSE)
+  saveRDS("test", input)
+  try(rmarkdown::render(file, quiet = TRUE), silent = TRUE)
```
yihui, line 40: `file -> input` and remove `try()`:

```suggestion
 rmarkdown::render(input, quiet = TRUE)
```

yihui (approved):
Thank you!

tests/testthat/test-render.R
```diff
@@ -32,3 +32,11 @@ test_that("file_scope split correctly input file", {
   expect_snapshot_file(splitted[1])
   expect_snapshot_file(splitted[2])
 })
+
+test_that("Input file is not corrupted", {
+  input <- tempfile()
+  on.exit(unlink(input), add = TRUE, after = FALSE)
+  saveRDS("test", input)
+ rmarkdown::render(input, quiet = TRUE)
```
yihui, line 40:
```suggestion
  rmarkdown::render(input, quiet = TRUE)
```

yihui (approved):

PR #2550, opened 2024-04-03
Optionize rmarkdown temp directory suffix

certara-tzweers:
We are making this pull request because we have been experiencing an issue with OneDrive deleting html files shortly after knitting them. We have been in email contact with @yihui about updating the knitr_files_dir funtion to aviod this issue (28 Mar 2024, Subject: Rmarkdown: patch for knitr_files_dir).

We have experienced apparently random deletions of html files from OneDrive synced folders shortly after completing the knitting of corresponding Rmd scripts. It appears that this is related to a feature (not bug...) described on this page: 

[Managing the File System - Win32 apps | Microsoft Learn](https://nam11.safelinks.protection.outlook.com/?url=https%3A%2F%2Flearn.microsoft.com%2Fen-us%2Fwindows%2Fwin32%2Fshell%2Fmanage%3Fredirectedfrom%3DMSDN%23connected-files&data=05%7C02%7Cthijs.zweers%40certara.com%7C0c51552ea15d4a508ac008dc52ea4203%7C7287abd30220456e98514352bae208c9%7C1%7C0%7C638476413285887341%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C0%7C%7C%7C&sdata=RgGn24MoT69xwfHNl6zGmTJ02FXARdk%2BN6XOskDrPWg%3D&reserved=0)

As such, the issue is a consequence of different components following their designs with an undesirable outcome. A solution to this could be to change the knitr_files_dir function in util.R so the temporary folder for knitr files is not created with a suffix of "_files" but almost anything else, such as "_knitrfiles".

With this update we have made the knitr temp directory suffix configurable via options(rmarkdown.files.suffix = "_knitrfiles") in the knitr_files_dir function.

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -184,11 +184,7 @@ file_with_meta_ext <- function(file, meta_ext, ext = xfun::file_ext(file)) {
 }
 
 knitr_files_dir <- function(file) {
-  if (!is.null(getOption("rmarkdown.files.suffix"))) {
-    suffix <- getOption("rmarkdown.files.suffix")
-  } else {
-    suffix <- "_files"
-  }
+  suffix <- getOption("rmarkdown.files.suffix", "_files")
   paste(xfun::sans_ext(file), suffix, sep = "")
 }
 
--- a/R/util.R
+++ b/R/util.R
@@ -184,7 +184,12 @@ file_with_meta_ext <- function(file, meta_ext, ext = xfun::file_ext(file)) {
 }
 
 knitr_files_dir <- function(file) {
-  paste(xfun::sans_ext(file), "_files", sep = "")
+  if (!is.null(getOption("rmarkdown.files.suffix"))) {
+    suffix <- getOption("rmarkdown.files.suffix")
+  } else {
+    suffix <- "_files"
+  }
+  paste(xfun::sans_ext(file), suffix, sep = "")
 }
 
 knitr_root_cache_dir <- function(file) {
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 rmarkdown 2.27
 ================================================================================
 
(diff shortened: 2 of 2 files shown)


R/util.R
```diff
@@ -184,7 +184,12 @@ file_with_meta_ext <- function(file, meta_ext, ext = xfun::file_ext(file)) {
 }
 
 knitr_files_dir <- function(file) {
-  paste(xfun::sans_ext(file), "_files", sep = "")
+  if (!is.null(getOption("rmarkdown.files.suffix"))) {
+    suffix <- getOption("rmarkdown.files.suffix")
+  } else {
+    suffix <- "_files"
+  }
```
yihui, line 191: You can just do this:

```suggestion
  suffix <- getOption("rmarkdown.files.suffix", "_files")
```

yihui (changes requested):
Thanks!

cderv:
Thanks a lot. 

Just a little more curious: What exactly happens with this connected files feature ?  
When `*_files` folder is removed then the `.html` file is deleted too. Is that what happens?

Does it happen only on OneDrive folders?  

Thank you !

cderv:
Thanks a lot. I did not know that and I am really happy to learn this today. This could explain some behavior we have seen for a long time now, and also others on another project using **knitr** and **rmarkdown** (Quarto). 

Thanks again for finding this, and for the PR.

th3nme:
While researching this issue we also found a post from 2019 describing the same issue with a Windows 10 and Ubuntu machine linked by Google Drive.

https://forum.posit.co/t/html-from-rmd-run-automatically-removed/25469/3

yihui:
Originally we chose the `_files` suffix precisely because of the association mentioned above (when you copy/move `foo.html`, `foo_files/` is copied/moved automatically). It seems this "intelligent" association does more harm than good in the case of OneDrive.

> Adding the option to change the suffix will give the knitter a way of avoiding this issue but it may be worth considering changing the default from _files to literally almost anything else to avoid the Connected Files default behaviour.

I'm afraid it will be unlikely for us to change the default suffix in the future. It has been used for almost ten years now, and I'm sure that changing the default will create a lot of breaking changes to existing users...

yihui (approved):
@certara-tzweers Although this is a simple one-liner PR, could you sign the CLA anyway? https://cla-assistant.io/rstudio/rmarkdown?pullRequest=2550

PR #2554, opened 2024-04-30
Fix regression w/ image path processing from absolute to relative

cderv:
Processing of HTML needs to account for newlines because pandoc --wrap behavior can sometimes likes to split lines in the HTML like
```
<img>
src="..." </img>
```

Processing line by line would break a reg for img tag's src attribute

This breakage was found in pkgdown processing which uses absolute path for rendering leading to very long image source creating the split

It is also related to `knitr.graphics.rel_path = FALSE` being set fro

Related to 
- https://github.com/r-lib/pkgdown/issues/2496
- https://github.com/r-lib/pkgdown/issues/2341 (requires `knitr.graphics.rel_path = FALSE` due to https://github.com/yihui/knitr/issues/2171)
- https://github.com/r-lib/pkgdown/issues/2334 (requires `knitr.graphics.rel_path = FALSE` due to https://github.com/yihui/knitr/issues/2171) 

See discussion in 0951a2fea7e317f77d27969c25f3194ead38805e

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: rmarkdown
 Title: Dynamic Documents for R
-Version: 2.26.1
+Version: 2.26.2
 Authors@R: c(
     person("JJ", "Allaire", , "jj@posit.co", role = "aut"),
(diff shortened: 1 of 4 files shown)


yihui (approved):
Thanks!

yihui:
It has been almost two months since the last release, so we can plan for a new CRAN release soon (depending on whether we have other stuff that should go with this release).

PR #2559, opened 2024-06-14
Fix Pandoc nightly install in GHA

cderv:
as it seems to be not installing due to auth issue 

See https://github.com/rstudio/rmarkdown/actions/runs/9518128513/job/26246875020#step:7:47


yihui (approved):
Should this PR be merged now?

It seems bookdown needs the same fix: https://github.com/rstudio/bookdown/actions/runs/9763588097/job/26949839471

cderv:
Yes let's merge, and then I'll see how to fix `r-lib/actions` itself. The we'll revert maybe.

PR #2570, opened 2024-08-28
Update actions/checkout to v4

remlapmot:
I was looking at your workflow files for an unrelated reason and noticed all your `actions/checkout` are at old versions.

Merging this will get rid of these GHA warnings you are getting (screenshot below).

CleanShot 2024-08-28 at 10 36 14@2x

Also your `nwtgck/actions-netlify` is at an old version (current is `v3` for that) <https://github.com/nwtgck/actions-netlify>.


cderv (approved):
awesome thanks !

PR #2571, opened 2024-09-09
consistent dev install instructions

luisDVA:
code block uses pak, text above still mentioned remotes. amended.


yihui (approved):
Thanks!

PR #2597, opened 2025-07-17
Fix documentation mistake

uzsolt:
The description contains "title" but it should be "text" as in the example below.


cderv:
Thank you !

PR #2600, opened 2025-08-04
fix(pandoc): Throw error when pandoc isn't available

brianperdomo:
I discovered an issue with the pandoc_convert function while attempting to use the btw package in Github Copilot in VS Code on a machine that did not have Rstudio installed.

If you attempt to use pandoc_convert on a machine without pandoc installed, it throws the following error:

`pandoc document conversion failed with error 127`

.... but there is not a pandoc error 127.

I see that pandoc_convert runs find_pandoc at the beginning, but this doesn't catch the issue. All it does is set .pandoc$version to 0 and .pandoc$dir to NULL.

I have added a call to pandoc_available to catch this issue and provide a more informative error message to the user.

### DIFF
--- a/R/pandoc.R
+++ b/R/pandoc.R
@@ -46,6 +46,10 @@ pandoc_convert <- function(input,
                            verbose = FALSE,
                            wd = NULL) {
 
+  # ensure pandoc is available
+  if (!pandoc_available()) 
+    stop2("pandoc not found, ensure pandoc is installed")
+  
   # ensure we've scanned for pandoc
   find_pandoc()
 
--- a/R/pandoc.R
+++ b/R/pandoc.R
@@ -47,11 +47,7 @@ pandoc_convert <- function(input,
                            wd = NULL) {
 
   # ensure pandoc is available
-  if (!pandoc_available()) 
-    stop2("pandoc not found, ensure pandoc is installed")
-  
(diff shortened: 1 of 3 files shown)


R/pandoc.R
```diff
@@ -46,6 +46,10 @@ pandoc_convert <- function(input,
                            verbose = FALSE,
                            wd = NULL) {
 
+  # ensure pandoc is available
+  if (!pandoc_available()) 
+    stop2("pandoc not found, ensure pandoc is installed")
+  
   # ensure we've scanned for pandoc
   find_pandoc()
```
cderv, line 54: Thanks ! 

`pandoc_available()` is calling `find_pandoc()` so we can probably remove the call after. 

Also `pandoc_available(error = TRUE)` will throw an error. So I believe this could probably be replaced by 
```suggestion
  # ensure pandoc is available
  pandoc_available(error = TRUE)
```

cderv (commented):
Thanks! 

I think this is ok to throw an error in this function. 

@yihui do you agree ? Or am I missing anything ?

cderv (approved):
Thanks! 

I think this is ok to throw an error in this function. 

@yihui do you agree ? Or am I missing anything ?

yihui (approved):
I agree. Thanks!

PR #2608, opened 2026-02-06
Fix typo in env var name in shared.lua comment

huisman:
Fixing small typo in env var name in `shared.lua` file


yihui (approved):
Thanks!

PR #2616, opened 2026-04-13
Update actions

cderv:
for using latest version, and also to support github change to node 20 deprecation

### DIFF
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -20,7 +20,7 @@ jobs:
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
 
       - uses: r-lib/actions/setup-pandoc@v2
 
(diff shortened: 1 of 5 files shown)


.github/workflows/pkgdown.yaml
```diff
@@ -38,17 +38,50 @@ jobs:
...
+      - name: Comment deploy URL on PR
+        if: github.event_name == 'pull_request' && steps.netlify-deploy.outcome == 'success'
+        uses: actions/github-script@v8
+        with:
+          script: |
+            await github.rest.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              body: `🚀 Netlify deploy preview: ${{ steps.netlify-deploy.outputs.url }}`
```
yihui, line 86: For some reason this URL is empty?

yihui (approved):
Currently rlang can't be loaded on R-devel. Hopefully that problem will go away in a few days since rlang has already had a new CRAN release.

.github/workflows/pkgdown.yaml
cderv: Thanks I have seen that. I am trying another regex try

cderv: Url fixed now

yihui (approved):
Looks good to me (not sure about the pandoc installation failures in CI, though; rate limit or something?). Thanks!

PR #2623, opened 2026-07-08
Require Pandoc >= 2.8 and drop Pandoc 1.x support

yihui:
## Motivation

Follow-up to #2622. In that PR, the new `extract-data-uri.lua` filter (which replaced `--extract-media`) was noted by @bastistician to be [more complex than necessary](https://github.com/rstudio/rmarkdown/pull/2622#issuecomment-4767802091) because it had to cope with old-Pandoc quirks.

@cderv [pointed out](https://github.com/rstudio/rmarkdown/pull/2623#issuecomment-4924717587) that jumping straight to Pandoc 3.0 is a big move for organizations that upgrade slowly, and suggested splitting the change: first drop only Pandoc 1.x and bump to the latest Pandoc 2.x minor that our version guards actually need, and require Pandoc 3.0 separately later.

This PR does the first half. It raises the minimum required Pandoc version from **1.14 to 2.8** and removes the version guards and dead code paths that only existed for Pandoc 1.x and early 2.x (`< 2.8`). A follow-up PR (#2628) requires Pandoc 3.0 on top of this.

## Changes

**R**
- Dropped `pandoc2.0()` and the `pandoc_available("1.x"/"2.0"/"2.0.4")` branches: `--latex-engine` (now always `--pdf-engine`), the `markdown_github` variant, the `markdown` fallback in `pandoc_self_contained_html()`, the "requires Pandoc 2.0" guards for custom highlight themes / `anchor_sections` / non-mathjax math, the conditional `--lua-filter` args, and the ioslides Windows codepage workaround.
- Removed the `subtitle.tex` patch (Pandoc < 2.6), `fix_horiz_rule()` (< 2.8), and the beamer template patches (< 1.17.3).
- `render()` now requires Pandoc 2.8.

**Lua filters**
- Dropped the Pandoc 2.1 version guards from `number-sections`, `latex-div`, and `pagebreak`.

**Kept** (guards for features added after 2.8, so Pandoc 2.8–2.18 keep working): `--citeproc`/csljson (2.11), `--embed-resources` (2.19), native gfm math (2.10.1), docx `--number-sections` (2.10.1), pptx `--incremental` (2.15), `yaml_metadata_block` for gfm/commonmark (2.13), `--atx-headers` deprecation (2.11.2), the `accessible-code-block` (2.7.3–2.9.2.1) and `header-attrs` (2.9) HTML dependencies, and `pandoc.system.make_directory` / `pandoc.utils.type` fallbacks in the shared Lua filter.

## Testing

Affected test files pass locally against Pandoc 3.6.3 (old-Pandoc branches are skipped on modern Pandoc as expected). The CI matrix now tests Pandoc 2.8 through the nightly build.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -32,8 +32,10 @@ jobs:
           # testing R release with last shipped pandoc version in RStudio IDE and new pandoc
           - {os: windows-latest, pandoc: 'latest',  r: 'release'}
           - {os: macOS-latest,   pandoc: 'latest',  r: 'release'}
+          - {os: ubuntu-latest,  pandoc: 'latest',  r: 'release'}
           - {os: ubuntu-latest,  pandoc: 'nightly',    r: 'release'}
-          # testing older pandoc versions
+          # testing older pandoc versions (rmarkdown requires Pandoc >= 2.8)
+          - {os: ubuntu-latest,  pandoc: '3.8.3',    r: 'release'}
           - {os: ubuntu-latest,  pandoc: '3.1.11',   r: 'release'}
           - {os: ubuntu-latest,  pandoc: '2.19.2',   r: 'release'}
           - {os: ubuntu-latest,  pandoc: '2.18',     r: 'release'}
@@ -41,8 +43,7 @@ jobs:
           - {os: ubuntu-latest,  pandoc: '2.16.2',   r: 'release'}
           - {os: ubuntu-latest,  pandoc: '2.14.2',   r: 'release'}
           - {os: ubuntu-latest,  pandoc: '2.11.4',   r: 'release'}
-          - {os: ubuntu-latest,  pandoc: '2.7.3',    r: 'release'}
-          - {os: ubuntu-latest,  pandoc: '2.5',      r: 'release'}
+          - {os: ubuntu-latest,  pandoc: '2.8',      r: 'release'}
           # testing other R versions
           - {os: ubuntu-latest,  pandoc: 'latest',   r: 'devel'}
           - {os: ubuntu-latest,  pandoc: 'latest',   r: 'oldrel-1'}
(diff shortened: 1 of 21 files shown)


cderv:
My main concern with this idea of dropping support entirely is the support for organization, and entreprise that moves very slowly ! Old system where pandoc 2 would still be used, possibly shipped through an old RStudio. 

It is hard to know for sure but this is a worry. 

The problem with the lua script is really that ideally we need pandoc 2.8 which is from 2019 - which is 7 years now
https://github.com/rstudio/rmarkdown/blob/194c2a11e3d345e294906152b6f60f0439422022/inst/rmarkdown/lua/shared.lua#L78-L79 

Pandoc 3 is _only_ from 2023 and major bump. 🤷‍♂️ 

I see this allows to replace some pandoc 2.19 requirements too - so maybe this is oldest we could keep with this change ? pandoc 2.19 requirement , and not pandoc 3 ? 

In a way we remove Pandoc 1 support, and we bump minor to latest pandoc 2 minor, and soon will support pandoc 3 only ? 

I guess for such change I was thinking we would need to do a deprecation mechanism. 

- Do a release by throwing a message or a warning in all system that use Pandoc 2, saying we will change to pandoc 3 , and maybe reach out in an issue to explain to us why pandoc 3 would be a problem ? 

R Markdown is widely use. In many private and unknown environnment - so I am extra careful. Especially because R works in a way where using latest available CRAN package is a standard, and locking package version is not widely used (I think). 

So I wonder of the possible side effect, and if this is worth the risk. 
do we gain much to finally makes this Pandoc version support bump ? 
If we want to go through, what about this idea of going in a deprecation cycle like manner ? 

I'll ask around at Posit if we know organization with such constraint.

yihui:
@cderv I just revised this PR to bump Pandoc version to 2.8 first. I think we can consider Pandoc 3.x (PR #2628) after April 2027 when Ubuntu 22.04 reaches end of LTS.

PR #2629, opened 2026-08-24
Fix ioslides figures with Pandoc 3

### LINKED ISSUE #2607: ioslides WARNING: Undefined function 'Figure'
## Checklist

When filing a _bug report_, please check the boxes below to confirm that you have provided us with the information we need. Have you:

- [X] [formatted your issue](https://yihui.org/issue/#please-format-your-issue-correctly) so it is easier for us to read?

- [X] included a minimal, self-contained, and reproducible example?

- [X] pasted the output from `xfun::session_info('rmarkdown')` in your issue?

- [X] upgraded all your packages to their latest versions (including your versions of R, the RStudio IDE, and relevant R packages)?

- [ ] installed and tested your bug with the development version of the rmarkdown package using `remotes::install_github("rstudio/rmarkdown")` ?

## Issue

When knitting to ioslides, I am encountering "WARNING: Undefined function 'Figure'" in unexpected ways, and when I encounter it, the figure I expect to be produced is not produced. I have found two ways to generate the issue; both seem to have something to do with captions, so I believe they're related. The minimal example here contains both ways, along with explanation. (Note that you need some `image.jpg` around to make this example (not) work; it doesn't seem to matter what image it is, though.)

I've checked both output: ioslides (where the issue occurs) and output: html (where there are no warnings and the images display fine, but where I'm still confused about caption behaviour). It turns out that everything I've tried fits into one of two categories: either

| |output:html| output:ioslides|
|---|---|---|
|warning|no|no|
|image|yes|yes|
|caption|no|no|

or
| |output:html|output:ioslides|
|---|---|---|
|warning|no|yes|
|image|yes|no|
|caption|yes|no|

That is, the warning never occurs in output: html, and the image is always produced. But the caption only appears sometimes in output: html. (I don't know why.) And whenever the *caption* is displayed in output: html, the warning occurs and the image and caption both fail to appear in output: ioslides.

## Session info
```
xfun::session_info("rmarkdown")
R version 4.5.2 (2025-10-31)
Platform: aarch64-apple-darwin20
Running under: macOS Sequoia 15.7.1

Locale: en_AU.UTF-8 / en_AU.UTF-8 / en_AU.UTF-8 / C / en_AU.UTF-8 / en_AU.UTF-8

Package version:
  base64enc_0.1.3   bslib_0.9.0       cachem_1.1.0      cli_3.6.5         digest_0.6.39    
  evaluate_1.0.5    fastmap_1.2.0     fontawesome_0.5.3 fs_1.6.6          glue_1.8.0       
  graphics_4.5.2    grDevices_4.5.2   highr_0.11        htmltools_0.5.8.1 jquerylib_0.1.4  
  jsonlite_2.0.0    knitr_1.50        lifecycle_1.0.4   memoise_2.0.1     methods_4.5.2    
  mime_0.13         R6_2.6.1          rappdirs_0.3.3    rlang_1.1.6       rmarkdown_2.30   
  sass_0.4.10       stats_4.5.2       tinytex_0.58      tools_4.5.2       utils_4.5.2      
  xfun_0.54         yaml_2.3.11      

Pandoc version: 3.8.2.1
```

## Example 

````
---
title: example
author: example
output: ioslides
---

# First warning

## Warning 1
The following line generates the warning, and does not work to display an image 
in the rendered slide. With output: html, it generates no error and displays the image 
along with the caption. 

![caption](image.jpg)

# Suppressing the first warning

## Working options
This warning can be suppressed (and the image generated) by having some extra 
text in the markdown nearby. All of the following generate the image and produce 
no warning outputting to both html and ioslides; but none of them displays the caption 
in either output format.

## Text above 

Text on the line above
![caption](image.jpg)

## Text below

![caption](image.jpg)
Text on the line below

## Text before

Text preceding on the same line ![caption](image.jpg)

## Text after

![caption](image.jpg) Text following on the same line

# Second warning

## Warning 2
The code fence with options here is copied directly from 
https://bookdown.org/yihui/rmarkdown/ioslides-presentation.html. (The code inside 
is replaced with an image display rather than a plot.)

The image does not display and the warning is generated when outputting ioslides. 
When outputting html, there is no warning, and the image is displayed along with the caption.

```{r, cars, fig.cap="A scatterplot.", echo=FALSE}
knitr::include_graphics("image.jpg")
```

## Non-warning 2
This modified version of the code fence does not generate the warning and does 
display the image, outputting to either html or ioslides. No caption is displayed in 
either output format.

```{r, cars2, cars2.cap="A scatterplot.", echo=FALSE}
knitr::include_graphics("image.jpg")
```
````

LeonidasZhak:
## Summary

Pandoc 3 represents standalone images with captions as `Figure` elements. The ioslides custom writer did not define that writer function, so the figure content was dropped.

## Thanks to maintainers

Thanks for maintaining rmarkdown and its Pandoc compatibility layer.

## Issue or motivation

Fixes #2607. With Pandoc 3, captioned images in ioslides produced `WARNING: Undefined function 'Figure'` and disappeared from the rendered slide.

## Root cause

The classic writer handled `Image` and `CaptionedImage`, but not the `Figure` element introduced in the Pandoc 3 AST.

## Change

Add a `Figure` handler that preserves the existing ioslides caption markup while returning the rendered figure content. Add a regression test for the Pandoc 3 custom writer.

## Tests

- Focused ioslides figure test: 3/3 expectations passed.
- `R CMD build --no-manual`: passed.
- `R CMD check --as-cran --no-manual`: `Status: OK` on the patched rmarkdown 2.31 source tree, with the remote CRAN incoming lookup disabled because it hung locally.

## Scope

This only changes the ioslides custom writer and adds a regression test. It does not change other output formats, figure sizing, or caption text handling.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,8 @@ rmarkdown 2.32
 
 - Bumped the minimum version of **knitr** to 1.50 to avoid an incompatible combination of **knitr** (< 1.50) and **xfun** (>= 0.56), which fails to load because `xfun::attr()` was removed in xfun 0.56 (thanks, @jsinnett, #2627).
 
+- Fixed captioned figures disappearing from `ioslides_presentation()` output with Pandoc 3, which represents standalone captioned images as `Figure` elements not handled by the custom writer (thanks, @LeonidasZhak, #2607).
+
 
 rmarkdown 2.31
 ================================================================================
--- a/inst/rmd/ioslides/ioslides_presentation.lua
+++ b/inst/rmd/ioslides/ioslides_presentation.lua
@@ -236,6 +236,14 @@ function Plain(s)
   return s
 end
 
+function Figure(s, content, attr)
+  if fig_caption and string.len(s) > 0 then
+    return content .. "<p class='caption'>" .. s .. "</p>"
+  else
+    return content
+  end
+end
+
 function Para(s)
   return "<p>" .. s .. "</p>"
 end
--- a/tests/testthat/test-ioslides-figures.R
+++ b/tests/testthat/test-ioslides-figures.R
@@ -28,5 +28,5 @@ test_that("ioslides renders figures with Pandoc 3", {
   expect_false(any(grepl("Undefined function 'Figure'", messages, fixed = TRUE)))
   rendered <- paste(readLines(output), collapse = "\n")
   expect_match(rendered, "image.jpg", fixed = TRUE)
-  expect_match(rendered, "caption", fixed = TRUE)
+  expect_match(rendered, "<p class='caption'>caption</p>", fixed = TRUE)
 })
--- a/tests/testthat/test-ioslides-figures.R
+++ b/tests/testthat/test-ioslides-figures.R
@@ -1,0 +1,32 @@
+test_that("ioslides renders figures with Pandoc 3", {
+  skip_if_not(rmarkdown::pandoc_available("3.0"))
+
+  input <- tempfile(fileext = ".md")
+  output <- tempfile(fileext = ".html")
+  source_writer <- system.file(
+    "rmd/ioslides/ioslides_presentation.lua",
+    package = "rmarkdown"
+  )
+  writer <- tempfile(fileext = ".lua")
+  writeLines(c(
+    "local fig_caption = true",
+    "local incremental = false",
+    "local smaller = false",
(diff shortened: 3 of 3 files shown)


yihui (approved):
Thanks!

PR #2639, opened 2026-09-07
Use --math-method for Pandoc >= 3.11 to avoid deprecation warnings (#2638)

### LINKED ISSUE #2638: warning when running pandoc
I am running Version 1.24.1 of blogdown, and i get the following warning:
`[WARNING] Deprecated: --mathjax. Use --math-method=mathjax[:URL] instead.`
There seems to be a change in the command line options of pandoc. 
Perhaps this can be updated in the next version of blogdown?

yihui:
Fixes #2638.

## Problem

Pandoc 3.11 (2026-08-28) added `--math-method=METHOD[:URL]` and deprecated the per-engine flags `--mathjax`, `--katex`, `--webtex`, `--mathml`, `--gladtex`. The deprecated flags still work but print a warning:

```
[WARNING] Deprecated: --mathjax. Use --math-method=mathjax[:URL] instead.
```

## Fix

- `pandoc_math_args()` now emits `--math-method=ENGINE[:URL]` when Pandoc >= 3.11 is available, and falls back to the old per-engine flag otherwise. Note the URL separator changes from `=` to `:` in the new form.
- `html_fragment()` routes its hardcoded `"--mathjax"` through `pandoc_math_args()` so it benefits from the same handling.

A new test runs the real Pandoc binary with each math engine (and `self_contained`) arg and fails if Pandoc prints a `Deprecated` warning, so the daily nightly-Pandoc CI job catches such deprecations in advance.

> Note: the more general "warn on *any* deprecated Pandoc argument" mechanism is split into a separate PR, since capturing Pandoc's stderr needs to be gated behind an env var (a naive `2>` redirect broke vignette building on Windows).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


R/pandoc.R
```diff
@@ -492,6 +492,12 @@ pandoc_math_args <- function(engine, url = NULL) {
     stop2(sprintf("%s does not support setting a URL.", engine))
   }
 
+  # Pandoc 3.11 introduced --math-method=METHOD[:URL] and deprecated the
+  # per-engine flags (--mathjax, --katex, ...) which now emit a warning.
+  if (pandoc_available("3.11")) {
+    return(paste0(c("--math-method=", engine, if (!is.null(url)) c(":", url)), collapse = ""))
+  }
+
   paste0(c("--", engine, if (!is.null(url)) c("=", url)), collapse = "")
```
yihui: the two `paste0()` expressions are very similar and should be joined into one

tests/testthat/test-pandoc-deprecation.R
```diff
@@ -0,0 +1,36 @@
...
+    )
+  }
+})
+
+test_that("self_contained args are not deprecated by Pandoc", {
+  skip_if_not_pandoc()
+  skip_on_cran()
+  err <- pandoc_stderr(self_contained_args())
+  expect_no_match(err, "Deprecated", info = paste(err, collapse = " "))
+})
```
yihui: leave out this test; do it in the next PR instead

NEWS.md
```diff
@@ -3,6 +3,8 @@
 
 # rmarkdown 2.32
 
+- The math-engine Pandoc arguments (`--mathjax`, `--katex`, `--webtex`, `--mathml`, `--gladtex`) are now emitted as `--math-method=ENGINE[:URL]` when Pandoc >= 3.11 is available, since Pandoc 3.11 deprecated the per-engine flags and emits a `Deprecated: --mathjax. Use --math-method=mathjax[:URL] instead.` warning for them (thanks, @neuwirthe, #2638).
```
yihui: news item belongs to 2.33

